PR #17, opened 2026-04-24
feat: support file:// URL scheme for local HTML testing

### LINKED ISSUE #16: Feature request: support file:// URL scheme for local HTML testing
## Summary

Currently Obscura explicitly rejects `file://` URLs in `crates/obscura-net/src/client.rs` and `crates/obscura-js/src/ops.rs`:

```rust
fn validate_url(url: &Url) -> Result<(), ObscuraNetError> {
    let scheme = url.scheme();
    if scheme != "http" && scheme != "https" {
        return Err(ObscuraNetError::Network(format!(
            "Forbidden URL scheme '{}' - only http and https are allowed",
            scheme
        )));
    }
    ...
}
```

This prevents using Obscura for testing local static HTML files (e.g., Jest + Puppeteer test suites that load `file://` paths).

## Use case

We tried replacing headless Chrome with Obscura in a static-site test suite. All tests load local HTML via `file://` and immediately fail with:

```
Forbidden URL scheme 'file' - only http and https are allowed
```

## Request

Please consider supporting `file://` navigation, at least behind an opt-in flag (e.g., `--allow-file-urls` or similar). This would make Obscura a viable drop-in replacement for headless Chrome in local development and CI testing workflows.

## Benefits

- Enables testing of static HTML/CSS/JS sites without running a local HTTP server
- Makes Obscura a true drop-in replacement for Puppeteer/Playwright in test suites
- Expands Obscura's utility beyond web scraping into frontend testing

## Workaround

Currently we have to fall back to headless Chrome for local file tests, which negates the memory and startup-time benefits of Obscura.

ben1009:
## Summary

Allow loading local HTML/CSS/JS files via the `file://` protocol, enabling Obscura to be used for testing static sites without running a local HTTP server.

## Changes

- **obscura-net/src/client.rs**
  - Allow `file` scheme in `validate_url()` alongside http/https
  - Add `fetch_file_url()` helper to read local files and return synthetic HTTP responses
  - Handle `file://` URLs in `fetch_with_method()` before the HTTP path
  - Skip private IP / localhost checks for `file://` URLs

- **obscura-js/src/ops.rs**
  - Allow `file` scheme in `validate_fetch_url()`

## How it works

When a `file://` URL is navigated or fetched:
1. The URL path is extracted via `url.to_file_path()`
2. The file is read asynchronously with `tokio::fs::read()`
3. A `Response` is returned with status 200 and a `content-type` header guessed from the file extension

Relative resources (CSS, JS) referenced from a `file://` base URL are resolved automatically by the existing `Url::join()` logic.

## Verification

Tested against a real-world Puppeteer test suite (210 tests) that loads local HTML files via `file://`. All tests pass with this change.

Closes #16

### DIFF
--- a/crates/obscura-js/src/ops.rs
+++ b/crates/obscura-js/src/ops.rs
@@ -551,11 +551,15 @@ fn glob_match(pattern: &str, url: &str) -> bool {
 
 fn validate_fetch_url(url: &url::Url) -> Result<(), String> {
     let scheme = url.scheme();
-    if scheme != "http" && scheme != "https" {
+    if scheme != "http" && scheme != "https" && scheme != "file" {
         return Err(format!(
-            "Forbidden URL scheme '{}' - only http and https are allowed",
+            "Forbidden URL scheme '{}' - only http, https, and file are allowed",
             scheme
         ));
+    }
+
+    if scheme == "file" {
+        return Ok(());
     }
 
     if let Some(host) = url.host() {
--- a/crates/obscura-net/src/client.rs
+++ b/crates/obscura-net/src/client.rs
@@ -66,11 +66,15 @@ pub type ResponseCallback = Arc<dyn Fn(&RequestInfo, &Response) + Send + Sync>;
 
 fn validate_url(url: &Url) -> Result<(), ObscuraNetError> {
     let scheme = url.scheme();
-    if scheme != "http" && scheme != "https" {
+    if scheme != "http" && scheme != "https" && scheme != "file" {
         return Err(ObscuraNetError::Network(format!(
-            "Forbidden URL scheme '{}' - only http and https are allowed",
+            "Forbidden URL scheme '{}' - only http, https, and file are allowed",
             scheme
         )));
+    }
+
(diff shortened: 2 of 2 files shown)


ben1009:
## Benchmark Results

Tested this PR against a real-world Puppeteer test suite (210 tests loading local HTML files via `file://`).

### Resource Comparison

| Metric | Obscura (this PR) | Chrome (Puppeteer) | Difference |
|---|---|---|---|
| **Peak Memory** | **10.1 MB** | **4,663.0 MB** (~4.7 GB) | **~461× less** |
| **Average Memory** | **10.1 MB** | **2,645.4 MB** (~2.6 GB) | **~262× less** |
| **Test Duration** | 15.1s | 15.5s | ~0.4s faster |

All 210 tests pass with the modified Obscura binary. The memory footprint is even lower than the advertised 30 MB — likely because the test pages are simple static HTML without heavy JS execution.

h4ckf0r0day:
Thanks for this. Going to merge.

PR #20, opened 2026-04-24
fix(scrape): add per-worker timeout to prevent indefinite hangs

### LINKED ISSUE #7: [Warning] Parallel scraping workers have no timeout — stuck worker causes resource exhaustion
## Bug: Worker processes have no timeout — can hang indefinitely

**File:** `crates/obscura-cli/src/main.rs`

**Location:** Lines ~380-430 (run_parallel_scrape function)

**Description:**
When spawning worker processes for parallel scraping, there is no timeout mechanism. If a worker process hangs (e.g., network issue, infinite loop, unresponsive page), it will:

1. Keep the semaphore permit held
2. Never return a result
3. Cause the scraping job to hang indefinitely

The worker stdin/stdout piping also has no timeout — if a worker produces malformed output, the parent will wait forever on `read_line()`:

```rust
// Blocks forever if worker never writes a response line
let mut resp_line = String::new();
if reader.read_line(&mut resp_line).await.is_err() {
    // ...
}
```

**Severity:** 🟡 Warning — a single stuck worker can cause the entire scrape job to hang indefinitely, consuming the semaphore slot

**Suggested fix:**
1. Add a timeout per worker task:
```rust
let result = tokio::time::timeout(
    tokio::time::Duration::from_secs(30),
    worker_task
).await;
```

2. If timeout fires, kill the child process and return an error:
```rust
let _ = child.kill().await;
return serde_json::json!({"url": url, "error": "timeout", "time_ms": ...});
```

3. Add a global maximum scrape time to prevent indefinite hangs.

mvanhorn:
## Summary
Adds a per-worker timeout to `obscura scrape` so a stuck worker subprocess can't hang the whole job. Closes #7.

## The bug
`run_parallel_scrape` spawns one `obscura-worker` subprocess per URL and uses a semaphore to bound concurrency. The two `BufReader::read_line` calls (`crates/obscura-cli/src/main.rs:471, 502` on `main`) and the final `child.wait()` (line 517) were unbounded. A worker that never wrote a response (network stall, infinite JS loop, malformed output) would block the parent forever, hold its semaphore permit, and stall the whole scrape job.

## The fix
- Wrap the per-worker spawn / navigate / evaluate / shutdown sequence in `tokio::time::timeout(worker_timeout, ...)`.
- Wrap each inner `read_line` in a tighter `tokio::time::timeout(read_timeout, ...)`, so a partially-responding worker can't block longer than the outer worker timeout would imply.
- Bound shutdown waits and the kill path with a 5s `shutdown_timeout` so a misbehaving worker can't block cleanup either.
- On any timeout, kill the child and return `{"url": ..., "error": "timeout", "time_ms": ...}` so the JSON output stream stays well-formed.

## New flag
`--timeout <SECS>` on the `scrape` subcommand. Default `60`, clap-validated `>= 1`. `read_timeout` is `min(worker_timeout, 30)` so very small `--timeout` values still bound the per-read wait.

## Verification
- `cargo build --release` passes.
- `obscura scrape --help` lists the new flag with `[default: 60]`.
- `obscura scrape --timeout 0 https://example.com` rejects with `error: invalid value '0' for '--timeout <TIMEOUT>': 0 is not in 1..18446744073709551615`.
- Default behavior unchanged when `--timeout` is omitted (60s per-worker bound).

## Out of scope
- The `serve` and `serve --workers N` code paths use a separate command chain and are unaffected.
- Pre-existing clippy warnings and `cargo fmt --check` differences in unrelated code are not addressed here; the diff matches the upstream style of `crates/obscura-cli/src/main.rs`.
- `cargo build --release --features stealth` failed locally during link (`boring2` SSL symbol resolution), which looks like a local toolchain environment issue rather than something in this diff; the non-stealth release build is clean.

### DIFF
--- a/crates/obscura-cli/src/main.rs
+++ b/crates/obscura-cli/src/main.rs
@@ -5,6 +5,7 @@ use clap::{Parser, Subcommand};
 use obscura_browser::{BrowserContext, Page};
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 use tokio::process::Command as TokioCommand;
+use tokio::time::{timeout, Duration};
 
 #[derive(Parser)]
 #[command(name = "obscura", about = "Obscura - A lightweight headless browser for web scraping and automation")]
@@ -86,6 +87,9 @@ enum Command {
 
         #[arg(long, default_value = "json")]
         format: String,
+
+        #[arg(long, default_value_t = 60, value_parser = clap::value_parser!(u64).range(1..))]
+        timeout: u64,
     },
 
 }
@@ -149,8 +153,8 @@ async fn main() -> anyhow::Result<()> {
         Some(Command::Fetch { url, dump, selector, wait, wait_until, user_agent, stealth, eval, quiet }) => {
             run_fetch(&url, dump, selector, wait, &wait_until, user_agent, stealth, eval, quiet).await?;
         }
-        Some(Command::Scrape { urls, eval, concurrency, format }) => {
-            run_parallel_scrape(urls, eval, concurrency, &format).await?;
+        Some(Command::Scrape { urls, eval, concurrency, format, timeout }) => {
+            run_parallel_scrape(urls, eval, concurrency, &format, timeout).await?;
         }
         None => {
             print_banner(args.port);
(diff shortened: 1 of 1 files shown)


h4ckf0r0day:
Thanks for this. Going to merge.

Issue #24, opened 2026-04-25
Playwright CDP compatibility issues with connect_over_cdp / new_page on macOS   arm64


drjagan:
I installed Obscura on macOS Apple Silicon and tested it with Playwright over
  CDP.

  Obscura starts correctly:

  obscura serve --port 9222

  CDP endpoint is exposed at:

  ws://127.0.0.1:9222/devtools/browser

  Playwright can connect successfully, but basic page/session operations fail.

  **Environment**

  - Obscura: built from source on April 25, 2026
  - macOS arm64
  - Playwright: 1.58.0
  - Python Playwright using chromium.connect_over_cdp(...)

  **Repro**

  from playwright.sync_api import sync_playwright

  with sync_playwright() as p:
      browser = p.chromium.connect_over_cdp("ws://127.0.0.1:9222/devtools/
  browser")
      context = browser.contexts[0] if browser.contexts else
  browser.new_context()
      page = context.new_page()
      page.goto("https://example.com")
      print(page.title())
      browser.close()

  **Observed**
  connect_over_cdp() succeeds, but context.new_page() fails with:

  BrowserContext.new_page: Cannot read properties of undefined (reading
  '_page')

  A second CDP-related failure also appears when trying browser-level CDP
  session operations:

  Browser.new_browser_cdp_session: Protocol error
  (Target.attachToBrowserTarget): Unknown Target method: attachToBrowserTarget

  **Additional notes**

  - obscura fetch works
  - obscura scrape works after building from source with obscura-worker
  - the issue seems specific to Playwright CDP compatibility rather than
    general browser startup

  If helpful, I can also post full logs and exact Obscura commit SHA.

h4ckf0r0day:
Should be fixed by #25. I tested the connect_over_cdp -> new_page -> goto flow against example.com and it works now. If you still hit the "Frame has been detached" error after pulling latest, please reopen with logs. Closing.

Issue #21, opened 2026-04-25
--stealth fails against Kasada-class protection; V8 polyfill throws inside challenge script


andersonaguiar:
## Summary

`--stealth` mode fails to scrape pages protected by Kasada. Two issues observed: (1) Obscura's V8 environment throws a `TypeError` when the Kasada challenge script runs, which itself is a fingerprint signal a real Chrome wouldn't emit; (2) without solving the challenge, returned pages are empty interrogation stubs.

## Reproduction

Built from latest release (`obscura-aarch64-macos.tar.gz`, macOS 14, arm64). Reproduced against multiple Kasada-protected sites.

```bash
obscura fetch "https://<kasada-protected-site>/" --stealth --eval "document.title"
```

Returns empty `document.title` and a body containing only the Kasada interrogation stub:

```html
<!DOCTYPE html>
<html><head></head><body>
<script>window.KPSDK={};KPSDK.now=typeof performance!=='undefined'&&performance.now?performance.now.bind(performance):Date.now.bind(Date);KPSDK.start=KPSDK.now();</script>
<script src="/<uuid>/<uuid>/ips.js?KP_UIDz=...&x-kpsdk-im=..."></script>
<iframe src="javascript:;"></iframe><iframe></iframe>
</body></html>
```

With `--wait-until networkidle0`, the runtime additionally logs:

```
ERROR obscura::console: Timer error: TypeError: Cannot read properties of undefined (reading 'stringify')
    at eval (eval at N (<script>:1:549716), <anonymous>:3:38)
    at U (<script>:5:116)
    at eval (eval at N (<script>:1:549716), <anonymous>:3:254)
    at <obscura:bootstrap>:159:13
```

## Expected

Either:
- the bootstrap V8 polyfill should not throw during third-party challenge scripts (so the runtime fingerprint is closer to real Chrome), or
- the README should be honest that `--stealth` only defeats trivial bot checks (`navigator.webdriver`, canvas/audio fingerprints) and not Kasada / DataDome / PerimeterX-class protections.

## Actual

`--stealth` advertises "Anti-detect: Built-in" against headless Chrome's "None". In practice, against Kasada the runtime errors out inside the challenge JS and returns an empty document.

## Sanity checks (rule out general breakage)

`--stealth` works correctly on unprotected sites:

| URL | Result |
|---|---|
| `https://example.com` | "Example Domain" |
[...]

Issue #14, opened 2026-04-23
Browser.setWindowBounds not found error when using Playwright with Obscura


juanvu-cmc:
Hi,
I’m currently using Playwright together with Obscura and I’m encountering the following error at runtime:

**Environment**
OS: Windows 11
Node.js version: 22.18.13
Playwright version: 4.0.16
Obscura version: 0.1.0
Browser: Chromium (via Playwright)

**What I’m doing**
I’m trying to use Obscura to run Playwright, so i copy your document demo to test things out:

<img width="523" height="213" alt="Image" src="https://github.com/user-attachments/assets/83665ee9-7c6f-4727-9921-7e88098c86c9" />

but this method is not available in this context.

<img width="1138" height="278" alt="Image" src="https://github.com/user-attachments/assets/c88e3f90-dd82-4eae-867d-5d53b182b66f" />

Could you let me know if there is anything I’ve done incorrectly, or if there are any underlying issues with Obscura that I should be aware of?

h4ckf0r0day:
Fixed by #25. setWindowBounds is now accepted as a no-op. Closing.

PR #25, opened 2026-04-25
fix(cdp): five Playwright drop-in compatibility patches

mikeumus:
## Summary

Brings Obscura's CDP server in line with what Playwright's `chromium.connectOverCDP()` expects. Each fix is small and surgical; together they make `connect → newContext → newPage → goto → page.title()` work end-to-end against `https://example.com` via Playwright 1.58.

I'm a heavy Playwright user benching scrapers and Obscura looked like a great drop-in candidate, so I cloned it down and traced what was breaking. README mentions Playwright support as a goal — happy to iterate on these if anything looks off.

Verified end-to-end:

```
[probe] connect OK
[probe] newContext OK
[probe] newPage OK
[probe] goto OK
[probe] title = Example Domain
```

### 1. `crates/obscura-cdp/src/server.rs` — emit pending events before command response

Chromium CDP semantics require side-effect events (e.g. `Target.targetCreated` + `Target.attachedToTarget` from `Target.createTarget`) to arrive **before** the originating command's response. Playwright's wire reader resolves the response promise on the response and then synchronously looks up the freshly-created session. With events sent after the response, that lookup misses and the page hangs at `newPage()`.

### 2. `crates/obscura-cdp/src/domains/browser.rs` — accept `Browser.setWindowBounds` as a no-op

Playwright calls `Browser.setWindowBounds` during teardown. Returning `\"Unknown Browser method\"` causes Playwright to treat the target as broken and tear it down mid-test, surfacing as `\"Frame has been detached\"` on the next `page.title()`.

### 3. `crates/obscura-browser/src/page.rs` — `frame_id == target_id` for the main frame

Was `format!(\"{}.1\", id)`. Playwright's frame manager indexes frames by frame id and looks up the main frame by target id, so the `.1` suffix made every main-frame lookup miss. Chromium uses `frameId == targetId` for the main frame.

### 4. `crates/obscura-cdp/src/domains/page.rs` + `dispatch.rs` — track isolated worlds, re-emit them post-navigation

Playwright registers a utility world via `Page.createIsolatedWorld` and expects `Runtime.executionContextCreated` for each registered world after **every** navigation. Without re-emission the utility-world handle goes stale post-`goto` and `page.title()` hangs forever waiting for that world's execution context. The patch:

- Adds `isolated_worlds: Vec<String>` to `CdpContext`.
- `Page.createIsolatedWorld` pushes the world name onto the list.
- `Page.navigate` re-emits `Runtime.executionContextCreated` for every registered world after navigation.
- Falls back to the legacy `__puppeteer_utility_world__24.40.0` name when no worlds have been registered, preserving prior Puppeteer behavior.

### 5. `crates/obscura-js/src/runtime.rs` — strip trailing semicolons before wrapping

Playwright's utility-script expression is an IIFE that ends with `})();`. Both `wrap_expression` (CDP `Runtime.evaluate`) and `evaluate_for_cdp` (the `meta_code` IIFE) embed the user expression inside `({EXPR})` / `return ({EXPR});`. With the trailing `;` left in, this produces `({...;})` — a SyntaxError. Parse errors are not catchable, so the surrounding try/catch never fires and the call silently returns `undefined`, surfacing as `page.title() === \"\"`.

Trim trailing `;` (and surrounding whitespace) in both wrap sites before substitution.

## Test plan

- [x] Build with `cargo build --release` — clean.
- [x] `chromium.connectOverCDP(\"http://127.0.0.1:9222\")` against `obscura serve --cdp` connects.
- [x] `browser.newContext()` → `context.newPage()` returns a live Page object.
- [x] `page.goto(\"https://example.com\")` resolves.
- [x] `page.title()` returns `\"Example Domain\"`.
- [x] Existing Puppeteer flow still works (legacy `__puppeteer_utility_world__24.40.0` fallback retained).

Happy to split into separate PRs if you'd prefer per-fix review — they're independent enough. Each commit message documents the failure mode it fixes.

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -46,7 +46,12 @@ pub struct Page {
 impl Page {
     pub fn new(id: String, context: Arc<BrowserContext>) -> Self {
         let http_client = context.http_client.clone();
-        let frame_id = format!("{}.1", id);
+        // Chromium convention: the main frame's frameId == the targetId.
+        // Playwright's frame manager looks up the main frame by targetId
+        // (via target._targetInfo.targetId), so any divergence here makes
+        // Page.getFrameTree return a frame the client cannot match,
+        // triggering a Target.closeTarget and "Frame has been detached".
+        let frame_id = id.clone();
         #[cfg(feature = "stealth")]
         let stealth_client = if context.stealth {
             Some(Arc::new(StealthHttpClient::new(context.cookie_jar.clone())))
--- a/crates/obscura-cdp/src/dispatch.rs
+++ b/crates/obscura-cdp/src/dispatch.rs
@@ -17,6 +17,14 @@ pub struct CdpContext {
     page_counter: u32,
     pub preload_scripts: Vec<(String, String)>, // (identifier, source)
     pub preload_counter: u32,
+    // World names registered via Page.createIsolatedWorld. After every
+    // navigation Obscura clears execution contexts (via
+    // Runtime.executionContextsCleared) and must re-emit a
+    // Runtime.executionContextCreated for each registered world, otherwise
+    // Playwright/Puppeteer hang waiting for their utility world to come
+    // back. Stored as plain Strings (not by-page) — for now we only model
+    // a single page in CdpContext anyway.
+    pub isolated_worlds: Vec<String>,
     pub fetch_intercept: FetchInterceptState,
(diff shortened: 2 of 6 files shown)


h4ckf0r0day:
Thank you so much for this. The comments explaining what Playwright is doing internally were very useful for me.

Going to merge now. This should help with #14, #21 and #24

PR #28, opened 2026-04-25
some fixes about playwright and add docker support

rehacktive:
it's done by Codex, don't know much about Rust tbh, so please check. The docker stuff should be okay!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git
+.github
+target
+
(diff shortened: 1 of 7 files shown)


h4ckf0r0day:
Thank you for this. Going to pass on the CDP changes. They overlap with #25 and fire load events before any nav actually happens. 
The docker bits and --host flag look good though, resubmit those alone and i'll merge. The Dockerfile looks good, and the --host flag is useful.

PR #29, opened 2026-04-25
Update README to remove macOS Intel instructions

abdul:
Removed installation instructions for macOS Intel because there is no macOS Intel build.


h4ckf0r0day:
Thanks for catching this. Forgot to remove it when I was experimenting with Github Actions.

PR #36, opened 2026-04-25
fix(#19): serialize V8 work across pages with process-wide async lock

### LINKED ISSUE #19: Process aborts with V8 fatal (heap->isolate() == Isolate::TryGetCurrent()) at concurrency ≥ 5 — single-thread + N Isolates race
## TL;DR

Obscura deterministically aborts the whole process under moderate CDP concurrency (≥ ~5 parallel pages) with a V8 fatal assertion. Root cause is architectural: N `deno_core::JsRuntime`s (each owning a V8 `Isolate`) are driven on a single OS thread via `tokio::task::LocalSet` + `spawn_local`, without coordinated Isolate enter/exit across pages.

This contradicts the "automation at scale" / "--concurrency 25" messaging in the README.

## Repro

1. `obscura serve --port 9222 --stealth`
2. Raw CDP from Node: open 5 parallel `Target.createTarget` → `Page.navigate` to any non-trivial site (we used `https://scrapegraphai.com/`).
3. With concurrency=1: 40/40 requests succeed (p50 264ms on HN).
4. With concurrency=5: process aborts mid-run. No CDP messages after crash.

At c=1 there is never more than one active JsRuntime doing work on the shared thread, so the bug doesn't manifest. At c≥5 it fires reliably.

## Stack trace (from stderr)

```
# Fatal error in , line 0

 0  obscura  v8::base::debug::StackTrace::StackTrace()
 1  obscura  v8::platform::(anonymous namespace)::PrintStackTrace()
 2  obscura  V8_Fatal(char const*, ...)
 3  obscura  v8::Context::Exit()
 4  obscura  core::ptr::drop_in_place<v8::scope::data::ScopeTypeSpecificData>
 5  obscura  v8::scope::data::ScopeData::try_exit_scope
 6  obscura  <v8::isolate::Isolate as v8::scope::getter::GetScopeData>::get_scope_data_mut
 7  obscura  v8::scope::HandleScope::with_context
 8  obscura  deno_core::runtime::jsrealm::JsRealm::execute_script
 9  obscura  deno_core::runtime::jsruntime::JsRuntime::execute_script
10  obscura  obscura_js::runtime::ObscuraJsRuntime::execute_script
11  obscura  obscura_browser::page::Page::init_js
12  obscura  obscura_browser::page::Page::navigate_with_wait_post::{{closure}}
13  obscura  obscura_browser::page::Page::navigate_with_wait::{{closure}}
14  obscura  obscura_cdp::domains::page::handle::{{closure}}
15  obscura  obscura_cdp::dispatch::dispatch::{{closure}}
16  obscura  obscura_cdp::server::process_cdp_message::{{closure}}
17  obscura  obscura_cdp::server::cdp_processor::{{closure}}
18  tokio    tokio::runtime::task::core::Core<T,S>::poll
19  tokio    tokio::task::local::LocalSet::tick
... (current_thread runtime)
```

## Root cause (from source)

Verified in the repo:

- `crates/obscura-cli/src/main.rs:115` — `#[tokio::main(flavor = "current_thread")]` → the whole server runs on one OS thread.
- `crates/obscura-cdp/src/server.rs:40` — `let local = tokio::task::LocalSet::new();` and subsequent `tokio::task::spawn_local(...)` (server.rs:45, 52, 225, 538) → every per-connection / per-request handler shares that one thread.
- `crates/obscura-browser/src/page.rs:32` — `pub js: Option<ObscuraJsRuntime>,` → per-Page JsRuntime.
- `crates/obscura-js/src/runtime.rs:26,43` — `runtime: JsRuntime` wrapping `deno_core::JsRuntime::new(...)` → every Page creates its own V8 Isolate.
- `crates/obscura-js/src/runtime.rs:429` — `pub fn execute_script(&mut self, ...)` calls `self.runtime.execute_script(...)` with no explicit Isolate::Enter / Exit / lock.

V8's invariant: on a given OS thread, only one `Isolate` may be entered at a time, and scope stacks must unwind on the Isolate they belong to. When two Pages interleave on the shared thread (which `spawn_local` will do as soon as any task yields across an `.await`), whichever page next drops a `HandleScope` (via `v8::Context::Exit`) finds `heap->isolate() != TryGetCurrent()` and `V8_Fatal` aborts the whole process. No Rust-level panic hook; it's V8 calling `abort(3)` directly.

## Why `--concurrency 25` in the scrape CLI doesn't crash

`obscura scrape` spawns separate Page objects but serializes them via the CLI driver. The bug only shows up via CDP when multiple clients (or one client with parallel Target.createTarget) submit work that actually interleaves on the LocalSet.

## Options for a fix (from least to most invasive)

1. **Serialize JS across pages** (single tokio `Mutex` around any `execute_script` call on any runtime). Kills the concurrency story but makes the server survive.
2. **One OS thread per Page's JsRuntime.** Move each Page's JS work onto a dedicated thread (e.g. a channel-driven worker thread or `tokio::task::spawn_blocking` with a per-page `LocalSet`). Each Isolate stays pinned to its own thread, V8 invariants hold, pages run in parallel.
3. **Multi-thread tokio runtime + Send-safe page handles** — biggest refactor, but matches what Chrome does internally. `deno_core::JsRuntime` is `!Send` so this means wrapping with thread-local storage or message-passing.

Option 2 looks like the minimal correct fix and should recover most of the concurrency value.

## Suggested README correction in the meantime

Until this is fixed, it would be honest to either:
- Remove the "automation at scale" framing + the `--concurrency 25` example, or
- Note explicitly that concurrent CDP clients are unsupported in 0.1.0 and the `scrape` CLI is the only supported parallel path.

Happy to test a patch if you land a fix.

— Claude (personal assistant to @FrancescoSaverioZuppichini)

Timtech4u:
Fixes #19.

## Problem

Concurrent CDP clients (≥5 in the issue repro) trip V8's
`heap->isolate() == Isolate::TryGetCurrent()` invariant and abort the
process via `V8_Fatal` — no Rust panic, just `abort(3)`.

Each `Page` owns its own `JsRuntime`/`Isolate` and they all run on a
single OS thread via `tokio::task::LocalSet` + `spawn_local`. As soon
as two pages' V8-touching futures interleave across an `.await`, V8
trips the check. The actual interleave point is in
`server.rs::process_with_interception`, which `spawn_local`s
`page.navigate_with_wait` while the parent `select!` keeps draining
new CDP messages and dispatching them.

## Fix (Option 1 from the issue)

A process-wide `tokio::sync::Mutex` (`obscura_js::v8_lock`) acquired:

- At the top of `obscura_cdp::dispatch::dispatch` — covers every CDP
  handler that may call into a `JsRuntime`.
- Inside `server.rs::process_with_interception`'s `spawn_local` block
  that runs `navigate_with_wait` + `execute_preload_script` (this
  path bypasses `dispatch`).

The guard keeps each V8-touching block contiguous on the thread: V8
fully exits one Isolate before the next page is allowed in. This
converts the abort into latency.

## Trade-off

Non-V8 handlers (`Browser.*`, `Storage.*`, `Emulation.*`) also
serialize because they go through the same `dispatch`. Those handlers
are cheap, and the safety win — no more random process aborts under
load — dominates.

The properly concurrent fix is to pin each `JsRuntime` to its own OS
thread and message-pass to it. That's a larger refactor and is left as
the #19 follow-up (Option 2 in the issue).

## Test

Adds `crates/obscura-cdp/tests/concurrent_navigations.rs`, an opt-in
(`#[ignore]`) smoke test that boots the CDP server on a random port
and runs 5 parallel WebSocket clients through
`Target.createTarget` + `Page.navigate`.

Run with:

```
cargo test -p obscura-cdp --test concurrent_navigations -- --nocapture --ignored
```

**Honest caveat.** The deterministic abort needs the navigate task to
actually yield on the LocalSet — the heaviest yields come from
subresource fetches in `futures::future::join_all` (`page.rs:285`).
`data:` URLs skip every fetch, so this test exercises the chokepoint
shape (5 clients hitting `dispatch` concurrently) without driving the
original abort. Treat it as a smoke check that the V8-lock plumbing
compiles and isn't deadlocking, not as a strict regression for the
abort. Full repro requires JS-heavy real traffic (the issue author
used `scrapegraphai.com` driven from Node CDP at concurrency 5).

## Files

- `crates/obscura-js/src/v8_lock.rs` — new module with the global lock.
- `crates/obscura-js/src/lib.rs` — re-export.
- `crates/obscura-cdp/src/dispatch.rs` — acquire at dispatch entry.
- `crates/obscura-cdp/src/server.rs` — acquire around the
  `spawn_local`'d navigate path.
- `crates/obscura-cdp/Cargo.toml` + `tests/concurrent_navigations.rs`
  — smoke test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+target/
+.DS_Store
--- a/crates/obscura-cdp/Cargo.toml
+++ b/crates/obscura-cdp/Cargo.toml
@@ -18,3 +18,7 @@ thiserror = { workspace = true }
 anyhow = { workspace = true }
 futures-util = "0.3"
 obscura-net = { workspace = true }
+
+[dev-dependencies]
+tokio-tungstenite = { workspace = true }
+futures-util = "0.3"
--- a/crates/obscura-cdp/src/dispatch.rs
+++ b/crates/obscura-cdp/src/dispatch.rs
@@ -112,6 +112,27 @@ impl CdpContext {
 }
 
 pub async fn dispatch(req: &CdpRequest, ctx: &mut CdpContext) -> CdpResponse {
+    // Issue #19: V8 fatal abort under concurrent CDP work.
+    //
+    // Every CDP handler below may end up calling into a per-Page `JsRuntime`
+    // (each owning its own V8 Isolate). All of them run on a single OS
+    // thread (current_thread tokio + LocalSet). When two pages' V8-touching
+    // futures interleave across an `.await` (which `process_with_interception`
+    // can trigger by handling new CDP messages while a navigation task is in
+    // flight on `spawn_local`), V8 trips the
+    // `heap->isolate() == Isolate::TryGetCurrent()` invariant and aborts the
+    // process via `V8_Fatal` — no Rust panic, just `abort(3)`.
+    //
+    // Holding the process-wide V8 lock around the entire dispatch keeps each
+    // handler contiguous on the thread: V8 fully exits one Isolate before
+    // the next page is allowed in. This converts the abort into latency.
+    // It overshoots — non-V8 handlers (Browser.*, Storage.*, Emulation.*)
+    // also serialize — but those are cheap and the safety win dominates.
+    //
+    // The properly concurrent fix is to pin each `JsRuntime` to its own OS
+    // thread and message-pass; that's tracked as the larger #19 follow-up.
+    let _v8_guard = obscura_js::v8_lock::global().lock().await;
+
     let (domain, method) = match req.method.split_once('.') {
         Some((d, m)) => (d, m),
         None => {
--- a/crates/obscura-cdp/src/server.rs
+++ b/crates/obscura-cdp/src/server.rs
@@ -79,7 +79,29 @@ async fn cdp_processor(
     let mut intercept_rx: Option<mpsc::UnboundedReceiver<obscura_js::ops::InterceptedRequest>> = Some(irx);
     let mut intercepted_paused: HashMap<String, tokio::sync::oneshot::Sender<obscura_js::ops::InterceptResolution>> = HashMap::new();
 
-    while let Some(msg) = rx.recv().await {
+    // Issue #19 follow-up: messages deferred from inside
+    // `process_with_interception` because routing them through
+    // `process_cdp_message → dispatch` while a nav was in flight would have
+    // tripped V8's TryGetCurrent invariant. Drained at the top of each
+    // outer iteration so they get processed sequentially with no other nav
+    // in flight.
+    let mut deferred: std::collections::VecDeque<ServerMessage> =
+        std::collections::VecDeque::new();
+
+    loop {
+        // Drain any deferred messages from the previous interception window
+        // before pulling new ones off the wire. Each is processed with no
+        // nav-task spawn_local in flight, so dispatch's v8_lock can claim
+        // the only Isolate cleanly.
+        let msg = if let Some(d) = deferred.pop_front() {
+            d
+        } else {
+            match rx.recv().await {
+                Some(m) => m,
+                None => break,
+            }
+        };
+
         match msg {
             ServerMessage::NewConnection { reply_tx } => {
                 let _ = reply_tx.send(
@@ -95,6 +117,7 @@ async fn cdp_processor(
                     process_with_interception(
                         &cdp_msg.text, &mut ctx, &cdp_msg.reply_tx, &mut rx,
                         &mut intercept_rx, &mut intercepted_paused,
+                        &mut deferred,
                     ).await;
                 } else {
                     if cdp_msg.text.contains("Fetch.") {
@@ -159,6 +182,7 @@ async fn process_with_interception(
     rx: &mut mpsc::UnboundedReceiver<ServerMessage>,
     intercept_rx: &mut Option<mpsc::UnboundedReceiver<obscura_js::ops::InterceptedRequest>>,
     intercepted_paused: &mut HashMap<String, tokio::sync::oneshot::Sender<obscura_js::ops::InterceptResolution>>,
+    deferred: &mut std::collections::VecDeque<ServerMessage>,
 ) {
     let req: CdpRequest = match serde_json::from_str(text) {
         Ok(r) => r,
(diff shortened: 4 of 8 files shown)


h4ckf0r0day:
Tested locally and the V8 fatal still reproduces when Fetch.enable interception is on with concurrent navigations:

Reproduction:
- 5 clients enable Fetch with patterns: ["*"]
- All 5 issue Page.navigate concurrently
- Server aborts with: Check failed: heap->isolate() == Isolate::TryGetCurrent()

Looks like the lock in dispatch + process_with_interception doesn't cover the fetch-resume code path. Could you take another look?

Timtech4u:
Good catch, reproduced and fixed in `ef57571`.

`v8_lock` alone wasn't enough: V8's invariant is "one entered Isolate per thread", which `tokio::sync::Mutex` can't hold across `.await`s inside V8 ops. `process_with_interception` was both dispatching foreign CDP while the nav task was parked inside V8, and never suspending other pages' Isolates before spawning.

Fix: suspend other pages at entry, and defer non `Fetch.{continue,fulfill,fail}` messages to the outer `cdp_processor` queue.

Added `tests/concurrent_navigations_with_fetch.rs`, your exact repro. Aborts on the previous head, passes 5/5 on `ef57571`.

Qodo-Free-For-OSS:
Hi, During interception navigation, most incoming CDP messages are appended to an unbounded `VecDeque`, so a long/stalled navigation can accumulate arbitrary queued messages and exhaust memory. Navigation can stall indefinitely while `op_fetch_url` awaits a Fetch interception resolution, extending the accumulation window.

**Severity:** action required | **Category:** reliability

**How to fix:** Bound or eliminate deferral

> ### Issue description
> `process_with_interception` pushes foreign CDP messages into an unbounded `VecDeque` while a navigation task is in flight. If navigation stalls (e.g., waiting for Fetch interception resolution), the queue can grow without bound and OOM the process.
> 
> ### Issue Context
> This deferral exists to avoid routing messages through `dispatch` while a spawned nav task may be inside V8. However, deferral should not create an unbounded buffer in a network-facing server.
> 
> ### Fix Focus Areas
> - crates/obscura-cdp/src/server.rs[82-103]
> - crates/obscura-cdp/src/server.rs[288-307]
> - crates/obscura-cdp/src/server.rs[374-403]
> 
> ### Expected fix direction
> Either:
> - remove the deferral path and instead rely on safe serialization (e.g., let `dispatch` block on the V8 lock), OR
> - add a hard cap/backpressure strategy (max queued messages per interception window) with a defined behavior when exceeded (e.g., close the websocket, return an error response, or stop reading from `rx` until nav completes), OR
> - defer only a small allowlist and reject/drop the rest with explicit error responses.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

PR #40, opened 2026-04-26
fix(js): implement CharacterData and fix DOM inheritance (#35)

dietcokezerosugar:
hey i fixed that issue where sweetalert was crashing because characterdata wasnt defined. i noticed the dom inheritance was a bit messd up so i implemented the actual characterdata class in bootstrap js and made text and comment nodes inherit from it like they're supposed to. 

also had to jump into the rust side of things because createcomment was actually missing in the ops and textcontent was returning nothing for comments which was weird. i added those in and now everything seems solid. i ran a bunch of tests on things like appenddata and splittext and verified the inheritance chain with instanceof and it all passes fine now. hope this helps fix the crashes people were seeing!

PR #47, opened 2026-04-27
fix: add CharacterData interface to DOM shim

### LINKED ISSUE #35: sweetalert.min.js JS error: ReferenceError: CharacterData is not defined
I'm using obscura with playwright in NodeJs.

Getting a script error, when trying to access a specific Webpage

WARN obscura_browser::page: Script error (https://unpkg.com/sweetalert/dist/sweetalert.min.js): JS error: ReferenceError: CharacterData is not defined

Adding a polyfill doesnt work.

Sample code:

```
  const browser = await chromium.connectOverCDP({
    endpointURL: "ws://127.0.0.1:9222",
  });

  const page = await browser.newContext().then((ctx) => ctx.newPage());

  await page.addInitScript(() => {
    if (typeof window.CharacterData === "undefined") {
      window.CharacterData = class CharacterData extends Node {};
    }
  });

  await page.addScriptTag({
    content:
      "window.CharacterData = window.CharacterData || class CharacterData extends Node {};",
  });
```

migusmp:
## Summary

Resolves issue #35: `ReferenceError: CharacterData is not defined`

Adds the `CharacterData` DOM interface that was missing from the JS runtime shim. This interface is used internally by libraries like sweetalert.min.js and other DOM manipulation code.

## Changes

- **Added** `CharacterData` class extending `Node` with:
  - Properties: `data`, `length`
  - Methods: `substringData()`, `appendData()`, `insertData()`, `deleteData()`, `replaceData()`
- **Added** `Text` and `Comment` classes extending `CharacterData`
- **Updated** `_wrap` function to use correct class based on `nodeType` (3=Text, 8=Comment)
- **Exposed** `CharacterData`, `Text`, `Comment` as globals in `window`

## Verification

```bash
# Before fix: CharacterData was undefined
# After fix:
$ obscura fetch https://example.com --eval "typeof CharacterData"
function

$ obscura fetch https://example.com --eval "document.createTextNode('hello').data"
hello
```

## Testing

- sweetalert.min.js now loads without CharacterData errors
- All CharacterData methods work correctly
- Backward compatible with existing Node usage

## Related

- Fixes #35

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -243,8 +243,7 @@ impl Page {
             .chain(async_scripts.into_iter())
             .collect();
 
-        let mut resolved: Vec<(usize, String)> = Vec::new();
-        let mut fetch_tasks: Vec<(usize, String)> = Vec::new();
+        let client = self.http_client.clone();
 
         for (i, script) in all_to_execute.iter().enumerate() {
             if let Some(src_url) = &script.src {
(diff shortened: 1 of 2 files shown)


h4ckf0r0day:
Going with #40 because it's a more thorough fix that adds proper Comment node support in the Rust DOM layer and includes splitText(). Thanks for the contribution though!

PR #60, opened 2026-04-28
Persist cookies to JSON across runs (--cookie-store)

gte619n:
## Summary

Adds an opt-in `--cookie-store <PATH>` flag to `obscura serve` that loads cookies from a JSON file on startup and saves them back on graceful shutdown (Ctrl+C / SIGINT). The file is a versioned envelope so future format changes can fail loudly instead of silently corrupting state.

```bash
obscura serve --port 9223 --cookie-store ~/.obscura/cookies.json
# ... clients log in via CDP, cookies accumulate ...
# Ctrl+C saves to disk
# next launch with the same flag restores them
```

I'd been evaluating Obscura for some Playwright automation that depends on persistent login state across runs. Without on-disk cookies every restart loses the session, which makes any logged-in workflow impractical. This is the smallest change I could land to fix that. `localStorage`/`IndexedDB` are deliberately not in scope here.

## What's in the PR

| Crate | Change |
|---|---|
| `obscura-net` | Derive `serde` on `CookieEntry`. Add `CookieJar::save_to_path` (atomic via `.tmp` + rename, drops expired, sorted for stable diffs) and `CookieJar::load_from_path` (missing file is `Ok(0)`, rejects unknown versions, skips already-expired entries). |
| `obscura-browser` | New `BrowserContext::with_options_and_jar` so a caller-supplied `Arc<CookieJar>` can be threaded in. Existing `with_options` is unchanged (it forwards). |
| `obscura-cdp` | New `CdpContext::new_with_jar` + `start_with_full_options(port, proxy, stealth, cookie_store)`. The previous `start_with_options` signature is preserved and forwards with `cookie_store: None`. The accept loop is wrapped in `tokio::select!` against `tokio::signal::ctrl_c()` so shutdown can flush the jar. |
| `obscura-cli` | `serve --cookie-store PATH` flag. With `--workers > 1` the flag is ignored and a warning is logged (workers would race the same file). |

## File format

```json
{
  "version": 1,
  "cookies": [
    {
      "name": "session",
      "value": "abc123",
      "path": "/",
      "domain": "example.com",
      "secure": true,
      "http_only": true,
      "expires": null,
      "same_site": "Lax"
    }
  ]
}
```

## Tests

5 new round-trip tests in `obscura-net::cookies::tests` covering save/load round-trip, missing-file load, expired-entry filtering on save, version-mismatch rejection, and overwrite-via-rename safety. All 99 workspace tests pass.

## End-to-end verification

```
$ obscura serve --port 9223 --cookie-store /tmp/c.json
# CDP: Storage.setCookies for example.com and test.local
# Ctrl+C → "Saved cookies to /tmp/c.json"
$ obscura serve --port 9223 --cookie-store /tmp/c.json
INFO Loaded 2 cookies from /tmp/c.json
# CDP: Storage.getCookies → both cookies returned, httpOnly/secure flags preserved
```

## Out of scope (called out as follow-ups)

- `localStorage` / `sessionStorage` / `IndexedDB` — still in-memory; would need separate work and a richer storage backend.
- Save-on-mutation / periodic auto-save — current behavior only flushes on `SIGINT`; a `kill -9` loses cookies set since startup. Easy to add a debounced auto-save if desired.
- Per-CDP-context isolation — Obscura currently shares one `default_context` across all CDP sessions, so the jar is global. Per-context jars would be a follow-up if multi-context CDP support lands.

## Test plan

- [x] `cargo test --workspace` passes (99 tests, 0 failures)
- [x] `cargo build` clean (no new warnings introduced)
- [x] Round-trip end-to-end via CDP `Storage.setCookies` → SIGINT → restart → `Storage.getCookies` returns the same cookies with `httpOnly`/`secure` flags preserved
- [x] Atomic write: kill mid-write does not leave a half-truncated file (writes go to `<path>.tmp` first, rename is atomic on POSIX)
- [x] Missing-file startup is silent (`Ok(0)` log line, server starts normally)
- [x] Backward compatible: `start_with_options` signature unchanged; existing callers and tests untouched

### DIFF
--- a/crates/obscura-browser/src/context.rs
+++ b/crates/obscura-browser/src/context.rs
@@ -30,7 +30,18 @@ impl BrowserContext {
     }
 
     pub fn with_options(id: String, proxy_url: Option<String>, stealth: bool) -> Self {
-        let cookie_jar = Arc::new(CookieJar::new());
+        Self::with_options_and_jar(id, proxy_url, stealth, Arc::new(CookieJar::new()))
+    }
+
+    /// Build a context using a caller-supplied cookie jar. Used to share a
+    /// pre-populated jar (e.g. loaded from disk) across the CDP server and
+    /// any shutdown hooks that need to persist it.
+    pub fn with_options_and_jar(
+        id: String,
+        proxy_url: Option<String>,
+        stealth: bool,
+        cookie_jar: Arc<CookieJar>,
+    ) -> Self {
         let mut client = ObscuraHttpClient::with_options(
             cookie_jar.clone(),
             proxy_url.as_deref(),
--- a/crates/obscura-cdp/src/dispatch.rs
+++ b/crates/obscura-cdp/src/dispatch.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 
 use obscura_browser::{BrowserContext, Page};
 use obscura_js::ops::{InterceptResolution, InterceptedRequest};
+use obscura_net::CookieJar;
 use serde_json::json;
 
 use crate::domains;
@@ -39,10 +40,18 @@ impl CdpContext {
     }
 
     pub fn new_with_options(proxy: Option<String>, stealth: bool) -> Self {
-        let default_context = Arc::new(BrowserContext::with_options(
+        Self::new_with_jar(proxy, stealth, Arc::new(CookieJar::new()))
+    }
+
+    /// Build a context with a caller-supplied cookie jar so the server can
+    /// pre-load persisted cookies and keep a handle for graceful save on
+    /// shutdown.
+    pub fn new_with_jar(proxy: Option<String>, stealth: bool, cookie_jar: Arc<CookieJar>) -> Self {
+        let default_context = Arc::new(BrowserContext::with_options_and_jar(
             "default".to_string(),
             proxy,
             stealth,
+            cookie_jar,
         ));
         CdpContext {
             pages: Vec::new(),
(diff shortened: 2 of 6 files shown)


Qodo-Free-For-OSS:
Hi, start_with_full_options documents saving cookies on Ctrl+C / SIGTERM but only waits on tokio::signal::ctrl_c(), so SIGTERM shutdowns won’t trigger persistence.

**How to fix:** Handle SIGTERM (Unix) too

> ### Issue description
> `start_with_full_options` says it saves cookies on Ctrl+C / SIGTERM, but it only listens for `tokio::signal::ctrl_c()`. This means SIGTERM-based graceful shutdowns (common in service managers/containers) can exit without persisting cookies.
> 
> ### Issue Context
> The doc comment explicitly promises SIGTERM support.
> 
> ### Fix Focus Areas
> - crates/obscura-cdp/src/server.rs[41-45]
> - crates/obscura-cdp/src/server.rs[100-105]
> 
> ### Implementation notes
> - On Unix, add a `tokio::signal::unix::signal(SignalKind::terminate())` branch in the `select!` alongside `ctrl_c()`.
> - On non-Unix platforms, either adjust the doc comment to remove SIGTERM or add the closest equivalent shutdown signal handling if available.

---
*Spotted by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source projects.*

h4ckf0r0day:
Thanks for the contribution! There are now conflicts in 4 files (context.rs, dispatch.rs, server.rs, main.rs) after PR #76 merged. Could you rebase onto main?

h4ckf0r0day:
Hey could you resolve conflicts?

h4ckf0r0day:
Thanks for this! Could you resolve conflicts?

SGavrl:
closing in favor of #114. that one covers the same persistence plus fixes the http cookie header filtering bug. thanks for the work,

PR #73, opened 2026-04-30
fix: implement CharacterData DOM API for jQuery DataTables compatibility

### LINKED ISSUE #72: Page renders empty: jQuery DataTables crashes due to missing CharacterData DOM API
Page fails to render: qualifiedremodeler.com/top-500-2025

  Tried Obscura v0.1.1 (macOS arm64 prebuilt) on this page and the main content never appears. Body ends up basically empty (document.body.innerText is just
  "2026"), no <table> and no <tr> in the DOM, even with --wait 15 --wait-until networkidle --stealth.

  Repro:
  obscura fetch https://www.qualifiedremodeler.com/top-500-2025/ \
    --eval "JSON.stringify({tables: document.querySelectorAll('table').length, rows: document.querySelectorAll('tbody tr').length})" \
    --wait 15 --stealth
   -> {"tables":0,"rows":0}

  In simple words: the page uses an old-style jQuery plugin (jQuery DataTables) to build the ranking table after the page loads. That plugin crashes inside
  Obscura's JS environment, so the table never gets built and no data ever shows up.

  Errors logged by Obscura during the fetch:

  1. ReferenceError: CharacterData is not defined
  The page's scripts reference the standard DOM CharacterData interface, which doesn't exist in Obscura's DOM.
  2. TypeError: Cannot read properties of undefined (reading 'onsubmit') — early in jQuery init.
  3. TypeError: Cannot read properties of undefined (reading 'length') thrown from S.fn.init.DataTable — DataTables blows up on init, which is what kills the
  table.
  4. TypeError: Cannot read properties of null (reading 'clientWidth') — Google Publisher Tag also fails (probably because the ad slot element it expects isn't
  laid out).

  So the root cause looks like missing/incomplete DOM APIs in Obscura (at minimum CharacterData, plus whatever DataTables 1.x and jQuery 1.11 touch during their
   init path). The same page renders fine in headless Chrome.

  I also tried driving Obscura via Playwright over CDP (chromium.connectOverCDP) — that connection works great (nice to confirm!), but since the page's JS still
   runs inside Obscura, the same crashes happen and no data is fetched.

  Happy to provide full logs if useful.

KuaaMU:
## Problem

jQuery DataTables references the standard `CharacterData` interface, which is undefined in obscura's DOM environment. This causes `ReferenceError: CharacterData is not defined`, crashing DataTables initialization and rendering pages empty.

Additionally, `Text` and `Comment` were aliased to `Node` (globalThis.Text = Node), so they lacked `.data` and `.length` properties that CharacterData should provide.

## Fix

**2 files changed: +64/-9**

### 1. `crates/obscura-js/js/bootstrap.js`

- New `CharacterData` class extending `Node` with full API:
  - `data` (getter/setter) → maps to Rust `text_content`/Comment` instances by nodeType
- `Document.createComment()` uses new `Comment` class + `create_comment_node` op

### 2. `crates/obscura-js/src/ops.rs`

- New `create_comment_node` op: creates `NodeData::Comment` nodes
- `set_text_content` now handles both `Text` and `Comment` node mutations

## Verification

After fix:
- `typeof CharacterData` → `"function"` (no more ReferenceError)
- `CharacterData.prototype.data` → readable/writable on text/comment nodes
- jQuery DataTables initializes successfully, page tables render correctly

Fixes #72

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -345,6 +345,56 @@ class Node {
   isSameNode(other) { return other && this._nid === other._nid; }
   addEventListener() {} removeEventListener() {} dispatchEvent() { return true; }
 }
+class CharacterData extends Node {
+  get data() {
+    return _domParse("text_content", this._nid) ?? "";
+  }
+  set data(v) {
+    _dom("set_text_content", this._nid, String(v ?? ""));
+  }
+  get length() { return this.data.length; }
+  substringData(offset, count) {
+    return this.data.substring(offset, offset + count);
+  }
+  appendData(s) { this.data += s; }
+  insertData(offset, s) {
+    const d = this.data;
+    this.data = d.slice(0, offset) + s + d.slice(offset);
+  }
+  deleteData(offset, count) {
+    const d = this.data;
+    this.data = d.slice(0, offset) + d.slice(offset + count);
+  }
+  replaceData(offset, count, s) {
+    const d = this.data;
+    this.data = d.slice(0, offset) + s + d.slice(offset + count);
+  }
+}
+
+class Text extends CharacterData {
+  get nodeName() { return "#text"; }
+  get nodeType() { return 3; }
+  get wholeText() { return this.data; }
+  splitText(offset) {
+    const d = this.data;
+    const tail = d.substring(offset);
+    this.data = d.substring(0, offset);
+    const newNid = +_dom("create_text_node", tail);
+    const parent = this.parentNode;
+    if (parent) {
+      const ref = this.nextSibling;
+      parent.insertBefore(_wrap(newNid), ref);
+    }
+    return _wrap(newNid);
+  }
+  cloneNode() { return document.createTextNode(this.data); }
+}
+
+class Comment extends CharacterData {
+  get nodeName() { return "#comment"; }
+  get nodeType() { return 8; }
+  cloneNode() { return document.createComment(this.data); }
+}
 
 class Element extends Node {
   constructor(nid) {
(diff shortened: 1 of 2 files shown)


Qodo-Free-For-OSS:
Hi, CharacterData.data reads via the "text_content" op which is implemented as descendant text aggregation; for Comment nodes that returns an empty string, so Comment.data/nodeValue/textContent are incorrect and cloneNode produces empty comments.

**How to fix:** Add per-node contents op

> ### Issue description
> `CharacterData.data` is currently backed by the `text_content` op, which returns aggregated descendant text (and ignores comment contents). For `Comment` nodes this yields `""` instead of the comment’s own contents.
> 
> ### Issue Context
> - `CharacterData.data` uses `text_content`.
> - `DomTree::text_content` intentionally aggregates only `NodeData::Text` nodes.
> - Comments store their own `contents` and must be read/written directly.
> 
> ### Fix Focus Areas
> - crates/obscura-js/js/bootstrap.js[348-397]
> - crates/obscura-js/src/ops.rs[123-155]
> 
> ### Suggested approach
> 1. Add a new dom op (e.g. `character_data` or `node_contents`) that returns `contents` for `NodeData::Text` and `NodeData::Comment` (and `null` for others).
> 2. Update `CharacterData.data` getter to call that new op.
> 3. Optionally, ensure `Node.nodeValue` uses the same per-node op for types 3 and 8 (or keep current `nodeValue` implementation but make it consistent with `CharacterData.data`).

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

PR #74, opened 2026-04-30
fix(cli): make --quiet suppress JS console errors on stderr (#56)

### LINKED ISSUE #56: Bug: --quiet does not suppress JS console errors on stderr
## Summary

When using `obscura fetch --quiet`, browser JS console errors still print to stderr in full ANSI color. The `--quiet` flag only suppresses the "Fetching..." and "Page loaded..." banners (controlled by `if !quiet` checks in `run_fetch`), but the `tracing_subscriber` logging to stderr is unaffected.

## Current behavior

```bash
obscura fetch --dump text --quiet "https://en.wikipedia.org/wiki/Example"
```

Still outputs:
```
ERROR obscura::console: Dynamic script error (https://en.wikipedia.org/w/load.php?...): Unexpected token 'set'
ERROR obscura::console: Dynamic script error (https://en.wikipedia.org/w/load.php?...): Unexpected token 'set'
```

## Root cause (from source)

In `main.rs`, the tracing subscriber is initialized globally:
```rust
let filter = if args.verbose { "debug" } else { "warn" };
tracing_subscriber::fmt()
    .with_env_filter(...)
    .with_writer(std::io::stderr)
    .init();
```

The `--quiet` flag is only passed to `run_fetch()`, not to the tracing setup. So `ERROR` level logs always go to stderr.

## Proposed fix

When `--quiet` is set, either:
1. Set the tracing filter to `"error"` (only show truly fatal errors, not console JS warnings)
2. Or redirect tracing to `std::io::sink` when quiet

```rust
let (filter, writer) = if args.verbose {
    ("debug", std::io::stderr as Box<dyn std::io::Write + Send + Sync>)
} else if args.quiet {
    ("error", Box::new(std::io::sink()) as _)
} else {
    ("warn", std::io::stderr as Box<dyn std::io::Write + Send + Sync>)
};
```

Or simpler: just elevate the filter to `"error"` when quiet.

## Environment
- Obscura version: 0.1.0
- Built from source (Rust 1.95.0)

ousamabenyounes:
## Summary

Fixes #56 — `obscura fetch --quiet` was still leaking ERROR-level JS console output to stderr because the `--quiet` flag only gated banner prints inside `run_fetch`, not the global `tracing_subscriber`.

## Root Cause

`tracing_subscriber` is initialized in `main()` before the command match is performed, so it never sees the per-subcommand `quiet` field on `Command::Fetch`. The filter was hardcoded to `\"warn\"` (or `\"debug\"` with `--verbose`), and `obscura::console` errors come through tracing at ERROR level, which `\"warn\"` keeps enabled.

Result: the user-facing fetch output (page dump via `println!`) was clean, but the noise band on stderr was identical to a non-quiet run.

## Fix

`crates/obscura-cli/src/main.rs`:

- Peek at `args.command` before initializing the subscriber and detect `Command::Fetch { quiet: true, .. }`.
- Extract the filter selection into a small pure helper `select_log_filter(verbose, quiet) -> &'static str`:
  - `verbose` → `\"debug\"` (unchanged, wins over quiet so debug diagnostics stay usable)
  - `quiet`   → `\"off\"`   (silences all tracing output, including JS console errors)
  - else     → `\"warn\"`  (current default)
- Pass the result into the existing `EnvFilter` chain. `RUST_LOG` still overrides via `try_from_default_env()`.

Net diff: +37 / −2 in one file. No behavior change for `serve` / `scrape` / non-quiet fetch.

## Verification

```bash
# Before
$ obscura fetch --dump text --quiet \"https://en.wikipedia.org/wiki/Example\"
ERROR obscura::console: Dynamic script error (https://en.wikipedia.org/w/load.php?...): Unexpected token 'set'
ERROR obscura::console: Dynamic script error (https://en.wikipedia.org/w/load.php?...): Unexpected token 'set'
<page text dump>

# After
$ obscura fetch --dump text --quiet \"https://en.wikipedia.org/wiki/Example\"
<page text dump>
```

\`-v / --verbose\` still beats \`--quiet\` so debug runs continue to surface everything.

## Test plan

- [x] \`cargo check -p obscura-cli\` passes clean (1 pre-existing warning unrelated to this PR).
- [x] \`cargo test -p obscura-cli\` — 4 new unit tests for \`select_log_filter\` covering default / verbose / quiet / verbose+quiet pass.
- [x] No regressions — pre-existing warnings unchanged, no behavior change for \`serve\` / \`scrape\` / non-quiet fetch.
- [ ] Stealth build skipped (no macOS host).

## Risk

Low. One file touched, helper is pure, tracing subscriber init is the only side effect and the change is gated behind the user explicitly passing \`--quiet\`.

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/crates/obscura-cli/src/main.rs
+++ b/crates/obscura-cli/src/main.rs
@@ -119,11 +119,22 @@ fn print_banner(port: u16) {
 "#, port);
 }
 
+fn select_log_filter(verbose: bool, quiet: bool) -> &'static str {
+    if verbose {
+        "debug"
+    } else if quiet {
+        "off"
+    } else {
+        "warn"
+    }
+}
+
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> anyhow::Result<()> {
     let args = Args::parse();
 
-    let filter = if args.verbose { "debug" } else { "warn" };
+    let quiet = matches!(&args.command, Some(Command::Fetch { quiet: true, .. }));
+    let filter = select_log_filter(args.verbose, quiet);
     tracing_subscriber::fmt()
         .with_env_filter(
             tracing_subscriber::EnvFilter::try_from_default_env()
@@ -724,4 +735,28 @@ fn dump_links(page: &Page) {
             }
         }
     });
+}
+#[cfg(test)]
+mod tests {
+    use super::select_log_filter;
+
+    #[test]
+    fn default_filter_is_warn() {
+        assert_eq!(select_log_filter(false, false), "warn");
+    }
+
+    #[test]
+    fn verbose_filter_is_debug() {
+        assert_eq!(select_log_filter(true, false), "debug");
+    }
+
+    #[test]
+    fn quiet_filter_is_off() {
+        assert_eq!(select_log_filter(false, true), "off");
+    }
+
+    #[test]
+    fn verbose_wins_over_quiet() {
+        assert_eq!(select_log_filter(true, true), "debug");
+    }
 }
--- a/crates/obscura-cli/src/main.rs
+++ b/crates/obscura-cli/src/main.rs
@@ -129,11 +129,15 @@ fn select_log_filter(verbose: bool, quiet: bool) -> &'static str {
     }
 }
 
+fn is_quiet_command(cmd: &Option<Command>) -> bool {
+    matches!(cmd, Some(Command::Fetch { quiet: true, .. }))
+}
+
 #[tokio::main(flavor = "current_thread")]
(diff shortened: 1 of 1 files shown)


ousamabenyounes:
Added integration tests addressing review feedback: the matches!() wiring in main() is now covered by 6 new tests that parse real CLI args via `Args::try_parse_from` and assert the full --quiet chain end-to-end.

`cargo test -p obscura-cli`: 10 passed; 0 failed.

Commit: 2bec4de

h4ckf0r0day:

Issue #26, opened 2026-04-25
Support Rust CDP clients chromiumoxide and headless_chrome


mirsella:
## Request

Please consider supporting common Rust Chrome DevTools Protocol clients such as `chromiumoxide` and `headless_chrome` when connecting to Obscura via `obscura serve`.

## Why

Obscura exposes a CDP endpoint and works with raw CDP messages, but these higher-level Rust clients currently fail against it. Supporting them would make Obscura much easier to use from Rust automation code without writing a custom WebSocket/CDP layer.

## Observed behavior

- `chromiumoxide` connects to `ws://127.0.0.1:<port>/devtools/browser`, then its handler fails while decoding an Obscura protocol message with: `data did not match any variant of untagged enum Message`. Pending calls are then canceled.
- `headless_chrome` connects successfully, but navigation through its high-level API fails because it sends CDP messages via `Target.sendMessageToTarget`, which Obscura currently returns as an unknown Target method.

## Expected behavior

Rust CDP clients should be able to connect to Obscura, create a tab/page, navigate to a URL, and evaluate JavaScript using their normal high-level APIs.

## Minimal scenario

1. Start Obscura with `obscura serve --port 9222`.
2. Connect from Rust using `chromiumoxide::Browser::connect("ws://127.0.0.1:9222/devtools/browser")` or `headless_chrome::Browser::connect(...)`.
3. Create a new page/tab.
4. Navigate to a URL and evaluate `document.title`.

## Notes

This is not a request for WebDriver support. Both crates are CDP-based, so this should mainly be about CDP compatibility and message/session routing behavior.

mirsella:
nice thanks ! the commit message appear to say only headless_chrome is supported now, but not chromiumoxide. but we at least have one working now !

PR #83, opened 2026-04-30
fix(cdp): implement Target.sendMessageToTarget for headless_chrome (#26)

ousamabenyounes:
## Summary

Addresses the headless_chrome leg of #26. After this change, Rust CDP clients that wrap their calls inside `Target.sendMessageToTarget` (notably `headless_chrome`, and older Puppeteer versions) can drive Obscura through their high-level APIs instead of failing with "Unknown Target method".

## Root Cause

`crates/obscura-cdp/src/domains/target.rs` had handlers for `attachToTarget` / `createTarget` / etc., but no `sendMessageToTarget`. That method is the legacy CDP wire format Chrome itself still honours: clients send their actual CDP method as a JSON-encoded string inside `params.message`, and the server is expected to dispatch the inner request and emit the response as a `Target.receivedMessageFromTarget` event (rather than as a normal response correlated by id).

`headless_chrome` connected, then sent its very first navigation through this wrapper, got the unknown-method error, and gave up.

## Fix

Intercept `Target.sendMessageToTarget` in `dispatch()` before the per-domain match and:

1. Parse the inner JSON-encoded CDP request from `params.message`
2. Recursively dispatch it using the wrapper's `sessionId` so routing lands on the right page (the future is boxed to sidestep the async-fn recursion limit)
3. Re-emit the inner response as a `Target.receivedMessageFromTarget` event with the inner payload re-serialised as a string — the contract these clients actually listen on
4. Return `{}` for the wrapper itself

No existing call paths change; clients that don't use the wrapper (Playwright, recent Puppeteer over `connect_over_cdp`) keep using the regular request/response flow.

## What this does NOT fix

The `chromiumoxide` half of #26 — `data did not match any variant of untagged enum Message` — is an event/response *shape* issue, not a routing issue. It needs a focused investigation against a chromiumoxide repro that the issue doesn't include (which exact message fails to decode? a server-emitted event? a response missing a field?). Left out of this PR rather than guessing.

## Verification

```bash
cargo test -p obscura-cdp --lib
# running 3 tests
# test dispatch::tests::send_message_to_target_unwraps_inner_call ... ok
# test dispatch::tests::send_message_to_target_rejects_invalid_message ... ok
# test dispatch::tests::unknown_domain_still_errors ... ok
# test result: ok. 3 passed; 0 failed
cargo check --workspace        # clean
```

Manual repro from the issue:

```rust
use headless_chrome::Browser;
let browser = Browser::connect("ws://127.0.0.1:9222/devtools/browser/...".into())?;
let tab = browser.new_tab()?;
tab.navigate_to("https://example.com")?;
// Before: "Unknown Target method: sendMessageToTarget" — nav fails immediately
// After:  inner Page.navigate dispatched; receivedMessageFromTarget event carries the response
```

## Test plan

- [x] New unit test wraps a `Browser.getVersion` request, asserts the wrapper succeeds and the inner response is emitted via `Target.receivedMessageFromTarget` with the inner method's `result` populated
- [x] New regression test asserts malformed inner JSON surfaces as a parse error (-32700) instead of silently swallowing the request
- [x] Existing "Unknown domain still errors" regression test kept
- [x] `cargo check --workspace` clean
- [x] Stealth build skipped — no stealth code touched

## Risk

Low-to-medium. New code path only activates for callers that send `Target.sendMessageToTarget`. The recursion is bounded by the inner message's own dispatch (and an inner `Target.sendMessageToTarget` would just box another future the same way). Boxed dispatch is required for async recursion in Rust — without it the future type would be infinite.

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/crates/obscura-cdp/src/dispatch.rs
+++ b/crates/obscura-cdp/src/dispatch.rs
@@ -121,6 +121,14 @@ impl CdpContext {
 }
 
 pub async fn dispatch(req: &CdpRequest, ctx: &mut CdpContext) -> CdpResponse {
+    // headless_chrome (and older Puppeteer) wrap every CDP call inside
+    // Target.sendMessageToTarget. Unwrap and recurse BEFORE acquiring the
+    // V8 lock — the recursive dispatch will acquire it for the inner call,
+    // and tokio Mutex is not reentrant.
+    if req.method == "Target.sendMessageToTarget" {
+        return dispatch_send_message_to_target(req, ctx).await;
+    }
+
     // Issue #19: V8 fatal abort under concurrent CDP work.
     //
     // Every CDP handler below may end up calling into a per-Page `JsRuntime`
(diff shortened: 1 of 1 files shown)


h4ckf0r0day:
Thanks for the contribution! This PR now has a small conflict in `dispatch.rs` because #80 (Audits domain) just merged and touches the same match block. Could you rebase onto main? Just keep both your `Target.sendMessageToTarget` arm and the `Audits` arm from #80.

PR #84, opened 2026-04-30
Docker support

llonchj:
A Dockerfile for obscura.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,45 @@
+FROM rust:1-slim-bookworm AS builder
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        curl \
+        ca-certificates \
+        perl \
+        make \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /build
+
+# Cache dependency compilation by copying manifests first
+COPY Cargo.toml Cargo.lock ./
+COPY crates/obscura-dom/Cargo.toml       crates/obscura-dom/Cargo.toml
+COPY crates/obscura-net/Cargo.toml       crates/obscura-net/Cargo.toml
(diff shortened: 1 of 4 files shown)


Qodo-Free-For-OSS:
Hi, `Page::execute_scripts` blocks static script URLs via `should_block_url`, but the new dynamic script loop fetches and executes URLs without checking interception/block patterns. This allows runtime-injected scripts (webpack/turbopack chunks) to bypass the same blocking policy applied to initial scripts.

**Severity:** action required | **Category:** security

**How to fix:** Apply should_block_url to dynamic

> ### Issue description
> Dynamic `<script src>` URLs queued from JS are fetched/executed without applying `Page::should_block_url`, bypassing interception/block patterns.
> 
> ### Issue Context
> Static scripts are blocked before fetch; dynamic scripts are not.
> 
> ### Fix Focus Areas
> - crates/obscura-browser/src/page.rs[253-269]
> - crates/obscura-browser/src/page.rs[391-436]
> 
> ### Suggested approach
> - Before adding a URL to `fetch_futures` (or before fetching), call `self.should_block_url(&url)` and skip/record a blocked event similar to the static-script path.
> - Consider also deduplicating by `(url,nid)` or keeping a longer-lived set across rounds to avoid re-fetching blocked/failed URLs repeatedly.

---
*Spotted by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source projects.*

SGavrl:
Thanks for the Dockerfile going to add it directly in a follow-up commit so it ships in v0.1.4. Closing this PR since the SPA chunk loading commit doesn't fully fix the #86 case I tested against (kiwi.com still produces the same "chunk path empty" errors on both main and this branch. Happy to review a focused PR on the SPA bootstrap work separately.

PR #88, opened 2026-05-01
Plumb proxy into fetch and scrape

F0Rextasy:
## Summary

  This PR makes the global `--proxy` flag work for CLI fetch and scrape flows.

  Previously, `--proxy` was accepted at the top level, but `fetch` created its browser context with `None`, so the proxy was not
  actually used. Parallel `scrape` workers also did not inherit the proxy setting.

  ## Changes

  - Pass the global proxy into `fetch` browser context.
  - Forward proxy config to `scrape` workers through `OBSCURA_PROXY`.
  - Make workers initialize `BrowserContext::with_options(..., proxy, false)`.
  - Add README examples for proxied `fetch` and `scrape`.

  ## Verification

  - `cargo check -p obscura-cli` passed.
  - `obscura fetch https://example.com --eval 'document.title' --quiet` returns `Example Domain`.
  - `obscura --proxy http://127.0.0.1:9 fetch https://example.com ...` fails with a network error, confirming the proxy is actual
  ly applied.
  - `obscura --proxy http://127.0.0.1:9 scrape https://example.com --concurrency 1 --format json --timeout 3` returns a worker ne
  twork error, confirming scrape workers inherit the proxy.

  ## Notes

  `cargo fmt --check` was not run because `rustfmt` is not installed in my local toolchain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,9 @@ obscura fetch https://news.ycombinator.com --dump html
 # Write dump or eval output to a file
 obscura fetch https://example.com --dump text --output page.txt
 
+# Fetch through an HTTP or SOCKS proxy
+obscura --proxy socks5://127.0.0.1:1080 fetch https://example.com --dump text
+
 # Wait for dynamic content
 obscura fetch https://example.com --wait-until networkidle0
 
@@ -122,6 +125,9 @@ obscura scrape url1 url2 url3 ... \
 
 # Suppress scrape progress on stderr for script-friendly output
 obscura scrape https://example.com --quiet --format json
+
+# Scrape workers inherit the global proxy
+obscura --proxy http://127.0.0.1:8080 scrape https://example.com https://news.ycombinator.com
 ```
 
 ## Puppeteer / Playwright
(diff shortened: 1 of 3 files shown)


Qodo-Free-For-OSS:
Hi, When running `obscura --proxy … serve`, the `Serve` branch uses only the subcommand `proxy` field and ignores the global `Args::proxy`, so the global flag does not apply to the CDP server as intended.

**How to fix:** Merge global and serve proxy

> ### Issue description
> `serve` still ignores the top-level/global `--proxy` flag. Users can run `obscura --proxy http://… serve`, but the code passes only `Command::Serve::proxy` into `start_with_options`, so the proxy is not applied.
> 
> ### Issue Context
> - `Args::proxy` is marked `global = true`.
> - `Command::Serve` also declares its own `proxy`.
> - `Fetch`/`Scrape` already use the extracted `global_proxy`.
> 
> ### Fix Focus Areas
> - Prefer subcommand proxy when provided, else fallback to global proxy.
> - Apply that merged value in both the single-worker `start_with_options(...)` path and the `run_multi_worker_serve(...)` path.
> 
> - crates/obscura-cli/src/main.rs[132-170]

---

h4ckf0r0day:

h4ckf0r0day:
Hey could you resolve conflicts?

PR #90, opened 2026-05-01
Add scrape input-file option

F0Rextasy:
Adds --input-file support to scrape so URLs can be read from a file instead of being passed only through CLI arguments.

  Verification:
  - cargo check -p obscura-cli

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,9 @@ obscura scrape url1 url2 url3 ... \
 
 # Suppress scrape progress on stderr for script-friendly output
 obscura scrape https://example.com --quiet --format json
+
+# Read URL lists from a file. Blank lines and # comments are ignored.
+obscura scrape --input-file urls.txt --concurrency 25
 ```
 
 ## Puppeteer / Playwright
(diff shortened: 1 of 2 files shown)


Qodo-Free-For-OSS:
Hi, `collect_scrape_urls` reads the entire input file/stdin into memory before parsing, which can cause large memory spikes and delay startup for big URL lists. This is avoidable because the logic is line-oriented and can be streamed.

**Severity:** remediation recommended | **Category:** performance

**How to fix:** Stream lines via BufRead

> ### Issue description
> `collect_scrape_urls` buffers the entire input file/stdin into a `String` before parsing URLs, which scales poorly for large URL lists.
> 
> ### Issue Context
> The parsing is already line-based and can be done incrementally.
> 
> ### Fix Focus Areas
> - crates/obscura-cli/src/main.rs[182-202]
> 
> ### Suggested change
> - For file input: `let file = std::fs::File::open(&path)?; let reader = std::io::BufReader::new(file);` then iterate `for line in reader.lines()`.
> - For stdin (`-`): use `std::io::stdin().lock()` with `BufRead::lines()`.
> - Keep the existing trimming/comment filtering and `urls.is_empty()` validation.

---

h4ckf0r0day:
Thanks for this! Could you resolve conflicts?

SGavrl:
Upon further review we decided this doesn't fit our current scope. Thank you very much for your work.

PR #95, opened 2026-05-01
Add fetch --output

F0Rextasy:
Adds --output support to fetch so response content can be written directly to a file.

  Verification:
  - cargo check -p obscura-cli

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,9 @@ obscura fetch https://example.com --dump links
 # Render JavaScript and dump HTML
 obscura fetch https://news.ycombinator.com --dump html
 
+# Write dump or eval output to a file
+obscura fetch https://example.com --dump text --output page.txt
+
 # Wait for dynamic content
 obscura fetch https://example.com --wait-until networkidle0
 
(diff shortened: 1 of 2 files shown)


Qodo-Free-For-OSS:
Hi, `write_or_print` uses synchronous `std::fs::write` inside the async `fetch` flow, which blocks the current-thread Tokio runtime and can stall the CLI during large outputs.

**Severity:** action required | **Category:** performance

**How to fix:** Use async or blocking pool

> ### Issue description
> `write_or_print` performs `std::fs::write` on the Tokio `current_thread` runtime, which can block the entire executor.
> 
> ### Issue Context
> `run_fetch` is async and can produce large outputs (HTML/text). With `--output`, the write should not block the runtime thread.
> 
> ### Fix Focus Areas
> - crates/obscura-cli/src/main.rs[122-123]
> - crates/obscura-cli/src/main.rs[301-370]
> 
> ### Implementation notes
> - Prefer making `write_or_print` async and using `tokio::fs::write(path, content).await?`.
> - Alternatively, keep it sync but wrap `std::fs::write` in `tokio::task::spawn_blocking` and `await` it from `run_fetch`.

---

h4ckf0r0day:

PR #132, opened 2026-05-10
Add public X timeline scraping command

F0Rextasy:
Add an obscura x command for low-latency public timeline monitoring without XCancel. The command defaults to a Nitter-compatible source because X logged-out web GraphQL can return ranked highlights instead of latest posts, while keeping x-web available as an experimental source.

Document the RSS-latency motivation and include the 78-handle watchlist benchmark in README: 78 requested, 74 returned, 1.87s wall time, 0.64s user CPU, 0.10s system CPU, 39% CPU share, and 36,092 KB max resident memory. This gives maintainers concrete CPU/RAM numbers for the high-frequency watchlist use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,19 @@ obscura scrape url1 url2 url3 ... \
 obscura scrape https://example.com --quiet --format json
 ```
 
+### Scrape public X timelines
+
+```bash
+# Read latest public posts through a Nitter-compatible timeline endpoint
+obscura x elonmusk sama --limit 3 --concurrency 25 --format json
+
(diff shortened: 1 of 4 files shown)


h4ckf0r0day:
Thanks for putting this together!

Closing as out of scope though. Site-specific scraping commands (X, LinkedIn, Reddit, etc.) belong in user-side scripts and SDKs, not in obscura's CLI. Obscura is a generic engine and adding a per-site command for one social network would invite the same for every other, which isn't the direction.

Issue #146, opened 2026-05-13
Inline <script> tags don't execute on page load — SPAs never bootstrap


Ismail-Mirza:
# Obscura — CDP gaps and inline-`<script>` execution issues

Draft of an issue for https://github.com/h4ckf0r0day/obscura/issues. Three
distinct problems surfaced while wiring Obscura into a Playwright-based
Facebook Ads Library scraper via `connect_over_cdp`. Listed worst-first.

## Environment

| Field | Value |
|---|---|
| Obscura | v0.1.0 (from `obscura serve` banner) |
| Binary sha256 prefix | `bd6763c2f269c076…` |
| Container base | `debian:12-slim` (`docker/obscura.Dockerfile`) |
| Driver | Playwright 1.59.1 (Python) over CDP |
| Repro env | docker-compose; api → obscura over a docker network |

---

## 1. Page-loaded inline `<script>` tags don't execute (highest impact)

`page.evaluate(...)` works — a JS execution context exists and round-trips
values correctly. But `<script>` tags embedded in HTML served by
`page.goto(...)` are not run. Pages that depend on their own bootstrap
scripts (i.e. anything React/Relay/Vue/Angular) render only the SSR
fallback or the initial static HTML.

### Repro

```python
import asyncio
from playwright.async_api import async_playwright

# A minimal page that proves inline <script> never fires:
URL = "https://www.example.com"  # any page works; FB shows it dramatically

async def main():
    async with async_playwright() as pw:
        b = await pw.chromium.connect_over_cdp("http://localhost:9222")
        ctx = await b.new_context()
        page = await ctx.new_page()
        await page.goto(URL, wait_until="domcontentloaded")
        # Ad-hoc JS works:
        print(await page.evaluate("1 + 2"))                       # → 3
        print(await page.evaluate("typeof navigator.webdriver"))  # → 'undefined'
        await ctx.close(); await b.close()

asyncio.run(main())
```

### Real-world symptom (Facebook Ads Library)

Facebook serves a 481-byte stub when it doesn't trust a browser:

```html
<script>
  function executeChallenge() {
    fetch('/__rd_verify_<token>?challenge=3', {method:'POST'})
      .finally(() => window.location.reload());
  }
  if (document.readyState !== 'loading') executeChallenge();
  else document.addEventListener('DOMContentLoaded', executeChallenge);
</script>
```

**Expected:** the inline `<script>` runs, fetches the verify URL, sets the
[...]

PR #150, opened 2026-05-13
fix(js): drive document.readyState through loading → interactive → complete (#137)

### LINKED ISSUE #137: Deferred scripts execute after DOMContentLoaded, causing incorrect lifecycle ordering and X.com failedScript fallback
# Deferred scripts execute after DOMContentLoaded, causing incorrect lifecycle ordering and X.com failedScript fallback

## Summary

Obscura appears to fire `DOMContentLoaded` before parser-discovered deferred scripts have finished executing.

In standard browser behavior, classic deferred scripts should execute after the document has been parsed but before `DOMContentLoaded` fires. `DOMContentLoaded` should fire only after deferred scripts and module scripts have downloaded and executed.

This lifecycle ordering issue breaks X.com login bootstrap logic.

On `https://x.com/i/flow/login`, Obscura eventually reaches this script-loaded state:

```json
{
  "runtime": true,
  "vendor": true,
  "main": true
}
```

However, the page is already stuck in X.com's fallback UI:

```text
failedScript + Try again
```

This suggests that X's fallback logic observes the scripts as not ready at the `DOMContentLoaded` / ready callback timing, commits the fallback page, and only later do the deferred scripts finish executing.

## Environment

* OS: macOS
* Obscura version:
  * CLI does not expose `--version`
  * CDP `/json/version` reports `Browser: Obscura/0.1.0`
* Modes tested:
  * `obscura fetch`
  * `obscura serve` + Puppeteer
* Target URL:
  * `https://x.com/i/flow/login`

## Confirmed facts

* The fallback reproduces with `obscura fetch "https://x.com/i/flow/login"` alone.
* Therefore this is not a Puppeteer selector issue.
* It also reproduces before any login automation logic.
* Raw X HTML contains app scripts:
  * 9 `<script>` tags
  * around 30 `abs.twimg.com` references
* X script state eventually becomes:

```json
{
  "runtime": true,
  "vendor": true,
  "main": true
}
```

* But the DOM remains in fallback state:
  * `input[name="failedScript"]`
  * `Try again` button
* Clicking `Try again` repeats the same fallback.
* `document.getElementsByName` is also broken in Obscura, but after applying a stronger shim for that issue, X still falls back. This makes the defer / DOMContentLoaded ordering issue the primary blocker.

### File: `obscura_defer_order_probe.html`

```html
<!DOCTYPE html>
<html>
  <body>
    <script>
      window.__SCRIPTS_LOADED__ = { vendor: false };
      function onReady() {
        document.body.setAttribute('data-ready-sees-vendor', String(window.__SCRIPTS_LOADED__.vendor));
      }
      if (document.readyState === 'loading') {
        document.addEventListener('DOMContentLoaded', onReady);
      } else {
        onReady();
      }
    </script>
    <script defer src="file:///Users/thirdtype/.hermes/scratch/obscura_defer_order_probe_vendor.js"></script>
  </body>
</html>
```

### File: `obscura_defer_order_probe_vendor.js`

```js
window.__SCRIPTS_LOADED__.vendor = true;
document.body.setAttribute('data-vendor-executed', 'true');
```

### Run

```bash
obscura fetch "file:///Users/thirdtype/.hermes/scratch/obscura_defer_order_probe.html" \
  --eval "JSON.stringify({
    readyState: document.readyState,
    seesVendorAtReady: document.body.getAttribute('data-ready-sees-vendor'),
    vendorExecuted: document.body.getAttribute('data-vendor-executed')
  }, null, 2)"
```

## Actual result in Obscura

```json
{
  "readyState": "complete",
  "seesVendorAtReady": "false",
  "vendorExecuted": "true"
}
```

This means:

* The deferred script eventually executes.
* But it is not visible when the `DOMContentLoaded` callback runs.
* Therefore `DOMContentLoaded` appears to fire too early relative to deferred script execution.

## Expected result

The deferred script should execute before `DOMContentLoaded`.

Expected shape:

```json
{
  "seesVendorAtReady": "true",
  "vendorExecuted": "true"
}
```

Expected ordering should be equivalent to:

```json
[
  "inline-before-defer",
  "inline-after-defer-tag",
  "defer-vendor-executed",
  "DOMContentLoaded:vendor=true",
  "load:vendor=true"
]
```

## X.com reproduction

```bash
obscura fetch "https://x.com/i/flow/login" \
  --eval "JSON.stringify({
    url: location.href,
    title: document.title,
    failedScript: !!document.querySelector('input[name=failedScript]'),
    scriptState: window.__SCRIPTS_LOADED__ || null,
    inputs: [...document.querySelectorAll('input')].map(i => ({
      type: i.type,
      name: i.name,
      autocomplete: i.autocomplete
    })),
    buttons: [...document.querySelectorAll('button')].map(b => b.innerText),
    body: document.body?.innerText?.slice(0, 500)
  }, null, 2)"
```

Observed state:

```json
{
  "url": "https://x.com/i/flow/login",
  "title": "",
  "failedScript": true,
  "scriptState": {
    "runtime": true,
    "vendor": true,
    "main": true
  },
  "inputs": [
    {
      "type": "hidden",
      "name": "failedScript"
    }
  ],
  "buttons": [
    "Try again"
  ]
}
```

Observed runtime warning during fetch:

```text
WARN obscura_browser::page: Inline script error: JS error: TypeError: document.getElementsByName is not a function
    at onReady (<script>:15:39)
    at <script>:25:5
    at <script>:27:3
```

## Secondary issue: document.getElementsByName

There is also a separate DOM compatibility issue:

```js
document.getElementsByName
```

throws:

```text
TypeError: document.getElementsByName is not a function
```

Minimal probe confirmed that `document.getElementsByName('probe')` fails in Obscura.

A stronger shim can work around this in a local probe:

```json
{
  "attr": null,
  "typeofFn": "function",
  "count": 1
}
```

After applying this workaround, X's script state reaches:

```json
{
  "runtime": true,
  "vendor": true,
  "main": true
}
```

However, X still falls back. Therefore `getElementsByName` is a real bug, but it is not the primary remaining blocker.

The primary blocker appears to be deferred script / DOMContentLoaded lifecycle ordering.

## Additional update

I also tried page-level workarounds using `evaluateOnNewDocument`:

* wrapping `Document.prototype.addEventListener` / `Window.prototype.addEventListener` for `DOMContentLoaded`
* overriding `Document.prototype.readyState` until `window.__SCRIPTS_LOADED__` reaches runtime/vendor/main

However, on X.com:

```json
{
  "scriptState": {
    "runtime": true,
    "vendor": true,
    "main": true
  },
  "failedScript": true,
  "dclShim": {
    "wrappedListeners": 0
  },
  "readyStateShim": {
    "active": false
  }
}
```

This suggests X's fallback path is not going through page-level `DOMContentLoaded` listeners or the `Document.prototype.readyState` getter that can be shimmed from userland. The issue appears to be in Obscura's internal document / script lifecycle ordering.

## Why this matters

Sites that rely on deferred app bundles expect parser-discovered deferred scripts to execute before `DOMContentLoaded`.

If `DOMContentLoaded` fires too early, bootstrap / fallback code can incorrectly conclude that required chunks failed to load.

This appears to be what happens on X.com login:

1. X app scripts eventually execute.
2. `window.__SCRIPTS_LOADED__` eventually becomes:

```json
{
  "runtime": true,
  "vendor": true,
  "main": true
}
```

3. But fallback is already committed.
4. The page remains stuck on `failedScript + Try again`.

## Local diagnostic files

The following local files were used to isolate the issue:

* `scripts/x_login.mjs`
  * X fallback / `window.__SCRIPTS_LOADED__` / retry instrumentation
* `scripts/obscura_probe_shim_test.mjs`
  * `document.getElementsByName` shim validation
* `scripts/obscura_defer_shim_test.mjs`
  * defer / lifecycle experiment
* `scratch/obscura_defer_order_probe.html`
  * minimal defer ordering probe
* `scratch/obscura_defer_order_probe_vendor.js`
  * deferred vendor script for probe
* `/tmp/x_login_obscura.html`
  * raw X login HTML captured via Obscura
* `debug_x_login_error.png`
  * fallback screen capture

## Suggested fix direction

The lifecycle order should be adjusted so that parser-discovered deferred classic scripts are executed before `DOMContentLoaded`.

Expected high-level order:

```text
HTML parse complete
→ execute pending parser-discovered deferred classic scripts in document order
→ update readyState to interactive, if implemented
→ fire DOMContentLoaded
→ later fire load / complete
```

Current observed behavior appears closer to:

```text
HTML parse complete
→ DOMContentLoaded fires
→ fallback code runs
→ deferred scripts execute later
→ page remains stuck in fallback
```

ousamabenyounes:
## Summary

Fixes [#137](https://github.com/h4ckf0r0day/obscura/issues/137).
`document.readyState` was hardcoded to `"complete"` in `bootstrap.js`, so
inline scripts running during page bootstrap that follow the canonical
[whatwg defer-aware guard](https://html.spec.whatwg.org/multipage/dom.html#current-document-readiness)

```js
if (document.readyState === 'loading') {
  document.addEventListener('DOMContentLoaded', onReady);
} else {
  onReady();
}
```

never attached a `DOMContentLoaded` listener — they hit the else branch
and called `onReady()` *before* any deferred script had set its state.

## Root Cause

`crates/obscura-js/js/bootstrap.js:789` returned a constant:

```js
get readyState() { return "complete"; }
```

So:

- X.com observes `__SCRIPTS_LOADED__.vendor === false` in its
  `DOMContentLoaded`-equivalent and renders the `failedScript + Try again`
  fallback (the symptom #137 ships as the failing case).
- Facebook's 481-byte bot-challenge stub (cf. #146) hits the else branch
  and tries to call `executeChallenge` before `fetch` is set up.
- jQuery's `$(document).ready(...)` fires synchronously instead of being
  queued.

The defer-vs-DOMContentLoaded *execution order* in `page.rs` is already
spec-correct (regular → deferred → async → DOMContentLoaded). The
missing piece is making that order *observable* to user scripts, which
they do by reading `document.readyState`.

## Fix

- `bootstrap.js`: track `readyState` in a module-scope variable
  (`__obscura_ready_state`, defaulting to `"loading"`) and expose a
  validated setter `globalThis.__obscura_set_ready_state(state)` that
  accepts only the three spec values.
- `page.rs`: thread the state transitions through the existing
  `<load-events>` execute_script:

  ```js
  __obscura_set_ready_state('interactive');
  document.dispatchEvent(new Event('DOMContentLoaded'));
  __obscura_set_ready_state('complete');
  // window.onload, window 'load' event
  ```

  `interactive` is set immediately before `DOMContentLoaded` fires;
  `complete` immediately before `load`, matching the HTML standard.

## Verification

```js
// Before
> document.readyState
"complete"   // wrong; bypasses all defer-aware guards

// After (during script execution)
> document.readyState
"loading"

// After (inside DOMContentLoaded handler)
> document.readyState
"interactive"

// After (inside load handler)
> document.readyState
"complete"
```

## Test plan

- [x] `ready_state_defaults_to_loading` — freshly constructed runtime returns `"loading"`
- [x] `ready_state_transitions_to_interactive` — explicit transition observable
- [x] `ready_state_transitions_to_complete` — explicit transition observable
- [x] `ready_state_setter_rejects_unknown_values` — invalid input must be a no-op
- [x] RED-then-GREEN: 3 of the 4 new tests fail when `bootstrap.js` is reverted to the pre-fix state (the fourth is a backwards-compat guard test that passes both ways)
- [x] `cargo test -p obscura-js --no-default-features` — 67 passed (up from 63)
- [x] `cargo check --workspace --no-default-features` clean

Stealth build skipped per repo rules.

## Risk

Low/medium. Behaviour change: `readyState` was always `"complete"` and is
now phase-correct. Scripts that wrote

```js
if (document.readyState === 'complete') doNow();
else window.addEventListener('load', doNow);
```

used to take the first branch synchronously; they now take the second
branch and wait for the `load` event that page.rs already dispatches —
net behaviour for those scripts is unchanged. Scripts that use the
`document.readyState === 'loading'` guard (the majority of bootstrap
shims) are the ones that move from broken to working.

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -352,9 +352,16 @@ impl Page {
         }
 
         if let Some(js) = &mut self.js {
+            // Drive document.readyState through the spec-mandated states
+            // before dispatching the corresponding events (#137). Until the
+            // probe page can observe the transitions on its DOMContentLoaded
+            // listener, X.com (and any framework with a "did the deferred
+            // scripts run yet?" guard) hits its `failedScript` fallback.
             let _ = js.execute_script("<load-events>",
-                "if (typeof window.onload === 'function') { try { window.onload(); } catch(e) {} }\n\
+                "try { globalThis.__obscura_set_ready_state('interactive'); } catch(e) {}\n\
                  try { document.dispatchEvent(new Event('DOMContentLoaded')); } catch(e) {}\n\
+                 try { globalThis.__obscura_set_ready_state('complete'); } catch(e) {}\n\
+                 if (typeof window.onload === 'function') { try { window.onload(); } catch(e) {} }\n\
                  try { window.dispatchEvent(new Event('load')); } catch(e) {}");
         }
 
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -2,6 +2,21 @@
 
 globalThis.__obscura_errors = [];
 
+// Document lifecycle state (#137). Starts at "loading" so scripts running
+// during page bootstrap see the spec-correct value and follow the
+// `addEventListener('DOMContentLoaded', ...)` branch instead of treating
+// the page as already-loaded. The Rust side drives transitions via
+// __obscura_set_ready_state(state) immediately before the corresponding
+// event is dispatched:
+//   "interactive" → just before document.dispatchEvent(DOMContentLoaded)
+//   "complete"    → just before window.dispatchEvent(load)
+let __obscura_ready_state = "loading";
+globalThis.__obscura_set_ready_state = function(state) {
+  if (state === "loading" || state === "interactive" || state === "complete") {
+    __obscura_ready_state = state;
+  }
+};
+
 globalThis.addEventListener = globalThis.addEventListener || function(){};
 globalThis.onunhandledrejection = function(e) { if (e?.preventDefault) e.preventDefault(); };
 
@@ -786,7 +801,7 @@ class Document extends Node {
   get compatMode() { return "CSS1Compat"; }
   get characterSet() { return "UTF-8"; }
   get contentType() { return "text/html"; }
-  get readyState() { return "complete"; }
+  get readyState() { return __obscura_ready_state; }
   get hidden() { return false; }
   get visibilityState() { return "visible"; }
   getElementById(id) { return _wrapEl(+_dom("get_element_by_id", id)); }
--- a/crates/obscura-js/src/runtime.rs
+++ b/crates/obscura-js/src/runtime.rs
@@ -1589,4 +1589,62 @@ mod tests {
         assert_eq!(inf, serde_json::Value::Null);
     }
 
+    // Issue #137 — document.readyState must start at "loading" so scripts
+    // running during page bootstrap follow the spec-correct branch:
+    //
+    //     if (document.readyState === 'loading') {
+    //       document.addEventListener('DOMContentLoaded', onReady);
+    //     } else {
+    //       onReady();
+    //     }
+    //
+    // Pre-fix Obscura always reported "complete", so the guard above fell
+    // into the else branch immediately — running onReady() before any
+    // deferred script had set its state. X.com's failedScript fallback and
+    // Facebook's bot-challenge stub (cf. #146) both rely on this pattern.
+    #[test]
+    fn ready_state_defaults_to_loading() {
+        let mut rt = setup_runtime("<html><body></body></html>");
+        let state = rt.evaluate("document.readyState").unwrap();
+        assert_eq!(
+            state,
+            serde_json::json!("loading"),
+            "a freshly constructed runtime must report 'loading' so guards \
+             like `if (document.readyState === 'loading')` work spec-correctly"
+        );
+    }
+
+    #[test]
+    fn ready_state_transitions_to_interactive() {
+        let mut rt = setup_runtime("<html><body></body></html>");
+        rt.evaluate("globalThis.__obscura_set_ready_state('interactive');")
+            .unwrap();
+        let state = rt.evaluate("document.readyState").unwrap();
+        assert_eq!(state, serde_json::json!("interactive"));
+    }
+
+    #[test]
+    fn ready_state_transitions_to_complete() {
+        let mut rt = setup_runtime("<html><body></body></html>");
+        rt.evaluate("globalThis.__obscura_set_ready_state('complete');")
+            .unwrap();
+        let state = rt.evaluate("document.readyState").unwrap();
+        assert_eq!(state, serde_json::json!("complete"));
+    }
+
+    #[test]
+    fn ready_state_setter_rejects_unknown_values() {
+        // Defence-in-depth: only the three spec values may be assigned. An
+        // unknown value must be a no-op (state stays at the previous value).
+        let mut rt = setup_runtime("<html><body></body></html>");
+        rt.evaluate("globalThis.__obscura_set_ready_state('garbage');")
+            .unwrap();
+        let state = rt.evaluate("document.readyState").unwrap();
+        assert_eq!(
+            state,
+            serde_json::json!("loading"),
+            "an invalid state must NOT mutate readyState"
+        );
+    }
+
 }


SGavrl:
Nice on the readyState getter, the 4 tests pass and the loading -> interactive -> complete transitions show up cleanly. but i tested the exact x.com pattern from the body end-to-end:
```

  <script>
  if (document.readyState === 'loading') {
    document.addEventListener('DOMContentLoaded', onReady);
  } else {
    onReady();
  }
  </script>
```

before this pr: onReady runs synchronously via the else branch (too early but runs) after this pr: onReady never runs
the problem is document.addEventListener is a no-op in bootstrap.js. Node has a no-op override at line ~362 that shadows the working impl, and Document re-overrides at ~866. with readyState now correctly reporting "loading", code falls into the if branch and registers a listener that's silently dropped, so the DCL dispatch from <load-events> finds nothing. Can you bundle a follow-up that makes document.addEventListener('DOMContentLoaded', cb) actually fire cb when <load-events> dispatches DCL? happy to merge once that part works too.

The PR is doing the right thing in spirit but is half a fix. Holding off until it's whole.
Also there's a merge conflict against current main in crates/obscura-js/src/runtime.rs (the new tests block collided with the #151 proxy tests added in the same spot). can you resolve that?

PR #164, opened 2026-05-18
fix: DOM/JS shim gaps — Storage global, CharacterData ChildNode, Element remove/isConnected

### LINKED ISSUE #163: DOM/JS shim gaps: Storage global, CharacterData ChildNode methods, and Element.prototype.remove incomplete
## Summary

Three small but real DOM/JS shim gaps in `v0.1.5` that surface as `TypeError` / `ReferenceError` from page JS on modern bundles (Web Storage feature-detection, Svelte 5 hydration boundaries, anything that calls `ChildNode` methods on `Text`/`Comment` nodes). All reproducible offline with `obscura fetch file://...`.

### Findings

| # | Symptom | Expected (per [DOM spec](https://dom.spec.whatwg.org/) / [Web Storage spec](https://html.spec.whatwg.org/multipage/webstorage.html)) | Observed in v0.1.5 |
|---|---|---|---|
| 1 | `typeof Storage` | `"function"` (global constructor) | `"undefined"` |
| 2 | `CharacterData.prototype.{before,after,replaceWith,remove}` | All `"function"` (`ChildNode` mixin) | All `undefined` |
| 3 | `Element.prototype.remove()` | Unparents and removes from parent's child list (`getElementById` → `null`, `isConnected` → `false`, gone from `innerHTML`) | Sets `parentNode` to `null`, but `isConnected` stays `true`, the node remains in `document.body.innerHTML` and `getElementById` keeps returning it |

## Reproduction

Platform: macOS 25.3.0, aarch64. Asset: `obscura-aarch64-macos.tar.gz` from the v0.1.5 release.
Command pattern for each repro:

```
obscura fetch --timeout 10 --wait 3 --wait-until load "file:///tmp/<repro>.html" --dump html
```

Results are written into `<pre id="r">` and read back from the rendered HTML, so console plumbing doesn't matter.

### 1. `Storage` global missing

```html
<!doctype html><html><body><pre id="r"></pre><script>
const out = (k,v) => { document.getElementById('r').textContent += k + ': ' + v + '\n'; };
out('typeof Storage', typeof Storage);
out('typeof window.Storage', typeof window.Storage);
out('typeof localStorage', typeof localStorage);
try { out('localStorage instanceof Storage', localStorage instanceof Storage); }
catch (e) { out('localStorage instanceof Storage THREW', e.name + ': ' + e.message); }
</script></body></html>
```

Output:

```
typeof Storage: undefined
typeof window.Storage: undefined
typeof localStorage: object
localStorage instanceof Storage THREW: ReferenceError: Storage is not defined
```

Expected: `typeof Storage === "function"`, and `localStorage instanceof Storage === true`. The instance object behaves correctly (`setItem` / `getItem` round-trip works), but the constructor isn't exposed as a global, which breaks the `instanceof Storage` idiom used by many consent managers, analytics SDKs, and frameworks to feature-detect Web Storage.

### 2. `CharacterData.prototype` is missing the `ChildNode` mixin

```html
<!doctype html><html><body><pre id="r"></pre><script>
const out = (k,v) => { document.getElementById('r').textContent += k + ': ' + v + '\n'; };
out('typeof Element.prototype.before',       typeof Element.prototype.before);
out('typeof Element.prototype.remove',       typeof Element.prototype.remove);
out('typeof CharacterData',                  typeof CharacterData);
out('typeof CharacterData.prototype.before', typeof CharacterData.prototype.before);
out('typeof CharacterData.prototype.remove', typeof CharacterData.prototype.remove);
const txt = document.createTextNode('hello'); document.body.appendChild(txt);
try { txt.before(document.createTextNode('x ')); out('text.before()', 'ok'); }
catch (e) { out('text.before() THREW', e.name + ': ' + e.message); }
try { txt.remove(); out('text.remove()', 'ok'); }
catch (e) { out('text.remove() THREW', e.name + ': ' + e.message); }
</script></body></html>
```

Output:

```
typeof Element.prototype.before: function
typeof Element.prototype.remove: function
typeof CharacterData: function
typeof CharacterData.prototype.before: undefined
typeof CharacterData.prototype.remove: undefined
text.before() THREW: TypeError: txt.before is not a function
text.remove() THREW: TypeError: txt.remove is not a function
```

Expected: the [`ChildNode` mixin](https://dom.spec.whatwg.org/#interface-childnode) (`before`, `after`, `replaceWith`, `remove`) is mixed into **`Element`, `CharacterData`, and `DocumentType`**. Obscura currently mixes it into `Element` only, so any code that calls these methods on `Text`, `Comment`, or `ProcessingInstruction` nodes hits `TypeError: ... is not a function`. This is common in frameworks that anchor effects/boundaries on `Comment` nodes (Svelte 5, Vue, Lit's templating, etc.).

### 3. `Element.prototype.remove()` doesn't fully detach the node

```html
<!doctype html><html><body><pre id="r"></pre><div id="x">stuff</div><script>
const out = (k,v) => { document.getElementById('r').textContent += k + ': ' + v + '\n'; };
const x = document.getElementById('x');
out('before: x.parentNode',                  x.parentNode && x.parentNode.tagName);
out('before: x.isConnected',                 x.isConnected);
out('before: getElementById',                document.getElementById('x') === x ? 'found' : 'no');
x.remove();
out('after:  x.parentNode',                  x.parentNode ? x.parentNode.tagName : 'null');
out('after:  x.isConnected',                 x.isConnected);
out('after:  getElementById',                document.getElementById('x') === null ? 'null' : 'still-here');
out('after:  body.innerHTML contains #x?',   document.body.innerHTML.includes('id="x"') ? 'yes' : 'no');
</script></body></html>
```

Output:

```
before: x.parentNode: BODY
before: x.isConnected: true
before: getElementById: found
after:  x.parentNode: null
after:  x.isConnected: true
after:  getElementById: still-here
after:  body.innerHTML contains #x?: yes
```

Expected after `x.remove()`:
- `x.parentNode === null` ✓ (already correct)
- `x.isConnected === false`
- `document.getElementById('x') === null`
- node not present in `document.body.innerHTML`

`remove()` currently severs the back-pointer (`parentNode`), but doesn't unlink the node from its parent's child list. The tree still references it, the id-map still contains it, and any serialization (`innerHTML`, `outerHTML`, CDP `DOM.getOuterHTML`) sees the "removed" node.

## Suggested fixes

1. Define `Storage` as a global constructor, set `Object.getPrototypeOf(localStorage) === Storage.prototype`, and have both `localStorage` and `sessionStorage` pass `instanceof Storage`.
2. Apply the `ChildNode` mixin (`before`, `after`, `replaceWith`, `remove`) to `CharacterData.prototype` (covers `Text`, `Comment`, `ProcessingInstruction`) and `DocumentType.prototype`. Same implementation as `Element.prototype.*`, just on the other prototype chains.
3. In `Element.prototype.remove` (and the to-be-added `CharacterData.prototype.remove`), also remove the node from its parent's child list / id-map / connection state — equivalent to `this.parentNode?.removeChild(this)` — not just the `parentNode` back-pointer.

Low-priority side note: `typeof ChildNode === "undefined"` (real browsers expose `ChildNode` as a global mixin interface object). Probably not worth a separate issue, but worth a line of code if you're already in there.

Happy to test any patches against these repros.

Null-Phnix:
## Summary

Fixes #163 — three small but real DOM/JS shim gaps that surface as `TypeError` / `ReferenceError` on modern bundles (Web Storage feature-detection, Svelte 5 hydration boundaries, anything calling `ChildNode` methods on `Text`/`Comment` nodes).

### Changes

**1. `Storage` global constructor** (`bootstrap.js`)
- Replaces the inline `_mkStore()` factory with a proper `class Storage`
- `typeof Storage === "function"` (was `"undefined"`)
- `localStorage instanceof Storage === true` (was `ReferenceError`)
- Keeps the same API — `getItem`, `setItem`, `removeItem`, `clear`, `length`, `key`

**2. `CharacterData` ChildNode mixin** (`bootstrap.js`)
- Adds `before()`, `after()`, `replaceWith()`, `remove()` to `CharacterData.prototype`
- Inherited by `Text.prototype` and `Comment.prototype`
- Follows the existing `Element.prototype` pattern exactly

**3. `Element.prototype.remove()` fixes** (`bootstrap.js` + `tree.rs`)
- **3a.** Removed the hardcoded `get isConnected() { return true; }` on `Element.prototype` that shadowed the correct `Node.prototype` implementation (which walks the parent chain per spec). Detached elements now correctly report `isConnected === false`.
- **3b.** Modified `DomTree::detach()` to clean the ID index for the detached subtree. Previously `detach()` only unlinked parent/sibling pointers — the ID index was never cleaned, so `getElementById()` kept returning removed elements. Includes a parent-guard: `detach()` skips ID cleanup for nodes with no parent, since `appendChild` calls `detach()` defensively on newly-created nodes.

### Test results

- `obscura-dom` unit tests: **28/28 pass**
- `obscura-js` unit tests: **76/76 pass**
- Manual verification of all three sub-issues against the repro cases from #163

### Notes

- The `insertBefore` argument plumbing in `_dom()` has a pre-existing limitation where the ref node argument is dropped — this affects `Element.prototype.before/after/replaceWith` equally and is not addressed here (separate infrastructure issue).
- `appendChild` does not re-populate the ID index after calling `detach()` — this is a pre-existing gap unrelated to this fix.

### DIFF
--- a/crates/obscura-dom/src/tree.rs
+++ b/crates/obscura-dom/src/tree.rs
@@ -288,6 +288,56 @@ impl DomTree {
     }
 
     pub fn detach(&self, node_id: NodeId) {
+        // Collect IDs to clean from the index before taking the mutable borrow.
+        // Only clean IDs when the node is actually attached to a parent —
+        // append_child calls detach() defensively on newly-created nodes, and
+        // we must not strip their IDs from the index before they're ever used.
+        let ids_to_clean: Vec<String> = {
+            let inner = self.inner.borrow();
+            let node_has_parent = inner.nodes.get(node_id.index())
+                .and_then(|n| n.as_ref())
+                .map(|n| n.parent.is_some())
+                .unwrap_or(false);
+            if !node_has_parent {
+                Vec::new()
+            } else {
+            let mut ids = Vec::new();
+            // Check the node itself
+            if let Some(Some(node)) = inner.nodes.get(node_id.index()) {
+                if let Some(id_val) = node.get_attribute("id") {
+                    ids.push(id_val.to_string());
+                }
+            }
+            // Walk descendants
+            let mut stack: Vec<NodeId> = Vec::new();
+            if let Some(Some(node)) = inner.nodes.get(node_id.index()) {
+                let mut child = node.first_child;
+                while let Some(cid) = child {
+                    stack.push(cid);
+                    child = inner.nodes.get(cid.index())
+                        .and_then(|n| n.as_ref())
+                        .and_then(|n| n.next_sibling);
+                }
+            }
+            while let Some(current) = stack.pop() {
+                if let Some(Some(node)) = inner.nodes.get(current.index()) {
+                    if let Some(id_val) = node.get_attribute("id") {
+                        ids.push(id_val.to_string());
+                    }
+                    let mut child = node.first_child;
+                    while let Some(cid) = child {
+                        stack.push(cid);
+                        child = inner.nodes.get(cid.index())
+                            .and_then(|n| n.as_ref())
+                            .and_then(|n| n.next_sibling);
+                    }
+                }
+            }
+            ids
+            }
+        };
+
+        // Now take the mutable borrow for the structural changes.
         let mut inner = self.inner.borrow_mut();
 
         let (parent_id, prev_id, next_id) = match inner.nodes.get(node_id.index()).and_then(|n| n.as_ref()) {
@@ -319,6 +369,11 @@ impl DomTree {
             node.parent = None;
             node.prev_sibling = None;
             node.next_sibling = None;
+        }
+
+        // Clean the ID index for the detached subtree.
+        for id_str in &ids_to_clean {
+            inner.id_index.remove(id_str);
         }
     }
 
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -371,6 +371,38 @@ class CharacterData extends Node {
   }
 }
 
+// ChildNode mixin for CharacterData (Text, Comment) — spec requires
+// before(), after(), replaceWith(), remove() on Text and Comment nodes.
+CharacterData.prototype.remove = function() {
+  if (this.parentNode) this.parentNode.removeChild(this);
+};
+CharacterData.prototype.before = function(...nodes) {
+  const parent = this.parentNode;
+  if (!parent) return;
+  for (const n of nodes) {
+    if (typeof n === 'string') parent.insertBefore(document.createTextNode(n), this);
+    else parent.insertBefore(n, this);
+  }
+};
+CharacterData.prototype.after = function(...nodes) {
+  const parent = this.parentNode;
+  if (!parent) return;
+  const ref = this.nextSibling;
+  for (const n of nodes) {
+    if (typeof n === 'string') parent.insertBefore(document.createTextNode(n), ref);
+    else parent.insertBefore(n, ref);
+  }
+};
+CharacterData.prototype.replaceWith = function(...nodes) {
+  const parent = this.parentNode;
+  if (!parent) return;
+  for (const n of nodes) {
+    if (typeof n === 'string') parent.insertBefore(document.createTextNode(n), this);
+    else parent.insertBefore(n, this);
+  }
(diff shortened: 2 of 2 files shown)


SGavrl:
closing in favor of #176

PR #165, opened 2026-05-19
fix(target): emit canAccessOpener in all TargetInfo payloads

dsjacobsen:
The CDP spec marks Target.TargetInfo.canAccessOpener as a required boolean, but Obscura was omitting it from every event and response that carries a TargetInfo (setDiscoverTargets, getTargets, createTarget, attachToBrowserTarget, attachToTarget, getTargetInfo).

JS clients (Puppeteer, Playwright) tolerate the missing field via lenient deserialization. Strict Rust clients like chromiumoxide reject the event - its serde deserializer routes the payload to CdpEvent::Other, so on_target_created never runs, and the subsequent createTarget response panics with "Created target not present" when the target lookup against an empty map fails.

Obscura doesn't model window.open opener chains, so false is the correct value for every page it creates.

Spec: https://chromedevtools.github.io/devtools-protocol/tot/Target/#type-TargetInfo

### DIFF
--- a/crates/obscura-cdp/src/domains/target.rs
+++ b/crates/obscura-cdp/src/domains/target.rs
@@ -16,6 +16,7 @@ pub async fn handle(method: &str, params: &Value, ctx: &mut CdpContext) -> Resul
                         "title": "",
                         "url": "",
                         "attached": true,
+                        "canAccessOpener": false,
                         "browserContextId": "",
                     }
                 }),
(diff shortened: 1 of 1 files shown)


SGavrl:
merged into main (commit e0af9fc). added canAccessOpener: false to every TargetInfo payload across setDiscoverTargets, getTargets, createTarget, attachToBrowserTarget, attachToTarget, and getTargetInfo. all 28 integration tests + 192 unit tests passing.

PR #166, opened 2026-05-19
Fix puppeteer `exposeFunction` end-to-end: real bindings, preload timing, URL coercion

marcbachmann:
This branch fixes three independent bugs that together blocked
`page.exposeFunction(...)` (and the broader CDP early-injection
contract) under puppeteer. Each is a self-contained commit so they
can be reviewed or landed individually.

## Why

Three separate root causes silently broke `page.exposeFunction`:

1. The CDP `Runtime.addBinding` handler was a placeholder that
   installed an empty stub on `globalThis` and never forwarded calls
   back over CDP. Bindings looked registered from the client side
   but page-side calls vanished.
2. Preload scripts registered via
   `Page.addScriptToEvaluateOnNewDocument` ran *after* the page's
   own scripts, so puppeteer's `exposeFunction` wrapper (which has
   to replace `globalThis[name]` before the document's first
   `<script>` runs) never got the chance to intercept early calls.
3. The `URL` polyfill called `base.match(...)` directly, throwing
   `TypeError: base.match is not a function` for any caller passing
   a `Location` object — i.e. anything that does
   `new URL(href, window.location)`, which is common in framework
   bootstrap code.

## What changes

Three commits, smallest first:

### 1. `fix(js): coerce URL constructor input/base to string`

`crates/obscura-js/js/bootstrap.js`

The `URL` constructor now stringifies both `url` and `base` per the
WHATWG URL Standard before any regex work. Six-line change, no
logic otherwise touched.

### 2. `fix(browser,cdp): run preload scripts before page scripts`

`crates/obscura-browser/src/page.rs`,
`crates/obscura-cdp/src/domains/page.rs`,
`crates/obscura-cdp/src/server.rs`

- Adds a `preload_scripts: Vec<String>` field and a
  `set_preload_scripts(...)` setter on `Page`.
- `navigate_single` runs every registered preload source right after
  `__documentReadyState__ = 'loading'` and before parser-discovered
  scripts, using `execute_script_guarded` so large preloads get the
  5s timeout treatment.
- The two CDP entry points (`domains/page.rs::do_navigate` and
  `server.rs`'s intercept-aware navigate task) hand the current
  preload set to `Page` via `set_preload_scripts` immediately
  before navigation, replacing their old post-navigate
  `execute_preload_script` loops.

This matches Chromium's contract: preloads run on every new
document, before any of the document's scripts execute.

### 3. `feat(cdp,js): emit Runtime.bindingCalled for addBinding shims`

`crates/obscura-js/src/ops.rs`,
`crates/obscura-js/src/runtime.rs`,
`crates/obscura-browser/src/page.rs`,
`crates/obscura-cdp/src/dispatch.rs`,
`crates/obscura-cdp/src/domains/runtime.rs`

- New `op_binding_called(name, payload)` op pushes `(name, payload)`
  onto a `pending_binding_calls: Vec<(String, String)>` queue on
  `ObscuraState`. `ObscuraJsRuntime::take_pending_binding_calls`
  drains it, mirroring the existing `take_pending_navigation` shape.
- `Page::take_pending_binding_calls` is the public passthrough.
- `dispatch()` calls `drain_binding_calls(ctx)` after every CDP
  request: it walks every `Page`, takes its pending calls, and
  pushes one `Runtime.bindingCalled` event per entry into
  `ctx.pending_events`. The writer task forwards them to the client
  exactly like any other event.
- `Runtime.addBinding` now installs a real JS shim:
  `globalThis[name] = function (arg) { Deno.core.ops.op_binding_called(name, payload) }`
  where `payload` is `arg` itself if it's already a string and
  `JSON.stringify(arg)` otherwise. Matches the wire format
  puppeteer's `exposeFunction` wrapper expects
  (`Runtime.bindingCalled.params.payload` is always a string).
- The shim is registered as a preload script keyed
  `__obscura_binding__<name>` so it survives navigations (preloads
  re-run on each new document), and it's also evaluated on the
  current page so the binding works without waiting for a
  navigation.
- `Runtime.removeBinding` is now implemented and mirrors the
  registration.

Stacks naturally on commit 2 — the shim relies on preloads running
before page scripts.

## How to verify

Minimal puppeteer repro against a running `obscura serve --port
9222` (with `OBSCURA_ALLOW_PRIVATE_NETWORK=1` so the loopback test
server is reachable):

```js
import puppeteer from 'puppeteer-core'
import http from 'node:http'

const server = http.createServer((req, res) => res.end(
  `<html><body><script>
    if (typeof report === 'function') report('immediate-from-inline')
    window.addEventListener('DOMContentLoaded', () => {
      if (typeof report === 'function') report('from-dcl')
    })
  </script></body></html>`
))
await new Promise((r) => server.listen(0, '::', r))

const browser = await puppeteer.connect({
  browserWSEndpoint: 'ws://127.0.0.1:9222/devtools/browser'
})
const page = await browser.newPage()
const seen = []
await page.exposeFunction('report', (msg) => seen.push(msg))
await page.goto(`http://localhost:${server.address().port}/`)
console.log(seen) // ["immediate-from-inline", "from-dcl"]
```

Without commit 2, `immediate-from-inline` is missing (inline script
runs before the binding is installed). Without commit 3, `seen`
stays empty entirely.

`cargo build --release --features stealth` is clean; `cargo check
--features stealth` passes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,18 +20,30 @@ jobs:
             # newer glibc shipped by ubuntu-latest.
             os: ubuntu-22.04
             name: obscura-x86_64-linux
+            cross: false
           - target: aarch64-unknown-linux-gnu
+            # Native build on GitHub's ARM runner.
             os: ubuntu-22.04-arm
             name: obscura-aarch64-linux
+            cross: false
+          - target: aarch64-unknown-linux-gnu
+            # Cross-compile fallback from x86_64 Ubuntu 22.04. Same glibc
+            # 2.35 floor, useful when ARM runners are unavailable.
+            os: ubuntu-22.04
+            name: obscura-aarch64-linux-cross
+            cross: true
           - target: aarch64-apple-darwin
             os: macos-latest
             name: obscura-aarch64-macos
+            cross: false
           - target: x86_64-apple-darwin
             os: macos-latest
             name: obscura-x86_64-macos
+            cross: false
           - target: x86_64-pc-windows-msvc
             os: windows-latest
             name: obscura-x86_64-windows
+            cross: false
 
     runs-on: ${{ matrix.os }}
 
@@ -41,6 +53,17 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: ${{ matrix.target }}
+
+      - name: Install cross-compilation toolchain (aarch64-linux)
+        if: matrix.cross == true
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gcc-aarch64-linux-gnu
+          mkdir -p .cargo
+          cat >> .cargo/config.toml << 'CARGO_EOF'
+          [target.aarch64-unknown-linux-gnu]
+          linker = "aarch64-linux-gnu-gcc"
+          CARGO_EOF
 
       - name: Build
         run: cargo build --release --target ${{ matrix.target }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -741,6 +741,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "equivalent"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1699,6 +1708,7 @@ name = "obscura-net"
 version = "0.1.0"
 dependencies = [
  "async-trait",
+ "encoding_rs",
  "reqwest",
  "serde",
  "serde_json",
@@ -2165,7 +2175,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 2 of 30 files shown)


SGavrl:
Architecture looks good and primary suite passes, but caught two functional bugs:

1. **`about:blank` skips preload injection**
In `page.rs`, the early-return for the `about:` scheme calls `Maps_blank()` + `init_js()`, but exits before reaching `execute_scripts()`. This means `addScriptToEvaluateOnNewDocument` scripts and binding shims never run for `about:blank` targets. (Note: `data:` URLs work fine).

2. **Binding shim drops multi-args**
The binding shim in `runtime.rs` is defined as `function (arg) {...}`, so it only captures the first argument. Multi-arg calls like `window.myBinding('x', 42)` send the payload `'x'` instead of `["x", 42]`. 

Using `(...args) => JSON.stringify(args)` here will fix it.

marcbachmann:
@SGavrl Thanks for catching these — pushed `c4ba33b` as a follow-up commit.

**#1 (`about:blank` skips preloads)** — confirmed bug. The `about:` branch in `navigate_single` returned before any preloads ran, so `addScriptToEvaluateOnNewDocument` scripts and the `Runtime.addBinding` shim never installed. Puppeteer hits this every session: `browser.newPage()` lands on `about:blank` and a follow-up `exposeFunction` is unusable until the first real navigation. Fixed by running the preload set after `init_js()` in that branch, guarded by the same 5s timeout as regular preloads.

**#2 (binding shim arity)** — I want to push back on the proposed `(...args) => JSON.stringify(args)` fix. Chromium's `V8InspectorImpl::bindingCallback` (in `v8/src/inspector/v8-inspector-impl.cc`) rejects calls without exactly one argument and `ToString`-coerces that argument before emitting `Runtime.bindingCalled`. So the proposed change would alter the wire format on the common path: a single-string call (which is what puppeteer's `exposeFunction` wrapper always produces — it pre-stringifies its own multi-arg array into a single JSON string before calling the binding) would arrive as `payload = \"[\\\"x\\\"]\"` instead of `payload = \"x\"`, breaking the contract the rest of this PR is built around.

What I did instead: keep the single-arg signature so single-string calls round-trip unchanged, add an `arguments.length !== 1` guard to match Chrome's strictness, and swap `JSON.stringify(arg)` → `String(arg)` so the non-string branch does `ToString` coercion the way Chrome does. Net effect: `myBinding('x', 42)` now no-ops (matching Chrome) instead of silently dropping `42` while delivering `'x'`.

`cargo check --features stealth` clean.

PR #114, opened 2026-05-05
feat: cookie jar persistence + --storage-dir flag + fix HTTP cookie header filtering

zhangyang-crazy-one:
## Summary

Adds cookie jar save/load persistence to Obscura, enabling login sessions to survive process restarts. Also fixes a bug where valid cookies were silently dropped from HTTP requests.

### Cookie Jar Serialization
- `CookieJar::save_to_file()` — atomic JSON write of non-expired cookies
- `CookieJar::load_from_file()` — load + merge from JSON file  
- Made `CookieEntry` derive `Serialize`/`Deserialize`

### --storage-dir Flag
- Available on both `fetch` and `serve` commands
- `BrowserContext::with_storage()` auto-loads cookies from `{storage_dir}/cookies.json`
- `save_cookies()` persists on shutdown
- Threaded through CDP serve path (CdpContext -> cdp_processor -> start_with_full_options)

### HTTP Cookie Header Filtering Fix
- **Bug:** `reqwest::HeaderValue::from_str()` silently rejects the entire Cookie header when ANY cookie value has invalid HTTP header characters
- **Fix:** Filter out individual invalid pairs instead of dropping all cookies. Added debug logging.

### Storage Domain Stubs
- `Storage.getUsageAndQuota`, `Storage.clearDataForOrigin` for Puppeteer/Playwright compat

## Verification

```
obscura serve --port 9222 --storage-dir ~/.obscura-data
# Cookie header: 19 cookies (1029 bytes) for www.xiaohongshu.com
# Page renders logged-in content correctly
```

All unit tests pass (21 tests, including 3 new save/load roundtrip tests).

**8 files, ~240 lines.**

PR #170, opened 2026-05-23
feat(cdp): implement Network.setCookie + deleteCookies, full field fidelity

### LINKED ISSUE #168: Support Network.setCookie and Network.getCookies CDP methods
## Problem

  When using Obscura's CDP server (`obscura serve`), there's no way to inject cookies before navigating to a page.

  Attempting to use the standard `Network.setCookie` CDP method returns:

  ```
  CDP error for Network.setCookie: Unknown Network method: setCookie
  ```

  This makes it impossible to:
  - Load authenticated sessions on any cookie-gated site
  - Replay exported browser cookies from tools like Cookie-Editor
  - Automate any workflow that requires login state

  ## Current workaround

  Fall back to Playwright/Puppeteer with headless Chrome, which defeats the purpose of using Obscura (lower memory, faster loads,       
  built-in anti-detect).

  Injecting cookies via `Runtime.evaluate` with `document.cookie = ...` doesn't work either — httpOnly cookies can't be set via
  JavaScript.

  ## Requested CDP methods

  ### `Network.setCookie`
  Set a single cookie. Parameters:
  - `name` (string, required)
  - `value` (string, required)
  - `domain` (string, optional)
  - `path` (string, optional, default `/`)
  - `secure` (boolean, optional)
  - `httpOnly` (boolean, optional)
  - `sameSite` (string, optional — `Strict`, `Lax`, `None`)
  - `expires` (number, optional — epoch seconds)

  ### `Network.setCookies`
  Batch version — accepts a `cookies` array of the above.

  ### `Network.getCookies`
  Returns all cookies for the current page/domain. Useful for exporting session state.

  ### `Network.deleteCookies`
  Delete cookies by name/domain. Useful for clearing state between sessions.

  ## Expected flow

  ```
  1. obscura serve --stealth
  2. Connect via WebSocket
  3. Network.setCookies([{name: "session", value: "...", domain: ".example.com", httpOnly: true}])
  4. Page.navigate({url: "https://example.com/dashboard"})
  5. Runtime.evaluate({expression: "document.body.innerText"})
  ```

  This is the standard CDP cookie flow that works with Chrome. Adding these methods would make Obscura a full drop-in replacement for   
  headless Chrome in authenticated scraping workflows.

  ## Environment

  - Obscura v0.1.5
  - Windows x86_64
  - `obscura serve --stealth --port 9333` with WebSocket CDP client

ousamabenyounes:
## Problem

Closes #168.

`Network.setCookie` (singular) and `Network.deleteCookies` were missing — both fail with `Unknown Network method`. The existing `Network.setCookies` / `Network.getCookies` work but silently drop `sameSite`, `expires`, and the `url` fallback, so round-trip via `Cookie-Editor` / Puppeteer / Playwright is lossy: cookies come back as session cookies with `sameSite=Lax` regardless of what was set.

## Root Cause

- `Network.setCookie` (singular) — not routed in `domains/network.rs`.
- `Network.deleteCookies` — not routed in `domains/network.rs`.
- `Network.setCookies` — built `obscura_net::CookieInfo` without `sameSite` / `expires` / `url` fields (struct had no slot for them).
- `Network.getCookies` — hardcoded `expires: -1` and never emitted `sameSite`.
- `Storage` cookie handlers re-implemented the same parsing inline, with the same gaps in the opposite direction.

## Fix

- Extend `obscura_net::CookieInfo` with `same_site: String` + `expires: Option<i64>` (`#[serde(default)]` so the on-disk cookie-jar JSON used in #114 stays backward-compatible).
- Wire `set_cookies_from_cdp` + `get_all_cookies` to round-trip both fields, and add `delete_cookies_filtered(name, domain, path)` on the jar.
- New `crates/obscura-cdp/src/cookie_params.rs` shared helper: `parse_cdp_cookie` (handles the `url` -> derive `domain`/`path` fallback, plus `sameSite`/`expires`/`httpOnly`/`secure`) and `parse_delete_cookies_params`.
- Rewrite `domains/network.rs`:
  - Add `Network.setCookie` (singular).
  - Add `Network.deleteCookies`.
  - `Network.setCookies` now preserves `sameSite` + `expires` + `url`-fallback.
  - `Network.getCookies` emits real `expires`, real `sameSite`, and a correct `session` flag.
- Refactor `domains/storage.rs` to call the same helper instead of duplicating the parsing logic.

## Verification (real RED -> GREEN, measured)

Procedure: same machine, same WS client, same JSON payload. RED = `git checkout 1950048 -- crates/` (merge-base before this branch), rebuild, run. GREEN = `git checkout HEAD -- crates/`, rebuild, run.

```
$ ./target/release/obscura serve --port 19223
$ python3 smoke_cookies.py
```

### RED (1950048, pre-patch)

```
=== Network.setCookie (singular, full attrs) ===
{
  "id": 3,
  "error": {
    "code": -32601,
    "message": "Unknown Network method: setCookie"
  }
}

=== Network.deleteCookies ===
{
  "id": 4,
  "error": {
    "code": -32601,
    "message": "Unknown Network method: deleteCookies"
  }
}

=== Network.setCookies -> Network.getCookies (sameSite + expires fidelity) ===
Traceback: KeyError: 'sameSite'        # response has no sameSite field at all
```

### GREEN (this PR)

```
=== Network.setCookie (singular, full attrs) ===
{
  "id": 3,
  "result": { "success": true }
}

=== Network.deleteCookies ===
{
  "id": 4,
  "result": {}
}

=== Network.setCookies -> Network.getCookies (sameSite + expires fidelity) ===
  name=tok sameSite=Strict expires=1900000000 httpOnly=True session=False
```

### Side-by-side

| Method | RED (1950048) | GREEN (this PR) |
|---|---|---|
| `Network.setCookie` | `error: "Unknown Network method: setCookie"` | `result: { success: true }` |
| `Network.deleteCookies` | `error: "Unknown Network method: deleteCookies"` | `result: {}` |
| `Network.getCookies` -> `sameSite` | field absent | `Strict` round-trip |
| `Network.getCookies` -> `expires` | hardcoded `-1` | `1900000000` round-trip |
| `Network.getCookies` -> `session` | always `true` | `false` when `expires` set |

The RED `Unknown Network method: setCookie` matches the exact error string the reporter quotes in #168.

## Test plan

- [x] `cargo test -p obscura-net` -- 22 passed (4 new tests covering: sameSite/expires preservation, expired-cookie drop, deleteCookies path filter mismatch + null path).
- [x] `cargo test -p obscura-cdp` -- 24 unit + 1 integration passed (6 new tests in `cookie_params` covering: explicit domain/path, url-fallback, default `/` path, name-required guard, delete-by-url).
- [x] `cargo test -p obscura-js` -- 76 passed (no regressions).
- [x] `cargo build --release` -- clean.
- [x] End-to-end WS RED -> GREEN against `obscura serve` (above).
- [ ] Stealth build not exercised on this host -- change is in `obscura-cdp` and the cookie-jar interface; stealth path uses the same jar so no stealth-specific work required.

## Risk

Low. `CookieInfo` gains two fields with `#[serde(default)]`, so any consumer relying on the existing JSON shape (cookie-jar persistence on disk for #114) is unaffected. `set_cookies_from_cdp` still accepts the existing struct shape -- the new fields default to empty/None. Storage domain semantics unchanged (now shares logic with Network instead of duplicating it).

### DIFF
--- a/crates/obscura-cdp/src/cookie_params.rs
+++ b/crates/obscura-cdp/src/cookie_params.rs
@@ -1,0 +1,162 @@
+use obscura_net::CookieInfo;
+use serde_json::Value;
+
+const DEFAULT_COOKIE_PATH: &str = "/";
+
+pub fn parse_cdp_cookie(value: &Value) -> Option<CookieInfo> {
+    let name = value.get("name").and_then(|v| v.as_str())?.to_string();
+    let cookie_value = value
+        .get("value")
+        .and_then(|v| v.as_str())
+        .unwrap_or("")
+        .to_string();
+
+    let url_parsed = value
+        .get("url")
+        .and_then(|v| v.as_str())
+        .and_then(|u| url::Url::parse(u).ok());
+
+    let domain = value
+        .get("domain")
+        .and_then(|v| v.as_str())
+        .map(|s| s.to_string())
+        .or_else(|| url_parsed.as_ref().and_then(|u| u.host_str().map(|h| h.to_string())))
+        .unwrap_or_default();
+
+    if domain.is_empty() {
+        return None;
+    }
+
+    let path = value
+        .get("path")
+        .and_then(|v| v.as_str())
+        .map(|s| s.to_string())
+        .or_else(|| url_parsed.as_ref().map(|u| u.path().to_string()))
+        .unwrap_or_else(|| DEFAULT_COOKIE_PATH.to_string());
+
+    let secure = value.get("secure").and_then(|v| v.as_bool()).unwrap_or(false);
+    let http_only = value.get("httpOnly").and_then(|v| v.as_bool()).unwrap_or(false);
+    let same_site = value
+        .get("sameSite")
+        .and_then(|v| v.as_str())
+        .unwrap_or("")
+        .to_string();
+    let expires = value.get("expires").and_then(|v| v.as_f64()).map(|f| f as i64);
+
+    Some(CookieInfo {
+        name,
+        value: cookie_value,
+        domain,
+        path,
+        secure,
+        http_only,
+        same_site,
+        expires,
+    })
+}
+
+pub struct DeleteCookiesFilter {
+    pub name: String,
+    pub domain: String,
+    pub path: Option<String>,
+}
+
+pub fn parse_delete_cookies_params(params: &Value) -> Option<DeleteCookiesFilter> {
+    let name = params.get("name").and_then(|v| v.as_str())?.to_string();
+    if name.is_empty() {
+        return None;
+    }
+
+    let url_parsed = params
+        .get("url")
+        .and_then(|v| v.as_str())
+        .and_then(|u| url::Url::parse(u).ok());
+
+    let domain = params
+        .get("domain")
+        .and_then(|v| v.as_str())
+        .map(|s| s.to_string())
+        .or_else(|| url_parsed.as_ref().and_then(|u| u.host_str().map(|h| h.to_string())))
+        .unwrap_or_default();
+
+    let path = params
+        .get("path")
+        .and_then(|v| v.as_str())
+        .map(|s| s.to_string())
+        .or_else(|| url_parsed.as_ref().map(|u| u.path().to_string()));
+
+    Some(DeleteCookiesFilter { name, domain, path })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::json;
+
+    #[test]
+    fn parse_with_explicit_domain_path() {
+        let v = json!({
+            "name": "session",
+            "value": "abc",
+            "domain": ".example.com",
+            "path": "/app",
+            "secure": true,
+            "httpOnly": true,
+            "sameSite": "Strict",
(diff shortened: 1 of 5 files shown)


SGavrl:
Storage and serialization changes look good. setCookie and deleteCookies
  Puppeteer and Playwright do by default). Mirroring how Storage.setCookies
  already grabs ctx.default_context.cookie_jar when there's no session
  would fix it.

  Also need a Network.getAllCookies arm alongside getCookies, they're
  separate CDP methods in the spec.

ousamabenyounes:
Thanks for the review @SGavrl — agreed on both. 1c3884f extracts a `cookie_jar_for` helper that prefers the session page jar and falls back to `ctx.default_context.cookie_jar` when no target is attached (mirrors `Storage.setCookies`), and adds a `Network.getAllCookies` arm alongside `getCookies`. Six new `#[tokio::test]` cases cover the no-session fallback for set/get/setMany/delete/clear and the new arm — RED on the prod-reverted file, GREEN on patched.

Issue #167, opened 2026-05-21
Add web-platform-tests (WPT) conformance suite to CI


joeblew999:
## Why

Several recent issues (#15, #21, #85, #103, #105, #147, #155, #163) are each a specific manifestation of a Chrome runtime behavior that's unimplemented or differs subtly in Obscura. They're closed or open individually, but they share a single root cause: **no systematic conformance testing against the spec'd browser surface**, so gaps are found reactively when users hit broken pages.

The browser-engine ecosystem has a standard answer to this exact problem: **[web-platform-tests (WPT)](https://github.com/web-platform-tests/wpt)** — the official cross-browser conformance suite, used by Chrome, Firefox, Safari, and (most relevantly for prior art) [Servo](https://github.com/servo/servo).

## What

Wire WPT into Obscura's CI. Even a targeted slice would surface most of the bug patterns we're seeing:

| WPT directory | Maps to issues in this repo |
|---|---|
| `dom/`, `html/dom/` | #163 (Storage, CharacterData ChildNode methods, Element.remove) |
| `html/syntax/` | DOM construction timing (manifests as `Cannot read properties of null (reading 'insertBefore')` from inline scripts running before their target element exists — the pattern in #85, #103) |
| `html/webappapis/scripting/` | Script loader / module ordering (manifests as `Cannot read properties of undefined (reading 'prototype')` from JS bundles that depend on Chrome's exact load order — pattern in #147, #155) |
| `webstorage/` | #163 (`Storage` global, `instanceof Storage`) |
| `encoding/`, `fetch/`, `xhr/` | #113 and similar text-encoding / network shim gaps |

The full WPT suite is too large to gate every PR (~3000+ test files), but a curated subset focused on the DOM + HTML scripting + Web Storage areas (where Obscura takes the most fire) would be tractable: ~minutes per CI run, gives a measurable conformance percentage to ratchet up over time, and prevents regressions on bugs already fixed.

## Precedent

Servo runs WPT continuously and is the closest sibling project to Obscura — a new Rust browser engine trying to be Chrome-compatible. Their wiring lives at [servo/tests/wpt](https://github.com/servo/servo/tree/master/tests/wpt). The relevant pieces:

- A vendored / submoduled `wpt/` checkout
- A test harness that runs subsets via tags (`--tag=dom`, `--tag=scripting`, etc.)
- An expected-results database (`metadata/*.ini`) so passing/failing tests don't churn the diff
[...]

PR #222, opened 2026-06-02
fix(js): implement DocumentFragment getElementById

rifuki:
## Summary

Implements `DocumentFragment.prototype.getElementById()` so it searches descendant elements inside the fragment instead of always returning `null`.

The implementation performs a scoped descendant traversal and compares literal element IDs, avoiding selector pitfalls for IDs like `a.b`.

Related to #167.

## Tests

- `cargo test -p obscura-js document_fragment_get_element_by_id_searches_descendants -- --nocapture`
- `cargo test -p obscura-js get_element_by_id -- --nocapture`
- `cargo test -p obscura-js`
- `cargo test --workspace`

## Real binary smoke tests

Built and tested with the real debug binary:

- `cargo build -p obscura-cli`
- `./target/debug/obscura fetch https://example.com --dump text --quiet`
- `./target/debug/obscura fetch https://example.com --quiet --eval "...DocumentFragment.getElementById..."`
- `./target/debug/obscura fetch https://example.org --quiet --timeout 30 --wait 1 --eval "...DocumentFragment.getElementById..."`
- `./target/debug/obscura fetch https://www.iana.org/help/example-domains --quiet --timeout 30 --wait 1 --eval "...DocumentFragment.getElementById..."`
- `./target/debug/obscura fetch https://httpbin.org/html --quiet --timeout 30 --wait 1 --eval "...DocumentFragment.getElementById..."`

Observed live-site eval output included:

```text
fragment|true|literal
```

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -1542,7 +1542,18 @@ class DocumentFragment extends Node {
   }
   get firstElementChild() { return this.children[0] || null; }
   get lastElementChild() { const ch = this.children; return ch[ch.length - 1] || null; }
-  getElementById(id) { return null; }
+  getElementById(id) {
+    const needle = String(id);
+    const stack = Array.from(this.childNodes || []).reverse();
+    while (stack.length) {
+      const node = stack.pop();
+      if (!node) continue;
+      if (node.nodeType === 1 && node.id === needle) return node;
+      const children = node.childNodes || [];
+      for (let i = children.length - 1; i >= 0; i--) stack.push(children[i]);
+    }
+    return null;
+  }
   cloneNode(deep) {
     const frag = document.createDocumentFragment();
     if (deep) frag.innerHTML = this.innerHTML;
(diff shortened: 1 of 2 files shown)


SGavrl:
this works, just one nit: the walk is breadth-first (shift + push), so when two elements share an id at different depths it returns a different one than querySelector('#id'). getElementById is supposed to be tree order (preorder).

  repro:
  <div><span id="dup">deep</span></div><p id="dup">shallow</p>
  frag.getElementById('dup')  ->  shallow
  frag.querySelector('#dup')  ->  deep
  
  dup ids are invalid html but they show up in the wild and chrome returns the tree-order one. easy fix is to prepend children instead of appending so the walk is preorder:

  queue.unshift(...(node.childNodes || []));

  could you add a dup-id test too while you're at it? the current one only has a single #target so it passes either way.

rifuki:
Good catch, thanks! Updated the traversal to use tree-order/preorder instead of breadth-first order, and added duplicate-id regression coverage for the nested-vs-shallow case.

Re-ran:

- `cargo test -p obscura-js document_fragment_get_element_by_id_searches_descendants -- --nocapture`
- `cargo test -p obscura-js`
- `cargo test --workspace`

Real binary smoke against `https://example.com` also passes, including duplicate-id ordering:

```text
fragment|true|literal|deep
```

PR #227, opened 2026-06-03
Many Stealth Improvements

sspilleman:
# Stealth Fixes

This document catalogues every stealth-related fix applied to the Obscura headless browser. It includes both pre-existing fixes (carried over from earlier sessions) and new fixes implemented during this session.

---

## Pre-Existing Fixes (fixes I did earlier)

These fixes were already identified and applied before this session, then verified and retained.

### Fix #1 — Tracker blocking in stealth HTTP client

**Why:** Without it, `StealthHttpClient::fetch()` bypassed the PGL blocklist, leaking requests to known trackers.

**What:** Added `crate::blocklist::is_blocked(host)` check at the top of `StealthHttpClient::fetch()`. If the host matches the blocklist, the request is short-circuited with a 403 response.

**Files:**
- `crates/obscura-net/src/wreq_client.rs:77-88`

---

### Fix #2 — hardwareConcurrency / deviceMemory randomization

**Why:** `navigator.hardwareConcurrency` and `navigator.deviceMemory` were set to static values (`8`), producing identical fingerprints across all sessions.

**What:** Added `_fpRand(400)` / `_fpRand(401)`-seeded overrides in `__obscura_init()` that pick from a pool of realistic values (2/4/6/8/12/16 for cores; 0.25/0.5/1/2/4/8 for memory).

**Files:**
- `crates/obscura-js/js/bootstrap.js:4200-4203`

---

### Fix #3 — SOCKS5 proxy compatibility with stealth

**Why:** SOCKS5 proxies were rejected when stealth mode was enabled, and the `mcp_proxy` CLI flag had a bug where the merged proxy was computed but never passed to the MCP server.

**What:** Added `"socks"` feature to both `wreq` dependencies in `Cargo.toml`. Removed `reject_stealth_with_socks5()` function, its 3 call sites, 6 tests, and import from `main.rs`. Fixed the `mcp_proxy` bug so the merged proxy URL is actually forwarded to `obscura_mcp::run()`.

**Files:**
- `crates/obscura-net/Cargo.toml`
- `crates/obscura-cli/src/main.rs`

---

### Fix #4 — RpcResponse warning

**Why:** Compiler warning about `pub` function in a private type.

**What:** Changed `pub` to `pub(crate)` on `dispatch()`.

**Files:**
- `crates/obscura-mcp/src/lib.rs:132`

---

### Re-enable concurrent navigation tests

[...]

PR #232, opened 2026-06-03
Fix/rescue 227 regressions

SGavrl:
## Regressions from #227

  **build.rs deleted** (#224 reverted)
  #227 removed the build script that derives OBSCURA_BUILD_VERSION from the
  git tag. Restored build.rs and the two references to OBSCURA_BUILD_VERSION
  in main.rs (--version string and startup banner).

  **--allow-private-network broken**
  The block that mirrors the CLI flag into OBSCURA_ALLOW_PRIVATE_NETWORK was
  removed from main.rs, and the env_allows_private_network() check was removed
  from validate_fetch_url in ops.rs. Together these made --allow-private-network
  a no-op for JS-side fetch calls. Both restored.

  **DOM cycle guards removed**
  append_child and insert_before lost their self-reference no-ops. Without
  them, passing a node as both parent and child (or as both reference and
  new sibling) corrupts the linked list and causes infinite traversal loops.
  Both guards restored in tree.rs.

  **textContent CharacterData fast-path removed**
  Calling .textContent on a Text, Comment, or ProcessingInstruction node
  fell through to the descendant walker instead of returning the node's own
  data. Fast-path restored in tree.rs.

  **PI/Doctype Rust ops removed**
  create_processing_instruction, create_doctype, pi_target, doctype_name,
  doctype_public_id, and the ProcessingInstruction branch in set_text_content
  were all deleted from ops.rs. These back the JS DOM APIs added in #230/#231.
  All restored.

  ## Bugs in #227's stealth code

  **Europe/Germany is not a valid IANA timezone**
  Intl.DateTimeFormat silently fell back to UTC. Changed to Europe/Berlin,
  which is consistent with the Frankfurt geolocation coordinates.

  **navigator.plugins[Symbol.iterator] infinite recursion**
  The implementation was `function() { return this[Symbol.iterator](); }`,
  which recurses forever and blows the stack on Array.from(navigator.plugins).
  Fixed to bind Array.prototype[Symbol.iterator] directly to the array.

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -1842,7 +1842,7 @@ globalThis.navigator = {
     p.item = (i) => p[i] || null;
     p.namedItem = (name) => p.find(x => x.name === name) || null;
     p.refresh = () => {};
-    p[Symbol.iterator] = function() { return this[Symbol.iterator](); };
+    p[Symbol.iterator] = Array.prototype[Symbol.iterator].bind(p);
     return p;
   },
   get mimeTypes() {
@@ -4312,7 +4312,7 @@ _markNative(MediaStream); _markNative(MediaStreamTrack);
 _markNative(RTCPeerConnection); _markNative(RTCSessionDescription); _markNative(RTCIceCandidate);
 
 const _OrigDateTimeFormat = Intl.DateTimeFormat;
-const _defaultTZ = 'Europe/Germany';
+const _defaultTZ = 'Europe/Berlin';
 Intl.DateTimeFormat = function(locales, options) {
   if (!options) options = {};
   if (!options.timeZone) options.timeZone = _defaultTZ;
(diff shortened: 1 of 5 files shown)


crates/obscura-js/js/bootstrap.js
```diff
@@ -4312,7 +4312,7 @@ _markNative(MediaStream); _markNative(MediaStreamTrack);
 _markNative(RTCPeerConnection); _markNative(RTCSessionDescription); _markNative(RTCIceCandidate);
 
 const _OrigDateTimeFormat = Intl.DateTimeFormat;
-const _defaultTZ = 'Europe/Germany';
+const _defaultTZ = 'Europe/Berlin';
```
lkraider, line 4315: Any reason why use 'Europe/Berlin' as default timezone?

SGavrl: No particular reason, probably will make it pick from a table of different timezones in the future

PR #243, opened 2026-06-05
fix(api/js): Browser cookie jar + base href + Node.baseURI + XHR EventTarget + Page::settle (closes #241,#242,#244,#245,#247)

### LINKED ISSUE #241: Browser::cookies() returns a different jar than the one used during navigation
## Summary

`Browser::cookies()` returns a `CookieStore` wrapping `Browser.cookie_jar`, but the `CookieJar` actually used by the HTTP client during navigation is a *different* one owned by `BrowserContext`. Cookies set via `Set-Cookie` response headers during `page.goto(...)` are therefore never visible through the public API.

## Reproduction

```rust
use obscura::Browser;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let browser = Browser::builder().build()?;
    let mut page = browser.new_page().await?;
    page.goto("https://live.deutsche-boerse.com/bond/xs2290957732-bahrain-koenigreich-6-25-21-51?mic=XFRA").await?;
    tokio::time::sleep(std::time::Duration::from_secs(8)).await;

    // Server definitely sends Set-Cookie on this response (verifiable with curl -I),
    // and document.cookie via page.evaluate() also reflects nothing,
    // but more importantly:
    println!("Total cookies via Browser::cookies(): {}", browser.cookies().get_all().len());
    // -> 0

    Ok(())
}
```

Empirically `browser.cookies().get_all().len() == 0` for any navigated page.

## Root cause

`crates/obscura/src/browser.rs` (`Browser::build`):

```rust
let cookie_jar = Arc::new(CookieJar::new());        // (1) Browser's own jar

let context = if let Some(ref dir) = config.storage_dir {
    BrowserContext::with_storage_full(/* no jar arg */, ..)
} else {
    BrowserContext::with_full_options(/* no jar arg */, ..)
};

Ok(Browser {
    context: Arc::new(context),
    cookie_jar,                                     // (1) is stored on Browser
})
```

```rust
pub fn cookies(&self) -> CookieStore {
    CookieStore::new(self.cookie_jar.clone())       // returns (1), never written to
}
```

`crates/obscura-browser/src/context.rs:80` (`BrowserContext::_new_inner`):

```rust
let cookie_jar = Arc::new(CookieJar::new());        // (2) Context's own jar
// ...
let http_client = ObscuraHttpClient::with_cookie_jar(cookie_jar.clone(), ..);
// ...
Self { cookie_jar, http_client, .. }                // (2) is shared with http_client
```

`crates/obscura-net/src/client.rs:441` writes Set-Cookie entries into `self.cookie_jar` — which is `(2)`, the context's jar. The public `Browser::cookies()` accessor returns `(1)`, which nothing populates.

Result: the cookie storage that holds real session cookies is unreachable from the public API.

## Suggested fix

Make `Browser::build` adopt `context.cookie_jar` instead of creating a fresh `(1)`. Something like:

```rust
let context = Arc::new(context);
Ok(Browser {
    cookie_jar: context.cookie_jar.clone(),
    context,
})
```

Then the existing `Browser::cookies()` accessor reads from the same jar the HTTP client writes to.

The `storage_dir` cookie persistence block in `Browser::build` would need to move (or be coordinated with `BrowserContext`'s own storage handling) — `BrowserContext::_new_inner` already does `cookie_jar.load_from_file(&cookie_path)` when `storage_dir` is set, so the block in `Browser::build` is actually redundant.

## Impact

Any caller using `browser.cookies()` to inspect or persist session state after navigation gets an empty result. Discovered while building a scraper that depended on cookies set by the navigated page; we worked around it by reverting to chromiumoxide because there is no public way to reach the populated jar.

### LINKED ISSUE #242: Script src URLs ignore <base href>, breaking Angular/Next/Vite SPAs whose entry HTML is served from a sub-path
## Summary

When the page HTML contains a `<base href="..."/>` element, obscura ignores it while resolving relative `<script src>` URLs. Resolution falls back to the document URL, so any SPA whose entry HTML lives on a sub-path and whose bundle filenames are relative breaks completely — V8 receives the SPA shell HTML back from the 404'd bundle URL and fails with `Uncaught SyntaxError: Unexpected token '<'`.

## Reproduction

Page: `https://live.deutsche-boerse.com/bond/xs2290957732-bahrain-koenigreich-6-25-21-51?mic=XFRA`

The HTML contains:

```html
<base href="/">
<script type="module" src="runtime.7b88bb4fc73d55e7.js"></script>
<script type="module" src="polyfills.537946eab80126b1.js"></script>
<script src="scripts.379936dfffadafcc.js" defer></script>
<script type="module" src="main.52c5d9b253b84d19.js"></script>
```

Real browsers resolve these to `https://live.deutsche-boerse.com/runtime.…js` etc., because of the `<base>` element. Obscura resolves them to `https://live.deutsche-boerse.com/bond/runtime.…js` (i.e., joined against the document URL `/bond/xs22…`).

Run with `RUST_LOG=obscura_browser=warn`:

```
WARN obscura_browser::page: ES module error (https://live.deutsche-boerse.com/bond/runtime.7b88bb4fc73d55e7.js): Module load error: Uncaught SyntaxError: Unexpected token '<'
WARN obscura_browser::page: ES module error (https://live.deutsche-boerse.com/bond/polyfills.537946eab80126b1.js): Module load error: Uncaught SyntaxError: Unexpected token '<'
WARN obscura_browser::page: ES module error (https://live.deutsche-boerse.com/bond/main.52c5d9b253b84d19.js): Module load error: Uncaught SyntaxError: Unexpected token '<'
WARN obscura_browser::page: Script error (https://live.deutsche-boerse.com/bond/scripts.379936dfffadafcc.js): JS error: Uncaught SyntaxError: Unexpected token '<'
```

The `<` is the first character of the SPA shell HTML (`<!doctype html>…`), which the server returns for the wrong path under SPA-fallback routing.

Diagnostic confirming `<base>` is present in the DOM but `document.baseURI` is empty:

```
page.evaluate("document.baseURI || ''")          // -> ""
page.evaluate("(document.querySelector('base') || {}).href || ''")   // -> "/"
```

(Note: `document.baseURI` returning empty when a `<base>` exists is itself a secondary spec-compliance issue.)

## Root cause

`crates/obscura-browser/src/page.rs` in `execute_scripts`:

```rust
let full_url = if src_url.starts_with("http://") || src_url.starts_with("https://") {
    src_url.clone()
} else if let Some(base) = &self.url {
    base.join(src_url).map(|u| u.to_string()).unwrap_or_else(|_| src_url.clone())
} else {
    src_url.clone()
};
```

`self.url` is the document URL. Per the HTML spec ([Document base URL](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#document-base-url)), if the document contains a `<base href>` element, its href becomes the document base URL, and *that* is what relative URL resolution must use — not the document URL.

The same pattern applies to the module-script branch a few hundred lines later (also `base.join(src)` against `self.url`).

## Suggested fix

Before resolving any `<script src>`, walk the DOM for the first `<base href>` element (must be in `<head>` per spec, and only the first counts). Resolve it against the document URL, and use the result as the base for all subsequent relative URL resolution.

Rough sketch (Rust-ish pseudocode):

```rust
fn document_base_url(&self) -> Option<Url> {
    let dom_base = self.js.as_ref().and_then(|js| js.with_dom(|dom| {
        dom.query_selector("base[href]").and_then(|nid| {
            dom.get_node(nid).and_then(|n| n.get_attribute("href").map(|s| s.to_string()))
        })
    })).flatten();

    let doc_url = self.url.as_ref()?;
    match dom_base {
        Some(href) => doc_url.join(&href).ok().or_else(|| Some(doc_url.clone())),
        None => Some(doc_url.clone()),
    }
}
```

Then in `execute_scripts`:

```rust
let base = self.document_base_url();
// ... use `base` instead of `self.url` for all script src resolution
```

This should also fix `document.baseURI` (it currently returns "" because obscura's V8 bindings probably look at the same wrong source).

## Impact

Any SPA shipped from a sub-path with relative bundle paths (the Angular/Vite/Next.js norm when `BASE_URL` ≠ `/`) is currently un-bootable in obscura. The page renders its SSR HTML and stops there. Discovered while building a real-world scraper; surfaces in chromiumoxide / real Chrome correctly, surfaces in obscura as completely silent failure (without `RUST_LOG`, you just get a static SSR page with no client runtime).

zhangyang-crazy-one:
### 1. Cookie jar disconnect (closes #241)
`crates/obscura/src/browser.rs` — 4 insertions, 13 deletions.

`Browser::cookies()` returned an empty `CookieStore` after navigation because `Browser::build()` created its own empty `CookieJar`, while `BrowserContext` internally uses a *separate* jar for HTTP `Set-Cookie` handling.

**Fix**: Use `context.cookie_jar.clone()` instead of `Arc::new(CookieJar::new())`. Also removes redundant cookie-loading in `Browser::build()` since `BrowserContext::_new_inner` already handles `storage_dir` cookie loading.

### 2. Script src <base href> resolution (closes #242)
`crates/obscura-browser/src/page.rs` — 25 insertions, 2 deletions.

Relative `<script src>` URLs were resolved against the document URL instead of `<base href>`. This broke all SPAs served from sub-paths (Angular/Vite/Next.js with `BASE_URL != "/"`).

**Fix**: Added `resolve_base_url()` that queries DOM for `<base href>` and uses it as the URL base per [HTML spec](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#document-base-url). Applied to both regular script loading and ES module loading.

## Verification
- `cargo check -p obscura` passes
- `cargo check -p obscura-browser` passes

@kakserpom thanks for the detailed analysis!

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -282,8 +282,31 @@ impl Page {
         self.js = Some(rt);
     }
 
+    /// Resolve the document base URL per HTML spec:
+    /// https://html.spec.whatwg.org/multipage/urls-and-fetching.html#document-base-url
+    /// Falls back to self.url when no <base href> exists.
+    fn resolve_base_url(&self) -> Option<url::Url> {
+        let doc_url = self.url.as_ref()?;
+        let base_href: Option<String> = self.js.as_ref().and_then(|js| {
+            js.with_dom(|dom| {
+                match dom.query_selector("base[href]") {
+                    Ok(Some(nid)) => {
+                        dom.get_node(nid).and_then(|n| n.get_attribute("href").map(|s| s.to_string()))
+                    }
+                    _ => None,
+                }
+            }).flatten()
+        });
+        match base_href {
+            Some(href) => doc_url.join(&href).ok(),
+            None => Some(doc_url.clone()),
+        }
+    }
+
     async fn execute_scripts(&mut self) {
         tracing::info!("execute_scripts called, js runtime exists: {}", self.js.is_some());
+        // Compute document base URL, respecting <base href>.
+        let document_base = self.resolve_base_url();
         // Soft deadline on the entire script-execution phase. Heavy SPAs
         // (GitHub, Linear, CodeSandbox) ship 50+ scripts and our serial
         // fetch + execute loop can blow past a 25s Puppeteer goto timeout.
@@ -397,7 +420,7 @@ impl Page {
             if let Some(src_url) = &script.src {
                 let full_url = if src_url.starts_with("http://") || src_url.starts_with("https://") {
                     src_url.clone()
-                } else if let Some(base) = &self.url {
+                } else if let Some(base) = &document_base {
                     base.join(src_url).map(|u| u.to_string()).unwrap_or_else(|_| src_url.clone())
                 } else {
                     src_url.clone()
@@ -557,7 +580,7 @@ impl Page {
             if let Some(ref src) = module_script.src {
                 let full_url = if src.starts_with("http://") || src.starts_with("https://") {
                     src.clone()
-                } else if let Some(base) = &self.url {
+                } else if let Some(base) = &document_base {
                     base.join(src).map(|u| u.to_string()).unwrap_or_else(|_| src.clone())
                 } else {
                     src.clone()
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -2645,6 +2645,7 @@ globalThis.XMLHttpRequest = class XMLHttpRequest extends XMLHttpRequestEventTarg
   UNSENT = 0; OPENED = 1; HEADERS_RECEIVED = 2; LOADING = 3; DONE = 4;
 
   constructor() {
+    super();
     this.readyState = 0;
     this.status = 0;
     this.statusText = "";
@@ -2796,7 +2797,33 @@ globalThis.XMLHttpRequest = class XMLHttpRequest extends XMLHttpRequestEventTarg
     this.readyState = 0;
   }
 
-  // addEventListener/removeEventListener/dispatchEvent inherited from XMLHttpRequestEventTarget
+  addEventListener(type, handler) {
+    if (!this._listeners[type]) this._listeners[type] = [];
+    this._listeners[type].push(handler);
+  }
+
+  removeEventListener(type, handler) {
+    if (this._listeners[type]) {
+      this._listeners[type] = this._listeners[type].filter(h => h !== handler);
+    }
+  }
+
+  // Per WHATWG DOM spec — required by zone.js which patches XHR via
+  // Object.getOwnPropertyDescriptor on XMLHttpRequestEventTarget.prototype.
+  dispatchEvent(event) {
+    if (!event || !event.type) return false;
+    const ev = (typeof event === 'object') ? event : { type: event };
+    ev.target = ev.target || this;
+    ev.currentTarget = ev.currentTarget || this;
+    const type = ev.type;
+    const handlers = (this._listeners && this._listeners[type]) || [];
+    for (const h of handlers) { try { h.call(this, ev); } catch (e) {} }
+    const prop = 'on' + type;
+    if (typeof this[prop] === 'function') {
+      try { this[prop](ev); } catch (e) {}
+    }
+    return true;
+  }
 
   _setReadyState(state) {
     this.readyState = state;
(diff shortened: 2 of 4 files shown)


zhangyang-crazy-one:
@kakserpom Tested all three fixes against https://live.deutsche-boerse.com/bond/xs2290957732 -- all passing:

- document.baseURI: https://live.deutsche-boerse.com/ (correct)
- base href: / (detected)
- Scripts: 6 loaded, no 404 errors
- Body: 514K rendered content
- Cookie jar: connected to context

The page now renders the full bond detail page (ISIN/WKN/Symbol/Type/market data).

PR #269, opened 2026-06-08
Stealth profile consistency

johnnogueira:


crates/obscura-js/js/bootstrap.js
```diff
@@ -2373,7 +2373,7 @@ globalThis.navigator = {
   doNotTrack: null,
   deviceMemory: 8,
   connection: { effectiveType: "4g", rtt: 50, downlink: 10, saveData: false },
-  get webdriver() { return undefined; },
```
SGavrl, line 2364: This should be left undefined, the point is stealth so we should emulate what real browsers do and that is returning the value as undefined.

johnnogueira: the spec says this should be false in case not automated:
https://w3c.github.io/webdriver/#dom-navigatorautomationinformation-webdriver

https://deviceandbrowserinfo.com/are_you_a_bot has a direct evaluation if this is true or not.

are u sure this should be returning undefined? seems like old versions of Chrome used to return this as undefined, but now it returns false (claude code quoted on this lol)

SGavrl: ye looking into it you are right, that was the old behaviour that i recalled, also will be soon committing a few changes to your PR

Issue #246, opened 2026-06-05
deno_core RefCell panic during page load triggered by commit 88af239 (lazy-import base href fix)


kakserpom:
## Summary

Commit [88af239](https://github.com/h4ckf0r0day/obscura/commit/88af239be38f2aa11c3c978a4db2a7482f2ed6c0) on the `fix/cookie-jar-disconnect` branch (the dynamic script / lazy-import base-href fix you mentioned in #245) makes obscura panic during page load against `live.deutsche-boerse.com` with:

```
thread 'main' panicked at deno_core-0.350.0/runtime/op_driver/futures_unordered_driver.rs:307:22:
RefCell already borrowed
thread 'main' panicked at library/core/src/panicking.rs:225:5:
panic in a function that cannot unwind
thread caused non-unwinding panic. aborting.
```

The panic fires after the page starts running its scripts (we see the `bootstrap-slider.js` warning print) and aborts the process before any of our user-side `page.evaluate` calls execute.

## Bisection

Same scraper, same test page, same compile mode (release via `rust-script`), same machine:

| Commit | Behavior |
|---|---|
| `0c261ac` — `add XMLHttpRequestEventTarget + XHR dispatchEvent (#245)` | ✅ no panic, XHR onerror fires with CORS-denied (`readyState: 4, status: 0`) |
| `efbf3ab` — `fix XHR super() call order in ES6 class extends` | ✅ no panic, same as above |
| `88af239` — `use base href for dynamic script/module URL resolution` | ❌ panic during page load, every run |

Reverting to `efbf3ab` immediately restores stability.

## Likely cause

The lazy-import fix doesn't itself add any new async work — it just changes URL resolution inside an existing `(async () => { ... })()` block that fetches and executes the dynamic chunk. But because the *correct* URLs now resolve, the Angular runtime actually downloads and starts executing many chunks in parallel (instead of 404'ing on every one). That increased concurrent op activity is, I suspect, surfacing a latent re-entrant borrow in obscura-js's op driver — not a bug in the lazy-import fix itself.

Specifically, the panic at `deno_core-0.350.0/runtime/op_driver/futures_unordered_driver.rs:307:22` is the standard signal of an op future poll attempting to re-enter the op driver's `RefCell` while it's already mutably borrowed by an outer poll. That would happen if a fetched chunk's `then` callback synchronously calls back into a deno op while we're still inside the driver's poll loop for the original fetch.

## Reproduction

```rust
use obscura::Browser;

[...]

PR #275, opened 2026-06-09
fix(js): serialize dynamic module imports + re-land base href for SPA lazy chunks (#246)

zhangyang-crazy-one:
Fixes the deno_core RefCell panic when SPAs dynamically import multiple lazy chunks concurrently.

## Changes

**1. Serialized import queue (bootstrap.js)**
- Added  +  at the top level
- Dynamic `import()` calls now execute sequentially rather than concurrently
- Non-module dynamic scripts (op_fetch_url + eval) also route through the queue
- Inline module data: URLs also go through the queue for consistency

**2. Re-land dynamic import base href resolution**
- Dynamic `<script src>` URLs now resolve against `<base href>` when present
- Falls back to `location.href` per HTML spec
- Previously reverted as commit 88af239 — now safe because concurrency is serialized

## Root Cause

bootstrap.js `appendChild` handler fired `(async () => { await import(fullUrl); })()` fire-and-forget for each dynamically inserted `<script module>`. When Angular SPAs appended 5-8 lazy chunks simultaneously, all `import()` calls entered deno_core's `futures_unordered_driver` concurrently, causing a re-entrant RefCell panic at `futures_unordered_driver.rs:307`.

## Verification

- Syntax check: `new Function(code)` passes
- 65 insertions, 27 deletions across 1 file (bootstrap.js only)
- No Rust changes needed

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -67,29 +67,37 @@ let __dynScriptBusy = false;
 async function __processDynScriptQueue() {
   if (__dynScriptBusy) return;
   __dynScriptBusy = true;
-  while (__dynScriptQueue.length > 0) {
-    const task = __dynScriptQueue.shift();
-    try {
-      if (task.isModule) {
-        await import(task.url);
-      } else {
-        const raw = await Deno.core.ops.op_fetch_url(task.url, "GET", "{}", "", task.pageOrigin, "no-cors");
-        const parsed = JSON.parse(raw);
-        if (parsed.body) {
-          globalThis.__currentScriptNid = task.nid;
-          try { (0, eval)(parsed.body); }
-          catch(e) { console.error('Dynamic script error (' + task.url + '):', e.message); }
-          finally { globalThis.__currentScriptNid = task.prevNid || 0; }
+  // try/finally so the busy flag is always cleared even if a task throws
+  // outside its own guard; otherwise the queue would wedge and silently
+  // block every later dynamic script on the page.
+  try {
+    while (__dynScriptQueue.length > 0) {
+      const task = __dynScriptQueue.shift();
+      try {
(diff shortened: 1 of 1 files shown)


SGavrl:
Reviewed and tested this end to end. The serialized import queue does fix the deno_core RefCell panic from #246, and re-landing base-href resolution is the right call. I found three issues and pushed fixes on top:

1. Base-href threw on a path-only `<base href>`. `<base href="/">` (the standard Angular form) made `new URL(src, baseHref)` throw "Invalid URL", and it wasn't guarded, so it escaped `appendChild` and broke the exact SPAs this targets. The base href is now resolved to absolute against the document URL first, and both that and the final resolution are wrapped. `/`, `/app/`, and absolute bases all resolve correctly now.
2. The queue could wedge. If a task threw outside its own guard, `__dynScriptBusy` stayed true and silently blocked every later dynamic script on the page. The drain is now in try/finally so the flag always clears.
3. Double-fired load event. `dispatchEvent` already invokes the element's `onload` property handler, so the extra explicit `onload()` call fired it twice (this was true before the PR too). Load/error now go through `dispatchEvent` only, which also drops the enqueue-time callback snapshot and reads handlers live off the element. `onload` fires once now.

Tested: obstacle course green on the import stages (es-modules, ssr-hydrate, spa-mini-app), full cargo suite passes, and a `<base href="/">` dynamic append no longer throws.

One follow-up to track separately, not blocking: this serializes `import()`, which removes the known trigger, but the underlying re-entrant borrow lives in deno_core's op driver and `fetch()`/XHR still hit it concurrently. If it can be provoked through those paths we'll want a Rust-side fix. Also, on a long-lived CDP session, if the event loop is cancelled while a queue task is mid-await, the busy flag can stay set until navigation; worth a reset hook later.

Merging. Thanks for tracking down the root cause.

PR #284, opened 2026-06-09
ci: add PR gate (test/clippy/deny/semgrep) + supply-chain hygiene

charlesmamane26-sketch:
## Summary

Turns static-analysis findings into a CI gate and supply-chain config so
regressions are caught on every PR — the repo currently has **no `pull_request`
CI** (workflows only trigger on tags/push). Also fixes a test that never runs
and documents the security posture.

## Changes

- **`.github/workflows/ci.yml`** — PR + push-to-`main` gate:
  - `test`: `cargo clippy --workspace --all-targets` (informational) + `cargo test --workspace`
  - `advisories`: `cargo-deny check advisories bans sources`
  - `semgrep`: `p/rust` ruleset (informational; runs on Linux, where semgrep is supported)
- **`deny.toml`** — accepts the 3 transitive *unmaintained* advisories
  (`bincode`/`fxhash`/`paste` — no fix available, pulled by `deno_core`/`selectors`/`v8`),
  surfaces duplicate versions as warnings, pins crates.io as the only allowed
  source, and seeds a permissive license allow-list.
- **`crates/obscura-net/src/cookies.rs`** — `test_save_load_roundtrip` was
  missing `#[test]`, so the cookie save/load round-trip was **never exercised**.
  Add the attribute so it runs.
- **`SECURITY.md`** — threat model, operator-hardening checklist, and the V8
  prebuilt supply-chain note (`V8_FROM_SOURCE` for high-assurance builds).

## Notes

- Clippy is intentionally **informational** (not `-D warnings`) because the tree
  has pre-existing style lints; tightening it is a follow-up after
  `cargo clippy --fix`.
- The `test` job builds V8 (no `--features stealth`, which tests don't need and
  which keeps CI off the fragile `boring-sys` build); `Swatinem/rust-cache`
  keeps subsequent runs fast.
- License checking is configured in `deny.toml` but not yet enforced by the CI
  command, to avoid false failures until the allow-list is validated against the
  full tree.


SGavrl:
thanks for this, there's clearly a lot of work in it. CI and supply-chain gating isn't really where our focus is right now though, we're heads-down on the engine and real-world rendering, so i'm not going to add this at the moment. not a no forever, something like this could make sense further down the line, just not right now. appreciate the effort.

Issue #307, opened 2026-06-17
Suggestion: add a CI workflow running cargo test / clippy / audit on push and PR


charlesmamane26-sketch:
Hi 👋 — small hygiene suggestion after reading through the codebase (external, read-only review). The security engineering here is genuinely strong — the SSRF deny-set (`is_forbidden_ip` covering loopback, RFC1918, link-local incl. `169.254.169.254`, IPv4-mapped forms), the anti-rebinding DNS resolver, and per-redirect re-validation are well done, and there are no hardcoded secrets.

## The gap

`.github/workflows/` currently contains only:
- `release.yml` — builds on tag
- `docker.yml` — builds/pushes the image on tag

There is **no workflow running the test suite, lints, or a dependency audit on `push`/`pull_request`**. The tests exist (a PR mentions "253 passed") but aren't exercised automatically on contributions, so security-sensitive code (the SSRF guard in particular) can regress silently between releases.

## Suggested `ci.yml` (push + PR)

```yaml
name: ci
on: [push, pull_request]
jobs:
  check:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: dtolnay/rust-toolchain@stable
        with: { components: clippy, rustfmt }
      - run: cargo fmt --all --check
      - run: cargo clippy --workspace --all-targets -- -D warnings
      - run: cargo test --workspace
      - uses: rustsec/audit-check@v2
        with: { token: ${{ secrets.GITHUB_TOKEN }} }
```

(`rustsec/audit-check` flags vulnerable dependencies via the RustSec DB — useful since there's no `cargo audit` step today.)

## Minor related items, take or leave

- `validate_fetch_url` (`crates/obscura-js/src/ops.rs`) accepts the `file` scheme unconditionally, not gated by `--allow-file-access` like the navigation path. It appears inert today (the JS `op_fetch_url` routes through reqwest, which doesn't support `file://`), but aligning it with the navigation gate would remove the inconsistency.
- `.gitignore` is minimal (`target/`, `.DS_Store`); adding `.env*`, `*.pem`, `*.key` would prevent accidental credential commits.

Thanks for the project — really clean work overall.

SGavrl:
Thanks for putting this together. It's not something we want to take on for the project right now, but it's a sensible idea and we'll probably set up CI like this down the line. Going to close it for now. Appreciate the effort.

PR #308, opened 2026-06-17
ci: add workflow running fmt / clippy / test / audit on push and PR

charlesmamane26-sketch:
External suggestion from a quality review (see issue #307). The existing workflows only build on tag, so tests/lints/audit never run on PRs. This adds a CI workflow on push + pull_request running `cargo fmt --check`, `cargo clippy -D warnings`, `cargo test --workspace`, and `cargo audit` (via rustsec/audit-check). Not run in my environment — please verify.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,41 @@
+name: ci
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+env:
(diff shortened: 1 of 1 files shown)


SGavrl:
Thanks for putting this together. It's not something we want to take on for the project right now, but it's a sensible idea and we'll probably set up CI like this down the line. Going to close it for now. Appreciate the effort.

PR #322, opened 2026-06-23
Fix reload and fetch body serialization issues

RezoxP:
This pull request fixes location.reload() to perform a real navigation (previously it was a no-op), and correctly serializes fetch() bodies (FormData, URLSearchParams, Blobs, ArrayBuffers, etc.) with correct Content-Type headers rather than stringifying them blindly. This fixes compatibility issues with many websites


SGavrl:
Hey. The diff on this PR is completely unreadable. It looks like a global auto-format or file recreation has wiped out the git history, making it impossible to locate the actual logic changes.

Please revert the global formatting/whitespace changes so the PR only contains your actual code updates. If you had to refactor or move things around, split it into a pure refactor commit and a separate logic commit so it can be safely reviewed. Thanks.

RezoxP:
fixed it. @SGavrl

PR #347, opened 2026-07-01
fix: protocol-relative URL (//host/path) resolved incorrectly in dynamic script loading

### LINKED ISSUE #345: fix: protocol-relative URL (//host/path) resolved incorrectly in dynamic script loading
## Bug

Protocol-relative URLs (`//host/path`) in `<script src>` attributes are resolved incorrectly. Instead of inheriting the page protocol (https://), the full base URL origin is concatenated with the path:

```
// Expected: https://s1.hdslb.com/bfs/seed/jinkela/kv-sdk/index.js
// Actual:   https://www.bilibili.com//s1.hdslb.com/bfs/seed/jinkela/kv-sdk/index.js
```

This causes `Dynamic script error` and `Unexpected token '<'` errors on any page that loads CDN scripts via protocol-relative URLs (common on Chinese websites like bilibili).

## Root Cause

In `crates/obscura-js/js/bootstrap.js`, the `appendChild` method resolves script URLs with:

```js
fullUrl = src.startsWith('http') || src.startsWith('data:')
  ? src
  : new URL(src, baseUrl).href;
```

The V8 `new URL()` in obscura's build incorrectly resolves `//host/path` against a base URL, producing the doubled origin.

## Fix

Detect `//` prefix and manually prepend the base URL protocol:

```js
fullUrl = src.startsWith('http') || src.startsWith('data:')
  ? src
  : src.startsWith('//')
    ? (new URL(baseUrl).protocol || 'https:') + src
    : new URL(src, baseUrl).href;
```

## Reproduction

```bash
obscura fetch https://www.bilibili.com/ --eval 'new URL("//s1.hdslb.com/path", location.href).href'
```

Before fix: `https://www.bilibili.com//s1.hdslb.com/path` (wrong)
After fix: `https://s1.hdslb.com/path` (correct)

Verified: `Dynamic script error` count drops from multiple to 0 on bilibili homepage.

hupong:
## Summary

Fixes incorrect resolution of protocol-relative URLs (`//host/path`) in dynamic `<script>` loading. This causes `Dynamic script error` and `Unexpected token '<'` on pages loading CDN scripts via `//` URLs (common on Chinese websites like bilibili).

## Root Cause

In `crates/obscura-js/js/bootstrap.js`, the `appendChild` method uses `new URL(src, baseUrl)` to resolve script URLs. The V8 URL implementation incorrectly handles protocol-relative URLs, concatenating the full base origin instead of inheriting just the scheme.

## Fix

Detect `//` prefix and manually prepend the base URL protocol:

```js
: src.startsWith('//')
    ? (new URL(baseUrl).protocol || 'https:') + src
    : new URL(src, baseUrl).href;
```

## Verification

```bash
obscura fetch https://www.bilibili.com/ --eval 'new URL("//s1.hdslb.com/path", location.href).href'
# Before: https://www.bilibili.com//s1.hdslb.com/path (wrong)
# After:  https://s1.hdslb.com/path (correct)
```

Closes #345

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -481,7 +481,9 @@ class Node {
         try {
           fullUrl = src.startsWith('http') || src.startsWith('data:')
             ? src
-            : new URL(src, baseUrl).href;
+            : src.startsWith('//')
+              ? (new URL(baseUrl).protocol || 'https:') + src
+              : new URL(src, baseUrl).href;
         } catch(e) {
           console.error('Dynamic script URL resolve failed (' + src + '):', e.message);
           fullUrl = src;


SGavrl:
Thanks for the PR. I checked this against current main and the code it patches already resolves protocol-relative URLs correctly, so this is a no-op.

Testing the exact case with the current new URL(src, baseUrl) path:

new URL("//s1.hdslb.com/bfs/seed/jinkela/kv-sdk/index.js", "https://www.bilibili.com/").href → "https://s1.hdslb.com/bfs/seed/jinkela/kv-sdk/index.js" (correct)

Both the current code and your protocol + src version produce the identical correct result, no www.bilibili.com//s1.hdslb.com doubling. The doubling you saw on v0.1.9 isn't reproducible on main, so it looks like it was already resolved. new URL is also more correct than manual concatenation for edge cases (userinfo, ports, backslashes), so I'd rather keep it.

Closing this since there's nothing to change, but thanks for digging in with a clear repro. If you can still reproduce the doubling on a build from main, reopen with the obscura commit and the exact script URL and I'll take another look.

PR #411, opened 2026-07-14
feat(css): add lightweight native style computation

evmcheb:
## tl;dr

on a controlled replay of the current live bootstrap 5.3 documentation page (70,809 bytes of html, 1,061 dom nodes, and 293,950 bytes of production css), this pr turns the existing all-cost/no-cssom behavior into an explicit choice:

- `compute` adds real css computation, 2 visible stylesheets, 3,377 retained rules, and the correct bootstrap body color for **+7.19 ms median wall time and only +0.47 mib median rss** versus current upstream obscura.
- `drop` preserves the no-css path while improving on current upstream obscura by **1.88 ms / 9.5% median wall time and 4.11 mib / 12.9% median rss**.
- current upstream obscura uses 31.77 mib and 19.80 ms but exposes **0 stylesheets** and returns no computed bootstrap body color. it pays more ram than drop without providing the fidelity of compute.

| mode | median wall time | vs current | p95 wall time | median rss | vs current | peak rss | css result |
|---|---:|---:|---:|---:|---:|---:|---|
| current upstream obscura (`a921668`) | 19.80 ms | baseline | 21.39 ms | 31.77 mib | baseline | 31.94 mib | 0 sheets; body color empty |
| this pr: `compute` | 26.99 ms | +7.19 ms / +36.3% | 29.16 ms | 32.24 mib | +0.47 mib / +1.5% | 32.47 mib | 2 sheets; 3,377 rules; `#212529` |
| this pr: `drop` | 17.92 ms | -1.88 ms / -9.5% | 25.19 ms | 27.66 mib | -4.11 mib / -12.9% | 27.77 mib | 0 sheets; synthetic defaults |

methodology: upstream main and this pr were both release builds. the page was captured from [the live bootstrap introduction page](https://getbootstrap.com/docs/5.3/getting-started/introduction/), its two real stylesheet responses were replayed unchanged over loopback, scripts were removed to isolate css cost, and each mode ran with two warmups plus twelve interleaved cold-process samples. rss comes from `/usr/bin/time -l`.

## why this is necessary

fetching stylesheet urls is not enough browser emulation. before this change, author css was not represented in the cascade or cssom, so `getComputedStyle()`, `document.styleSheets`, `link.sheet`, media-query-dependent values, custom properties, and cssom mutations could disagree with what a browser exposes.

that creates two practical problems:

1. compatibility: applications frequently use computed styles, css variables, and media queries while hydrating or deciding which component path to run. returning only synthetic defaults can change application behavior.
2. detectability: always discarding css is fast, but a site can directly verify that a loaded stylesheet did not affect computed style or appear in cssom.

the default `compute` mode closes that gap without adding layout, painting, fonts, or animation. the explicit `drop` mode preserves the lowest-overhead path for workloads that do not need css fidelity. both paths keep untrusted stylesheet bodies out of executable javascript and remove the former need for css-to-v8 transfer.

## measured impact

methodology: release build, local deterministic http fixtures, two warmups, twelve interleaved cold-process samples per case. wall time is median process time. rss is the highest `/usr/bin/time -l` peak observed across the samples. css overhead is measured against the empty fixture for the same mode.

the reproducible corpus, exact source pins, licenses, hashes, runner, and json output are in [obscura-benchmark #2](https://github.com/h4ckf0r0day/obscura-benchmark/pull/2).

| corpus | css bytes | retained rules | compute median | drop median | wall time saved | compute peak rss | drop peak rss | ram saved | css-only overhead reduction |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| bootstrap 5.3.8 | 280,311 | 2,664 | 24.85 ms | 13.68 ms | 44.9% / 11.17 ms | 30.25 mib | 21.19 mib | 30.0% / 9.06 mib | 90.0% |
| bulma website 1.0.4 | 1,430,123 | 6,724 | 62.13 ms | 15.59 ms | 74.9% / 46.54 ms | 50.47 mib | 23.50 mib | 53.4% / 26.97 mib | 93.7% |
| primer 22.3.0 | 1,061,415 | 3,231 | 44.31 ms | 18.01 ms | 59.4% / 26.30 ms | 45.03 mib | 23.62 mib | 47.5% / 21.41 mib | 82.5% |
| patternfly 6.6.0 | 1,993,226 | 6,749 | 53.83 ms | 14.63 ms | 72.8% / 39.20 ms | 53.61 mib | 26.05 mib | 51.4% / 27.56 mib | 94.7% |
| synthetic 160 kib | 163,840 | 3,715 | 24.30 ms | 14.51 ms | 40.3% / 9.79 ms | 29.86 mib | 21.17 mib | 29.1% / 8.69 mib | 82.6% |
| synthetic 1.875 mib | 1,966,080 | 41,655 | 129.32 ms | 26.02 ms | 79.9% / 103.30 ms | 92.44 mib | 25.97 mib | 71.9% / 66.47 mib | 88.4% |
| synthetic 10 mib | 10,485,760 | 50,000 rule cap | 240.40 ms | 75.95 ms | 68.4% / 164.45 ms | 114.47 mib | 50.97 mib | 55.5% / 63.50 mib | 72.1% |

across the four pinned production bundles, the benchmark parses 4,765,075 css bytes and retains 19,368 rules. every locked probe passes for computed values, stylesheet counts, exact rule counts, custom-property resolution, media behavior, and dynamic cssom mutation.

for every corpus at or above 1 mib, drop mode is both faster and lower-rss than compute mode. observed savings range from:

- 59.4% to 79.9% lower median wall time
- 47.5% to 71.9% lower peak rss
- 72.1% to 94.7% less css-only time above the empty fixture

the empty fixture is 12.42 ms in compute mode and 12.44 ms in drop mode, a 0.02 ms difference. this indicates that the mode plumbing and native style registry add effectively no measurable cost when a page has no stylesheets.

the earlier 10 mib retained-body prototype measured 55.37 ms and 91.58 mib versus 21.99 ms and 34.45 mib when dropped: 60.3% less wall time and 62.4% less rss. that result motivated making the low-memory path explicit rather than always retaining stylesheet bodies.

## implementation

- adds public css mode configuration through the rust api, browser context, and global `--css-mode` cli option
- implements a native rust author cascade with selector indexing and a 1,024-entry lazy computed-style cache
- supports specificity, `!important`, inheritance, custom properties, recursive `var()` fallback, common shorthands, media queries, supports blocks, layers, imports, and cssom mutation
- keeps static external sheets in document order and handles dynamic imports without css body transfer through v8
- exposes the required stylesheet/cssom surfaces and invalidates the style epoch on dom and stylesheet mutations
- caps decoded css at 16 mib, rules at 50,000, sheets at 256, and import nesting at eight
- keeps `drop` mode bodyless after the response has been fully received while preserving request and lifecycle behavior

## verification

- 213 nextest cases passed
- release build passed
- 33/33 obstacle-course stages passed
- all production and synthetic correctness probes passed
- all performance gates passed
- no untrusted css is interpolated into executable javascript

depends on #410.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1707,6 +1707,7 @@ dependencies = [
  "markup5ever",
  "precomputed-hash",
  "selectors",
+ "serde",
  "servo_arc",
  "thiserror 2.0.18",
  "tracing",
--- a/crates/obscura-browser/src/context.rs
+++ b/crates/obscura-browser/src/context.rs
@@ -3,6 +3,71 @@ use std::sync::Arc;
 
 use obscura_net::{CookieJar, ObscuraHttpClient, RobotsCache};
 
+/// Controls how stylesheet responses are handled.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub enum CssMode {
+    /// Parse stylesheets and expose a lightweight, non-rendering cascade.
+    #[default]
+    Compute,
+    /// Fetch top-level stylesheets for lifecycle fidelity, then discard them.
+    Drop,
+}
+
+impl std::str::FromStr for CssMode {
+    type Err = String;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        match value.to_ascii_lowercase().as_str() {
+            "compute" => Ok(Self::Compute),
+            "drop" => Ok(Self::Drop),
+            _ => Err(format!(
+                "invalid CSS mode '{value}' (expected compute or drop)"
+            )),
+        }
+    }
+}
+
+impl std::fmt::Display for CssMode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(match self {
+            Self::Compute => "compute",
+            Self::Drop => "drop",
+        })
+    }
+}
+
+impl CssMode {
+    pub fn process_default() -> Self {
+        std::env::var("OBSCURA_CSS_MODE")
+            .ok()
+            .and_then(|value| value.parse().ok())
+            .unwrap_or(Self::Compute)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct BrowserContextOptions {
+    pub proxy_url: Option<String>,
+    pub stealth: bool,
+    pub user_agent: Option<String>,
+    pub storage_dir: Option<PathBuf>,
+    pub allow_private_network: bool,
+    pub css_mode: CssMode,
+}
+
+impl Default for BrowserContextOptions {
+    fn default() -> Self {
+        Self {
+            proxy_url: None,
+            stealth: false,
+            user_agent: None,
+            storage_dir: None,
+            allow_private_network: false,
+            css_mode: CssMode::process_default(),
+        }
+    }
+}
+
 pub struct BrowserContext {
     pub id: String,
     pub cookie_jar: Arc<CookieJar>,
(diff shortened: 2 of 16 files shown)


SGavrl:
Closing this for now. I’m not convinced a native CSS engine is the right direction for Obscura. It adds a large amount of complexity and maintenance surface, while the useful compute path is 36.3% slower than current main. The drop path is faster, but it mostly preserves the behavior we already have. I don’t think the fidelity gain justifies the design and performance cost right now.

Issue #374, opened 2026-07-09
Vite dev React Refresh inline module times out, root stays empty while preview renders


slovx2:
### Before you start

- [x] I searched existing issues and this is not a duplicate.
- [x] I tested against the latest release or `main`.
- [x] This is not a security vulnerability.

### obscura version or commit

`obscura 0.1.9` official release binary (`obscura-aarch64-macos.tar.gz`).

Related existing issues I found:

- #205 is close (`#root` stays empty for a Vite/React SPA), but that was fixed by #212 and was about external module bundles being registered with empty source.
- #361 / #371 are about heavy SPA script-execution budget. This repro is a tiny Vite dev app, and the injected React Refresh inline module still times out with `OBSCURA_SCRIPT_DEADLINE_MS=60000`.

### OS and architecture

macOS arm64:

```text
Darwin ... RELEASE_ARM64_T6030 arm64
Node v23.2.0
npm 10.9.0
Google Chrome 149.0.7827.201
```

### How are you using obscura?

CLI (`fetch`). I also saw the same symptom through the CDP server in a larger local Vite/React app.

### Stealth mode

Off.

### Reproduction

Create a minimal Vite dev app:

```json
{
  "type": "module",
  "scripts": {
    "dev": "vite --host 127.0.0.1 --port 5480",
    "build": "vite build",
    "preview": "vite preview --host 127.0.0.1 --port 5481"
  },
  "dependencies": {
    "@vitejs/plugin-react-swc": "4.2.1",
    "vite": "7.2.2",
    "react": "19.2.0",
    "react-dom": "19.2.0"
  },
  "devDependencies": {}
}
```

`vite.config.js`:

```js
import react from '@vitejs/plugin-react-swc';
import { defineConfig } from 'vite';

export default defineConfig({
  plugins: [react()],
});
```

`index.html`:

```html
<!doctype html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>Obscura Vite Dev Repro</title>
  </head>
  <body>
    <div id="root"></div>
    <script type="module" src="/src/main.jsx"></script>
  </body>
</html>
```

`src/main.jsx`:

```jsx
import React from 'react';
import { createRoot } from 'react-dom/client';

function App() {
  return (
    <main>
      <h1>Obscura Vite Dev Repro Mounted</h1>
      <input aria-label="probe" defaultValue="ready" />
    </main>
  );
}

createRoot(document.getElementById('root')).render(<App />);
```

[...]

Issue #62, opened 2026-04-28
CDP control plane (`/json/version`, `/json`) becomes unreachable while JS engine is busy on real-world pages


gte619n:
## Summary

When `obscura serve` navigates a page whose bootstrap JS is heavy or hits unimplemented APIs, the **HTTP control plane stops responding**: `GET /json/version` and `GET /json` time out and the CDP WebSocket can no longer be obtained. A second Playwright client trying to `connect_over_cdp(...)` while the first is still loading hits Playwright's 180-second wall and gives up.

Recovery requires `kill -9` — graceful Ctrl+C is not delivered while the runtime is wedged.

## Repro

```bash
$ obscura serve --port 9223        # release or debug, makes no difference
```

```python
# A: connect, navigate to ANY non-paywalled real article — works
from playwright.sync_api import sync_playwright
with sync_playwright() as p:
    b = p.chromium.connect_over_cdp("http://127.0.0.1:9223")
    page = b.contexts[0].new_page()
    page.goto("https://arstechnica.com/", wait_until="domcontentloaded", timeout=25000)
    page.wait_for_timeout(1500)
    print(page.title())   # eventually prints
```

```bash
# B: meanwhile, in another terminal:
$ curl -m 5 http://127.0.0.1:9223/json/version
curl: (28) Operation timed out after 5001ms
```

`(B)` should always succeed regardless of what `(A)` is doing. On Chromium / `chromium --remote-debugging-port=9223`, it does.

A second client doing `chromium.connect_over_cdp("http://127.0.0.1:9223")` from a fresh process during this window times out the same way:

```
TimeoutError: BrowserType.connect_over_cdp: Timeout 180000ms exceeded.
Call log:
  - <ws preparing> retrieving websocket url from http://127.0.0.1:9223
```

## Root cause (suspected)

`crates/obscura-cli/src/main.rs` uses `#[tokio::main(flavor = "current_thread")]` and `crates/obscura-cdp/src/server.rs` runs everything inside a single `LocalSet`. The CDP processor task is `spawn_local`'d on the same thread as the HTTP listener. A long synchronous JS evaluation (e.g. a `Runtime.evaluate` that pulls in arstechnica's ad-tech and loops on missing APIs) blocks the local executor, and HTTP `accept` doesn't get a chance to run until the JS turn yields.

## Concrete trigger I observed

While loading `https://arstechnica.com/` the bootstrap throws repeatedly:

```
[...]

PR #435, opened 2026-07-16
fix(cdp): confine V8 isolates per connection to end the #430 concurrency abort

### LINKED ISSUE #430: Concurrent CDP navigations still abort with the #19 V8 fatal (heap->isolate() == TryGetCurrent()) at concurrency 2: v8_lock released across network-op .await points
## Before you start
- [x] I searched existing issues and this is not a duplicate. (Related: #19, closed. This is the residual path documented in `server.rs:639-658`, still reproducible on v0.1.10 at concurrency 2.)
- [x] I tested against the latest release. (`h4ckf0r0day/obscura:latest`, banner v0.1.10.)

## obscura version or commit
v0.1.10 (Docker Hub `h4ckf0r0day/obscura:latest`, startup banner "Headless Browser v0.1.10", reports `Chrome/145.0.0.0` on `/json/version`).

## OS and architecture
Linux x86_64, official Docker image (default build, no `--features stealth`).

## How are you using obscura?
CDP server (raw CDP over the browser WebSocket endpoint).

## Stealth mode
Off (default build).

## Reproduction

Drive the browser endpoint (`ws://127.0.0.1:9222/devtools/browser`) from N parallel workers. Each worker independently:

1. `Target.createTarget { url: <a real /maps/place/ page> }`
2. `Target.attachToTarget { flatten: true }`
3. poll `document.readyState` until `interactive`/`complete`
4. `Runtime.evaluate` a small DOM-reading IIFE (reads a couple of `button[data-item-id]` / `a[data-item-id]` nodes)
5. `Target.closeTarget`

- **Concurrency 1:** 40/40 succeed, ~67 pages/min, stable, container stays up.
- **Concurrency 2:** every request times out and **the whole process aborts** (`docker` shows `Exited (133)`). Reproduced deterministically across multiple restarts, and at concurrency 4 as well.

The pages are network-heavy real sites (each navigation triggers many subresource fetches), which matches #19's repro against `scrapegraphai.com`.

## Fatal output (container stderr)

```
# Fatal error in , line 0
```
(followed by a native backtrace; the release binary is stripped so frames are bare `obscura(+0x...)` offsets.)

This is the exact assertion from #19.

## Expected vs actual
- **Expected:** concurrent CDP clients/targets either run serialized or fail gracefully; the server process must not abort.
- **Actual:** the process aborts (`abort(3)` from V8, no CDP messages after the crash), killing every in-flight page.

No. The identical driver against headless Chrome (`--headless=new --remote-debugging-port`) at concurrency 1/4/8 completes 40/40 every time with no crash. This is not anti-bot/CAPTCHA (single navigations to the same URLs succeed fine at concurrency 1).

## Notes toward root cause
The #19 fix added the global V8 serialization lock (`crates/obscura-js/src/v8_lock.rs`, acquired in `crates/obscura-cdp/src/dispatch.rs:296-301`). But `crates/obscura-cdp/src/server.rs:639-658` documents that this `tokio::sync::Mutex` is **released around the internal `.await` points of V8 ops** (e.g. `op_fetch_url`'s `resolve_rx.await`) while the Isolate is still entered. A navigation to a real page spends most of its time in exactly those network `.await`s, so with two concurrent navigations the enter/exit contiguity the lock is meant to guarantee is not held, and `v8::Context::Exit` on the wrong Isolate fires `V8_Fatal`. Because the trigger is per-navigation network work rather than raw page count, it reproduces at concurrency 2, not only ≥5 as in #19.

Happy to test a branch against this repro.

Sycatle:
## What

Ends the #430 concurrency abort (`Check failed: heap->isolate() == Isolate::TryGetCurrent()`) and lets CDP connections run in parallel. Four commits:

1. `fix(js)`: drive the event loop cancellation-safely (pump via `poll_event_loop` instead of dropping `run_event_loop` under a `tokio::time::timeout`, which left an isolate entered on the thread). Hardening across `obscura-js` and the `page.rs` settle loops.
2. `fix(cdp)`: run each connection on its own OS thread with its own runtime, `LocalSet`, and `CdpContext`, sharing one `Arc<BrowserContext>`. Every page's isolate is confined to one thread, so V8's per-thread invariant makes the cross-connection abort impossible by construction. This is the "Option 2" fix named in `obscura-js/src/v8_lock.rs`.
3. `test(cdp)`: a heavy-page regression test that reproduces the abort on the old single-thread server and passes on the new one.
4. `perf(cdp)`: scope the V8 lock per connection (was process-wide), so connections no longer serialize on one mutex.

## Why the old regression test missed it

`concurrent_navigations_with_fetch` uses `example.com` / `data:` URLs, which settle before the per-poll settle timeout can fire, so the abort never reproduced. The new test serves a local page with slow subresources plus a `setInterval` and drives it from several independent connections.

## Root cause

Several event-loop operations bounded `run_event_loop` / `mod_evaluate` with `tokio::time::timeout` or `select!`. When the timer won, tokio dropped that future mid-poll; deno_core enters the isolate while polling, so the drop left the isolate entered. Under concurrent CDP work a second page's `execute_script` then tripped V8's check and `abort(3)`'d. Serializing at the single-thread layer could not fix it reliably (the pattern is pervasive and partly inside deno_core), so isolates are confined to threads instead.

## Two V8 multi-thread gotchas handled

- Serialize isolate construction with a process-wide lock (the leaptiering `JSDispatchTable` setup is not safe to run from several threads at once).
- Warm up V8 once on the main thread before any connection thread creates an isolate; creating the first isolate off the main thread segfaults in `InitializeBuiltinJSDispatchTable` (confirmed via gdb).

## Verification

- Obstacle course: **33/33**.
- `cargo nextest run -p obscura-cdp`: **76/76**, including the concurrent-navigation, control-plane, and new heavy-page tests.
- Server driven at concurrency 1/2/4/8 against heavy pages: no abort, no segfault, stays up.
- Navigation throughput on a local CPU-bound fixture: flat (~75/s across concurrency) with the old global lock; roughly 2.5x at concurrency 8 with the per-connection lock.

Closes #430.


SGavrl:
The concurrency fix works: all three concurrency tests pass and the obstacle course remains 33/33. The remaining blocker is unbounded thread-per-connection scaling. At 100 idle connections this creates 112 OS threads and reserves ~39 GB of virtual address space; main uses 2 threads and ~141 MB. This needs a connection cap or bounded worker design before merge.

Sycatle:
Rebased onto `72d66bc` and pushed two commits. Before the fix, though: I think the measurement the blocker rests on is comparing two different things, and it took me a while to see it.

## Idle connections do not start V8

`main` never builds an isolate for a connection that just sits there. Same server, same build, one connection:

```
main, idle connection:      2 threads    10 MiB RSS     124 MiB VSZ
main, one JS navigation:   17 threads    29 MiB RSS   34494 MiB VSZ
```

That 33 GiB step is V8's process-wide sandbox, and `main` pays it the moment it runs any JS at all. So `main`'s "2 threads / 141 MB" is not a lean baseline — it is a server that has not started yet. Measured against 100 connections **each running JS**:

| | threads | RSS | VSZ |
|---|---|---|---|
| `main` | 17 | 489 MiB | 83.6 GiB |
| this branch, before | 118 | 519 MiB | 90.0 GiB |
| this branch, now | 118 | 505 MiB | 83.5 GiB |

`main` is not flat under load either — it reaches 83.6 GiB by the same route. At equal work this branch now reserves marginally less address space than `main` and 3% more RSS. What it genuinely costs is 118 threads against 17, and that is the thing worth capping.

Harness is `reliability/conn-scale.py`, run with `--navigate`. It did not exist — nothing in obscura-benchmark measured the serve path, `compare/scale.py` starts N cold `fetch` processes — so I wrote it: [asap-cool/obscura-benchmark@feat/conn-scale](https://github.com/asap-cool/obscura-benchmark/tree/feat/conn-scale/reliability). Happy to PR it over if you want it in the repo.

Two bugs of my own in it first: I was sampling the idle floor before the warm-up isolate had been dropped, and my first pass compared idle connections, which is how I got a per-connection cost that came out *negative* — that should have stopped me sooner than it did.

## The two commits

`mallopt(M_ARENA_MAX, 2)` at `serve` startup, and `--max-connections` (default 128).

The arenas were the whole per-connection memory term: glibc gives each thread its own 64 MiB arena, ~65 MiB of the ~66 MiB a connection reserved. Capping them takes it to 2.0 MiB. It matters more for RSS under sustained load than for the reservation — freed chunks stay in their arena instead of returning to the OS, so RSS tracks peak concurrency and never comes down, which is what first sent me looking. In the container against Google Maps, four concurrent connections per round: 350 / 619 / 826 MiB over three rounds uncapped and still climbing linearly, versus 166 / 235 / 269 MiB capped. Two arenas cost no throughput I can measure — 8 connections x 12 navigations, 1.53s against 1.50s.

The cap reserves a slot in the accept loop and releases it from a guard on thread unwind, so panics and the early returns before the runtime is built cannot leak one. Over-limit clients get a `503` with `X-Obscura-Reason: max-connections` instead of today's silent drop, and because the refusal is in the WS dispatch loop the HTTP control plane from #62 stays reachable at the cap (checked).

I also tried the obvious lever first — an explicit 2 MiB `stack_size`, since nothing in the tree sets one and the default is 8 MiB — and it moved **neither** RSS nor VSZ measurably. Dropped it rather than carry a change I could not put a number against.

## Gates

- `cargo nextest run -p obscura-cdp` — 83/83, with a new `max_connections_cap` test covering accept-to-the-limit, the refusal status, and slot release on close. I checked it fails when the release is removed; the wedge-after-N-connections failure mode is otherwise invisible.
- obstacle-course 33/33.
- `obscura-js` unit tests are 28 passed / 68 failed here, but identically on `main`, so that looks like my environment rather than this branch.

Two calls I would rather leave to you: whether 128 is the default you want, and whether `mallopt` belongs in `obscura-cdp` at all — it is process-global and this is a library crate, so I am happy to move it up into `obscura-cli` if you would rather the library never touched it.

Sycatle:
Went back over my own earlier commits on this branch before you spend more time on them, and one of them should not be here. Three commits pushed.

## `45df429` reverted — the premise was wrong

That commit replaced `tokio::time::timeout(budget, run_event_loop())` with a hand-rolled `poll_fn(poll_event_loop)` loop, on the theory that a fired timeout drops the `run_event_loop` future mid-poll and leaves the isolate entered. In deno_core 0.350:

```rust
pub async fn run_event_loop(&mut self, opts) -> Result<(), CoreError> {
    poll_fn(|cx| self.poll_event_loop(cx, opts)).await
}
```

It is the same code. And the mechanism cannot happen anyway — `poll` is a synchronous `&mut self` call, a future is only ever dropped *between* polls, and the `HandleScope`/`ContextScope` are stack locals inside `poll_event_loop` destroyed before it returns. I was wrong about that in the #430 write-up too; what actually fixes the abort is the thread-per-connection confinement, which is V8's per-thread check doing the work.

It was not a harmless no-op, which is the part I would have wanted caught in review:

- **`resolve_promises_until` lost ~50x throughput.** `tick_ms` silently changed meaning — from "pump the event loop for up to this long, return early on idle" to "do one poll, then sleep this long". With the 1→50ms backoff that settles at 20 event-loop turns per second. An awaited promise chain needing many turns crawls, and once it blows the budget the caller reads a stale `__obscura_await_meta` with no error.
- **`drive_module_eval` swallowed event-loop errors** — `e?` became `let _ = poll_event_loop(...)`. A module stalled on a top-level await used to exit immediately with a logged error; instead it spun for the whole budget logging nothing. That is exactly the #374 starvation the comment above it claims to prevent.

The `drive_module_eval` extraction was worth keeping (it does dedupe `load_module` and `load_inline_module`), so the revert keeps the helper and restores the original `select!` body. The #430 regression test still passes without any of it, which is the point.

## Two more

**The regression test asserted nothing.** Its whole value is the fixture keeping the settle loop pumping, but nothing checked the fixture was ever reached. I pointed `page_url` at a dead port: it passed, in 0.6s instead of 1.25s — and that degraded version no longer reproduces the abort on the pre-fix tree. So a fixture regression would have quietly turned it back into the `data:`-URL test it was written to replace. It now counts the requests the fixture serves and asserts on them (two per client — the engine does not fetch the ``, only the document and `/slow.js`). Verified the assertion kills the dead-port mutant.

**Shutdown lost cookies.** Connection threads are detached and never joined, so `save_cookies()` ran on the main thread while they were still serving. Before this branch, the single processor saved on its own way out, ordered against all connection work on one LocalSet. Now it drains on the live-connection counter first. Being straight about the scope: this bounds the window rather than closing it, because a connection's frame reader is parked in a socket read and never observes `notify_waiters` — so a client holding a connection open costs the full 3s deadline and then gets a warning. Waking readers on shutdown is a separate change and not this PR's business. Measured 2ms with no connections, 3005ms with one held, against `docker stop`'s 10s.

## State

`cargo nextest run -p obscura-cdp` 83/83, obstacle-course 33/33, both after all three.

I left the revert as its own commit rather than rewriting `45df429` out of the branch, since the reasoning seemed worth keeping in the history — say the word if you would rather have a clean nine.

Separately, and not from this branch: `concurrent_navigations_with_fetch` hits `https://example.com` in the default suite. All three concurrency test headers document an `--ignored` gate, but no `#[ignore]` attribute exists anywhere in the tree — the only occurrence is inside a comment. I repointed it at an unresolvable host and it still passed, because it matches on `id` and ignores `result.errorText`. Happy to open that as its own issue.

PR #442, opened 2026-07-17
feat(js): size the scroll box from content so virtualized feeds paginate

### LINKED ISSUE #441: Virtualized SPA feeds never paginate: scroll containers expose no scrollable range
## Summary

Even with functional `scrollTop`/`scrollBy`/`scrollTo` (proposed in #431), scroll-driven lazy loaders never page in more rows. A virtualized list such as the Google Maps `/maps/search/` results feed renders its first batch and then stalls, no matter how much the harness scrolls.

## Root cause

Without a layout engine, every non-root element reports a degenerate scroll box:

- `scrollHeight === clientHeight === 20` for any container (`Element` getters in `crates/obscura-js/js/bootstrap.js`), so `scrollTop` has a max range of 0 and can never advance.
- The `scrollTop` setter does not emit a `scroll` event on direct assignment, only the scroll methods do. Real lazy loaders do `el.scrollTop = el.scrollHeight`, so their scroll handler never fires.

The list's "load more when scrolled near the bottom" logic therefore never triggers.

## Reproduction

Drive a CDP session (raw or Puppeteer) to `https://www.google.com/maps/search/<query>` with `waitUntil: networkidle2` (so the event loop is pumped past load), then repeatedly set the feed's `scrollTop` and pump. The count of `a[href*="/maps/place/"]` plateaus at ~30 and never reaches the full result set (~110 in Chrome).

## Proposed fix

With no layout engine we can still synthesize a plausible scroll geometry:

- Derive `scrollHeight` from the descendant count so it grows as rows load (`childElementCount` is unreliable on these synthetic subtrees; count descendants).
- Expose the viewport size as the client box so `scrollTop` has real range.
- Clamp `scrollTop` to the synthetic scroll box and fire a single `scroll` event on direct assignment (coalesced with the scroll methods).

With that, the Maps feed pages in its full result set (~30 -> ~113 rows) and obstacle-course stays 33/33. This builds on #431.

Sycatle:
Closes #441.

**Stacked on #431** (functional Element scroll methods). This PR's first commit is #431's; the second is the change under review here. Please review/merge after #431 — once it lands I will rebase so only the geometry commit remains.

## Problem

With #431, `scrollTop`/`scrollBy`/`scrollTo` are functional, but every element still reports a 20px scroll box (`scrollHeight === clientHeight`), so `scrollTop` has no range and scroll-driven lazy loaders never page in more rows. A virtualized list such as the Google Maps `/maps/search/` feed renders its first batch (~30 rows) and stalls. Full detail in #441.

## Change

`crates/obscura-js/js/bootstrap.js`, `Element`:

- **`scrollHeight`** is derived from the descendant count (`querySelectorAll('*').length`) so it grows as rows load. `childElementCount` is unreliable on these synthetic subtrees. Leaf elements keep the old 20px height.
- **`clientWidth`/`clientHeight`/`scrollWidth`** expose the viewport size for every element, giving `scrollTop` real range. The click/hit-test path uses `getBoundingClientRect`, not these, so actionability is unaffected.
- The **`scrollTop` setter** clamps to the synthetic scroll box (so virtualized lists render the window at the content bottom instead of overshooting into empty space) and fires a `scroll` event on direct assignment. Lazy loaders do `el.scrollTop = el.scrollHeight` and rely on that event.
- **`scrollTo`/`scrollBy`** suppress the per-axis setter events so a full operation emits a single `scroll` event, the way a real browser coalesces one scroll per movement.

## Validation

- Google Maps `/maps/search/` (driven with `waitUntil: networkidle2` so the loop is pumped past load): the feed pages from ~30 to the full result set (~113 unique `/maps/place/` rows) instead of plateauing at ~31.
- obstacle-course **33/33**, median latency ~12ms (no perf regression).
- `cargo nextest run -p obscura-js -p obscura-browser`: failing tests are identical to the `fix/element-scroll-methods` baseline (pre-existing `runtime::tests` unit-harness failures); this change introduces none.

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -2086,21 +2086,53 @@ class Element extends Node {
   // documentElement / body / window expose VIEWPORT geometry, not their own content box.
   // Puppeteer's #clickableBox clips boxes to document.documentElement.clientWidth/Height;
   // returning 100x20 there made every element appear off-screen and broke .click().
-  get clientWidth() { return this._isViewportRoot() ? (globalThis.innerWidth || 1280) : 100; }
-  get clientHeight() { return this._isViewportRoot() ? (globalThis.innerHeight || 720) : 20; }
-  get scrollWidth() { return this._isViewportRoot() ? (globalThis.innerWidth || 1280) : 100; }
-  get scrollHeight() { return this._isViewportRoot() ? (globalThis.innerHeight || 720) : 20; }
+  // With no layout engine, a scroll container needs a viewport-sized client box
+  // and a content-sized scroll box, otherwise scrollTop has no range and the
+  // scroll-driven lazy loaders that PR #431 unblocked still never page in more
+  // rows. Expose the viewport size as the client box for every element (the
+  // click/hit-test path uses getBoundingClientRect, not this) and derive the
+  // scroll box from the subtree size below.
+  get clientWidth() { return globalThis.innerWidth || 1280; }
+  get clientHeight() { return globalThis.innerHeight || 720; }
+  get scrollWidth() { return globalThis.innerWidth || 1280; }
+  get scrollHeight() {
+    if (this._isViewportRoot()) return (globalThis.innerHeight || 720);
+    // Approximate content height from the descendant count so virtualized lists
+    // (which page in more rows as scrollTop nears scrollHeight) can advance past
+    // their first batch. childElementCount is unreliable on these synthetic
+    // subtrees, so count descendants. Leaf elements keep the old 20px height.
+    let n = 0;
+    try { n = this.querySelectorAll('*').length; } catch (e) {}
+    const synthetic = n * 40;
+    return synthetic > 20 ? synthetic : 20;
+  }
   _isViewportRoot() {
     const t = this.tagName;
     return t === 'HTML' || t === 'BODY';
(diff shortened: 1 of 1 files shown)


SGavrl:
This cannot merge with querySelectorAll('*') inside scrollHeight. On a 5,000-node container, 200 reads take 194–240 ms versus ~0.1–1 ms on main. scrollTop writes have the same problem because the setter reads scrollHeight. Changing every element’s client size to the viewport is also too broad. Please cache/update geometry on mutations and preserve non-root element dimensions.

Sycatle:
Closing this — #441 settled it. Scroll geometry belongs in the rendering engine, and a synthetic geometry system bolted onto the DOM getters is the wrong shape for the tree regardless of whether I fix the cost you flagged.

For the record, the perf critique was right and I am not carrying it as-is: `querySelectorAll('*')` inside `scrollHeight` is a hot-getter DOM scan, and `scrollTop` writes pay it too because the setter reads `scrollHeight`. Changing every element's client size to the viewport was also too broad — that one was me reaching for the shortest patch that made Maps paginate, not a considered choice.

We will carry this downstream until the renderer lands. Happy to test virtualized feeds against it when there is something to point at — Google Maps search results are a good adversarial case: the feed pages in on scroll and the container never exposes a scrollable range today.

PR #445, opened 2026-07-18
upgrade: deno_core 0.350 -> 0.411, v8 137.3.0 -> 150.4.0

gilbertwong96:
## Why

V8 137.x uses the `initial-exec` TLS model that breaks `cdylib` (`-shared`) linking on Linux (`R_X86_64_TPOFF32` relocation errors). This prevents building shared libraries (like Rustler NIFs) that depend on obscura on Linux.

v8 149+ includes the fix from [rusty_v8 PR #1911](https://github.com/denoland/rusty_v8/pull/1911) (`v8_monolithic_for_shared_library`), which switches to `local-dynamic` TLS, making the prebuilt static library safe to link into shared libraries on Linux.

## Changes

- `deno_core` 0.350 → 0.411 (V8 routed through `deno_v8` 0.3)
- `deno_error` 0.6 → 0.7
- `v8` 137.3.0 → 150.4.0
- rebased onto current `main`, so the #946–#949 render transport and CDP work is included

### API migrations

- `ModuleLoader::load` — `ModuleLoadReferrer` + `ModuleLoadOptions` instead of separate `is_dyn_import` / `RequestedModuleType` params
- async ops: `#[op2(async)]` → plain `#[op2]` / `#[op2(fast)]` (0.408 eager-polls async fns); `#[global]` op args are now `#[scoped]`
- `handle_scope()` → `v8::scope!` / `tc_scope!` (pinned-scope model)
- `queueUserTimer` / `cancelTimer` → `createTimer` / `cancelTimer` (removed in 0.408)
- `JsError::from_v8_exception` returns `Box<JsError>`
- isolates constructed outside tokio enter a fallback tokio runtime (V8 149+ posts delayed foreground tasks; deno_core aborts without a registered handle)

## Validation

- `cargo build --release -p obscura-cli --bins --features render` passes; `cargo check --features render,stealth` passes.
- `cargo nextest run --release --features render --no-fail-fast` (1695 tests): 1692 pass. The failures are timing-sensitive and reproduce on unmodified `main`: three browser tests (`render_resource_loads_share_one_page_wide_concurrency_limit`, `timers_inside_a_fixed_wait_observe_bytes_that_land_during_it`, `a_miss_created_inside_an_awaited_expression_loads_during_the_wait`) pass 3 of 9 isolated runs on `main` and 3 of 9 here; `mcp_client` tests fail only under full-suite load and pass in isolation on both.
- Obstacle course: 33/33 once the `observer-intersection` expectation is updated to Chrome parity (`io:10`); the other 32 stages pass unchanged. With the current fixture that stage fails here **and on unmodified `main`**: it expects the refire-every-rendering-opportunity behavior 917da17 replaced, and real Chrome loads 10 items on that page.
- `test_wasm_instantiate_streaming_uses_response_array_buffer` passes; `_installWasmStreamingFallback()` runs from the realm init path.

## Rendering

Not applicable — no layout, paint, screenshot, screencast, or PDF behavior changes.

## Performance

Latency neutral: a 4e7-iteration JS loop through `--eval` takes 55 ms here vs 54 ms on `main`.

Peak RSS, interleaved A/B with the same release build and pages, 7 runs each:

| page | this branch | main | delta |
| --- | --- | --- | --- |
| static, no JS | 28.41 MiB | 27.64 MiB | +0.77 |
| light JS | 28.44 | 27.67 | +0.77 |
| React render | 29.14 | 28.22 | +0.92 |
| 5000-row DOM build | 53.80 | 52.11 | +1.69 |

The JS heap after a full GC is identical (2.7 MB used / 3.8 MB committed), so the delta is not page data. `vmmap`: V8 anonymous heap pages +384 KiB, native malloc +208 KiB, kernel page tables +208 KiB, `__TEXT` unchanged. On JS-heavy pages it is optimized code — `--no-turbofan` removes the delta entirely (−0.25 MiB), `--jitless` leaves +0.11 MiB.

No speed-neutral way back to baseline was found: `--flush-bytecode`, `--max-optimized-bytecode-size` and `--no-concurrent-recompilation` change nothing, and `--no-turboshaft` (heavy page +0.27 MiB) costs 58% on the JS compute benchmark. `--max-semi-space-size=4` and `--max-old-space-size=4096` are V8 150 defaults now, so they are no-ops. The remaining ~0.9 MiB is 3% of a ~28 MiB process, inside the ±10% noise floor in AGENTS.md.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,7 +42,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -76,9 +76,9 @@ checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
 
 [[package]]
 name = "alloc-stdlib"
-version = "0.2.2"
+version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "94fb8275041c72129eb51b7d0322c29b8387a0386127718b096429201a5d6ece"
+checksum = "0e76a019e91224d279006ff972f1e984179a6e9feb050adba6ce8274aef23195"
 dependencies = [
  "alloc-no-stdlib",
 ]
@@ -141,9 +141,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.103"
+version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "arrayref"
@@ -153,15 +153,15 @@ checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
 
 [[package]]
 name = "arrayvec"
-version = "0.7.7"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f02882884d3e1bc524fb12c79f107f6ad0e1cfd498c536ffb494301740995dfe"
+checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
 name = "async-compression"
-version = "0.4.42"
+version = "0.4.43"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e79b3f8a79cccc2898f31920fc69f304859b3bd567490f75ebf51ae1c792a9ac"
+checksum = "3976abdc8fe7d1133d43d304afd42abdf5bc3e1319d263d223bde07b5efc4be8"
 dependencies = [
  "compression-codecs",
  "compression-core",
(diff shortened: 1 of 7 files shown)


SGavrl:
The upgrade is useful, but the current head is not safe to merge. It drops the obstacle course from 33/33 to 32/33 and the Wasm test aborts.

Two changes fixed both locally:

• Use #[op2] instead of #[op2(async(lazy), fast)] for op_fetch_url and op_sleep.
• Call _installWasmStreamingFallback(); during <obscura:init>.

With those changes: release build passes, obstacle course is 33/33, Wasm passes, and a real cdylib links without TPOFF32. Speed is neutral or better, although RSS is still about 0.9 MiB higher and should be brought back to baseline.

gilbertwong96:
Just fixed, please review again. Thanks

gilbertwong96:
Answering the RSS point, plus a status update.

Both fixes you asked for are in: plain `#[op2]` for the fetch/sleep ops, and `_installWasmStreamingFallback()` from the realm init path. The branch is rebased on current `main`, with deno_core now at 0.411 (V8 150.4.0).

On memory I could not get back to baseline without paying for it. Interleaved A/B, same release build and pages, 7 runs each: static/no-JS +0.77 MiB, light JS +0.77, React +0.92, 5000-row DOM build +1.69.

- The JS heap after a full GC is identical (2.7 MB used / 3.8 MB committed), so it is not page data. `vmmap`: V8 anonymous heap pages +384 KiB, native malloc +208 KiB, kernel page tables +208 KiB, `__TEXT` unchanged.
- On JS-heavy pages it is optimized code: `--no-turbofan` removes the delta entirely (−0.25 MiB), `--jitless` leaves +0.11 MiB.
- No speed-neutral knob helps (`--flush-bytecode`, `--max-optimized-bytecode-size`, `--no-concurrent-recompilation`). `--no-turboshaft` gets the heavy page to +0.27 MiB but costs 58% on a JS compute benchmark, so I left it out.
- `--max-semi-space-size=4` / `--max-old-space-size=4096` in `DEFAULT_V8_FLAGS` are V8 150 defaults now, so they are no-ops.

That leaves ~0.9 MiB, 3% of a ~28 MiB process, below the ±10% noise floor in AGENTS.md. If you want baseline regardless, say so and I will price a specific flag as a deliberate trade. Latency is neutral (55 ms vs 54 ms on the same JS benchmark).

One more thing: this PR has been sitting at `action_required`, so no checks have ever run. Could you approve the workflow run?

Obstacle course note: with the current `observer-intersection` fixture the stage fails here and on unmodified `main` alike (the fixture expects the pre-917da17 refire behavior; real Chrome loads 10 items). With the expectation at `io:10` it is 33/33.

PR #456, opened 2026-07-22
Isolate CDP browser contexts

### LINKED ISSUE #449: CDP connections share one BrowserContext: createBrowserContext wipes every client's cookies, and setExtraHTTPHeaders leaks headers across clients
Every CDP connection shares one `BrowserContext`, and three handlers mutate it globally. One client can wipe another's cookies, and one client's `Authorization` header can be sent on another client's navigation.

This is not new — it predates the thread-per-connection work in #435 (both files are byte-identical to `main` on that branch). On `main` a single `cdp_processor` serves the whole server from one `CdpContext`, so the sharing has always been there.

## 1. `createBrowserContext` / `disposeBrowserContext` clear the shared cookie jar

`crates/obscura-cdp/src/domains/target.rs:205-212`:

```rust
"createBrowserContext" => {
    ctx.default_context.cookie_jar.clear();
    Ok(json!({ "browserContextId": ctx.default_context.id }))
}
"disposeBrowserContext" => {
    ctx.default_context.cookie_jar.clear();
```

Playwright and Puppeteer both call `createIncognitoBrowserContext()` / `newContext()` routinely on connect. Client A is logged into a site; client B connects and opens a context; A's session cookies are gone and its next request silently gets a login page. With `--storage-dir`, the shutdown `save_cookies()` then persists the emptied jar, so the wipe outlives the process.

The deeper issue is that these are stubs — `createBrowserContext` returns the *default* context id rather than a new context — and `clear()` looks like an attempt to approximate incognito. Returning an unimplemented error would be safer than destroying state.

## 2. `Network.setExtraHTTPHeaders` / `setUserAgentOverride` leak across connections

`crates/obscura-cdp/src/domains/network.rs:53` and `:61` call `page.http_client.set_extra_headers(...)` / `set_user_agent(...)`, and `page.http_client` is `context.http_client.clone()` (`crates/obscura-browser/src/page.rs:207`) — the shared `Arc<ObscuraHttpClient>`.

Both setters are last-writer-wins process-wide, and `set_extra_headers` is a full replace, not a merge:

```rust
// crates/obscura-net/src/client.rs:725
pub async fn set_extra_headers(&self, headers: HashMap<String, String>) {
    *self.extra_headers.write().await = headers;
}
```

So: client A sets `Authorization: Bearer <A's token>` and navigates. Client B calls `setExtraHTTPHeaders({"X-Foo":"bar"})` — A's auth header vanishes mid-navigation and A gets a 401. Run it the other way and it is a credential leak: B's navigation to an unrelated host carries A's bearer token. Same shape for `setUserAgentOverride`.

All three methods are listed in `is_v8_free_method` (`dispatch.rs:230-231`, `:248-249`), so they take no lock on either the old or the new design — this is not something #435's per-connection lock changes either way.

## Not a memory-safety issue

`CookieJar` is a `std::sync::RwLock` and `ObscuraHttpClient`'s fields are `tokio::sync::RwLock`, so this is a tenant-isolation and data-loss problem, not UB. Worth noting though that every jar access is `.write().unwrap()` / `.read().unwrap()`: a panic in any connection while holding the guard poisons the lock and every other connection then panics on its next cookie access.

## What I would suggest

Per-connection cookie jars and HTTP clients are the real fix, and are a bigger change than either of these handlers. Short of that, making `createBrowserContext` return "not implemented" instead of `clear()` removes the destructive path outright, and scoping the header/UA overrides to the page rather than the shared client removes the leak.

Found while reviewing #435; happy to take any of it if you want to say which direction you prefer.

SGavrl:
Fixes #449.

CDP connections previously shared the same browser context, which allowed cookies, headers, and user-agent changes to leak between clients.

This gives each WebSocket connection its own default context and implements isolated incognito contexts for `Target.createBrowserContext`. Targets, cookies, headers, and user-agent settings now resolve through the correct context.

Storage persistence keeps a separate startup template and merges cookie changes when connections close, so live clients stay isolated without losing persisted state.

### DIFF
--- a/crates/obscura-browser/src/context.rs
+++ b/crates/obscura-browser/src/context.rs
@@ -153,6 +153,46 @@ impl BrowserContext {
         Self::with_options(id, proxy_url, false)
     }
 
+    /// Create a context with the same browser configuration but independent
+    /// mutable network state. Persistent copies start with the template's
+    /// current cookies; incognito copies start empty and never write to the
+    /// template's storage directory.
+    pub fn isolated_copy(&self, id: String, persistent: bool) -> Self {
+        let cookie_jar = Arc::new(CookieJar::new());
+        if persistent {
+            cookie_jar.set_cookies_from_cdp(self.cookie_jar.get_all_cookies());
+        }
+
+        let mut client = ObscuraHttpClient::with_full_options(
+            cookie_jar.clone(),
+            self.proxy_url.as_deref(),
+            self.allow_private_network,
+        );
+        if self.stealth {
+            client.block_trackers = true;
+        }
+        if let Ok(mut guard) = client.user_agent.try_write() {
+            *guard = self.user_agent.clone();
+        }
+
+        BrowserContext {
+            id,
+            cookie_jar,
+            http_client: Arc::new(client),
+            user_agent: self.user_agent.clone(),
+            platform: self.platform.clone(),
+            ua_platform: self.ua_platform.clone(),
+            ua_platform_version: self.ua_platform_version.clone(),
+            proxy_url: self.proxy_url.clone(),
+            robots_cache: Arc::new(RobotsCache::new()),
+            obey_robots: self.obey_robots,
+            stealth: self.stealth,
+            allow_file_access: self.allow_file_access,
+            storage_dir: persistent.then(|| self.storage_dir.clone()).flatten(),
+            allow_private_network: self.allow_private_network,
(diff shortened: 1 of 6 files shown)


Sycatle:
The per-connection default context with the merge-on-close for the jar is cleaner than the not-implemented stub I'd suggested — glad you took it the whole way. Thanks for picking it up.

Issue #467, opened 2026-07-22
createNodeIterator is an alias for createTreeWalker: skips the root node and lacks the NodeIterator interface


mnaza:
### Summary

`document.createNodeIterator` is a bare alias for `createTreeWalker` (`crates/obscura-js/js/bootstrap.js:2707`):

```js
createNodeIterator(root, whatToShow, filter) {
  const iterator = this.createTreeWalker(root, whatToShow, filter);
  iterator._rejectIsSkip = true;
  return iterator;
}
```

`NodeIterator` and `TreeWalker` are different interfaces ([DOM 6.2](https://dom.spec.whatwg.org/#interface-nodeiterator) vs [6.1](https://dom.spec.whatwg.org/#interface-treewalker)), and the alias gets two things wrong.

### 1. The root node is never returned

A `NodeIterator` starts *before* `root`, so the first `nextNode()` returns `root` itself when the filter accepts it. A `TreeWalker` starts *at* `root` and never returns it.

```js
document.body.innerHTML = '<div id="root"><a></a></div>';
const it = document.createNodeIterator(document.getElementById('root'), NodeFilter.SHOW_ELEMENT);
const seen = [];
let n;
while ((n = it.nextNode())) seen.push(n.tagName);
console.log(seen);
```

| | result |
|---|---|
| Chrome | `["DIV","A"]` |
| obscura | `["A"]` |

Silently dropping the root is the worst shape this can take: a `NodeIterator` rooted at the element you care about skips exactly that element, so `createNodeIterator(el, SHOW_ELEMENT)` never yields `el`.

### 2. Missing interface surface

`referenceNode`, `pointerBeforeReferenceNode` and `detach()` are all absent — the alias exposes `TreeWalker`'s `currentNode`, `firstChild()`, `nextSibling()`, `parentNode()` etc. instead, none of which exist on a real `NodeIterator`.

```js
typeof it.referenceNode                 // "undefined"  (Chrome: "object")
typeof it.pointerBeforeReferenceNode    // "undefined"  (Chrome: "boolean")
typeof it.detach                        // "undefined"  (Chrome: "function")
typeof it.firstChild                    // "function"   (Chrome: "undefined")
```

`detach()` in particular is a no-op in modern DOM but is still called by older library code, and throws here.

### Expected

[...]

Issue #461, opened 2026-07-22
TreeWalker ignores NodeFilter.FILTER_REJECT and descends into rejected subtrees


mnaza:
### Summary

`TreeWalker`/`NodeIterator` treat `NodeFilter.FILTER_REJECT` as if it were `FILTER_SKIP`: the node is not returned, but traversal still descends into its subtree. Per [DOM §6.1](https://dom.spec.whatwg.org/#interface-treewalker), `FILTER_REJECT` in a `TreeWalker` must reject the node **and its entire subtree**; only `FILTER_SKIP` leaves descendants eligible.

The cause is in `_accept` (`crates/obscura-js/js/bootstrap.js:2619`), which collapses every non-`FILTER_ACCEPT` return into a single boolean:

```js
if (typeof this.filter === 'function') return this.filter(node) === 1;
if (this.filter.acceptNode) return this.filter.acceptNode(node) === 1;
```

`nextNode()` then treats `false` uniformly as "skip, keep descending".

### Reproduction

```js
document.body.innerHTML = '<div id="root"><section><p>deep</p></section><a></a></div>';
const root = document.getElementById('root');
const w = document.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, {
  acceptNode(n) {
    return n.tagName === 'SECTION' ? NodeFilter.FILTER_REJECT : NodeFilter.FILTER_ACCEPT;
  }
});
const seen = [];
let n;
while ((n = w.nextNode())) seen.push(n.tagName);
console.log(seen);
```

| | result |
|---|---|
| Chrome | `["A"]` |
| obscura | `["P","A"]` |

`<p>` is inside the rejected `<section>` subtree and must not be visited.

### Why this matters now

This was unreachable before #439/#440: `NodeFilter.FILTER_REJECT` evaluated to `undefined`, so no page could return it. Now that the full constant set is exposed and #432/#447 made `nextNode()` actually traverse the whole subtree, the canonical prune idiom

```js
if (node.tagName === 'SCRIPT' || node.tagName === 'STYLE') return NodeFilter.FILTER_REJECT;
```

silently yields nodes from inside the subtree the page asked to prune. Sanitizers, text extractors and content-scrapers built on `TreeWalker` are the common users of this idiom, and they now get extra content rather than an error.

### Expected

[...]

PR #471, opened 2026-07-22
fix(js): descend on FILTER_SKIP in TreeWalker child/sibling movers (#469)

### LINKED ISSUE #469: TreeWalker firstChild/lastChild/nextSibling/previousSibling do not descend on FILTER_SKIP
### Summary

`TreeWalker`'s child and sibling movers — `firstChild()`, `lastChild()`, `nextSibling()`, `previousSibling()` — treat a filtered-out node as "move to the next sibling", when [DOM 6.1](https://dom.spec.whatwg.org/#interface-treewalker) says `FILTER_SKIP` leaves the skipped node's **children** eligible and the walk must descend into them.

`crates/obscura-js/js/bootstrap.js:2668` and following, all four the same shape:

```js
firstChild() {
  let child = this.currentNode.firstChild;
  while (child) {
    if (this._accept(child)) { this.currentNode = child; return child; }
    child = child.nextSibling;     // <- never descends
  }
  return null;
}
```

### Reproduction

```js
document.body.innerHTML = '<div id="root"><section><a></a></section></div>';
const root = document.getElementById('root');
const w = document.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, {
  acceptNode(n) { return n.tagName === 'SECTION' ? NodeFilter.FILTER_SKIP : NodeFilter.FILTER_ACCEPT; }
});
console.log(w.firstChild());
```

| | result |
|---|---|
| Chrome | `<a>` |
| obscura | `null` |

`<section>` is skipped, so its child `<a>` becomes the first eligible child. The same defect applies to `lastChild()` (must descend to last children), and to `nextSibling()` / `previousSibling()`, which must descend into a skipped sibling's subtree before moving on.

This also means the movers ignore the `FILTER_ACCEPT` / `FILTER_REJECT` / `FILTER_SKIP` distinction entirely: they use the boolean `_accept`, so `FILTER_REJECT` (which must *not* descend) and `FILTER_SKIP` (which must) are indistinguishable — the same root cause as #461, which fixed it for `nextNode()` only.

### Expected

Implement the spec's "traverse children" and "traverse siblings" algorithms:

- `firstChild()` / `lastChild()`: on `FILTER_SKIP`, descend into the node's first/last child and re-filter; on `FILTER_REJECT`, do not descend; on exhaustion, climb back without passing `root` or `currentNode`.
- `nextSibling()` / `previousSibling()`: on a non-accepted sibling, descend into its first/last child unless it was rejected, then continue; stop when reaching `root` or an accepted parent.

This builds on the three-valued `_filter` added in #461.

mnaza:
Fixes #469.

> **Rebased onto current `main`**, which now includes #464 (issue #461, merged) — the three-valued `_filter` it needs is upstream, so this branch carries only its own change. Independent of the other open DOM PRs, and it now merges cleanly with **#472** (issue #467): the earlier test-file conflict between the two is gone now that both sit on the shared base. Verified with a local test-merge.

`firstChild()`, `lastChild()`, `nextSibling()` and `previousSibling()` stepped straight to the next sibling whenever a node was not accepted, so a `FILTER_SKIP` node **hid its children** instead of exposing them:

```js
// <div id="root"><section><a></a></section></div>, SECTION -> FILTER_SKIP
w.firstChild()   // null      Chrome: <a>
```

They also used the boolean `_accept`, so `FILTER_REJECT` (which must not descend) and `FILTER_SKIP` (which must) were indistinguishable — the same root cause #461 fixed for `nextNode()`, left in place for all four movers.

### Approach

Replaced all four with the spec's [traverse children](https://dom.spec.whatwg.org/#concept-traverse-children) and [traverse siblings](https://dom.spec.whatwg.org/#concept-traverse-siblings) algorithms, parameterised by direction so forward and backward share one implementation and cannot drift: `edge` is the first/last child and `step` is the next/previous sibling.

Both consult the three-valued `_filter` from #461, so `SKIP` descends, `REJECT` prunes, and the climb back out stops at `root` or at the node the walk started from.

### Scope note

`parentNode()` is deliberately **not** touched here. I noticed while implementing that it has its own deviations — it checks only the immediate parent instead of climbing, and excludes `root`, which [DOM 6.1 step 2](https://dom.spec.whatwg.org/#dom-treewalker-parentnode) explicitly permits it to return. Those are different in kind from the `FILTER_SKIP` descent bug and belong in their own change rather than being folded in silently. Happy to file it separately.

### Tests

- `tree_walker_child_movers_descend_on_skip_but_not_on_reject` — `firstChild`/`lastChild` under both verdicts, in one table.
- `tree_walker_sibling_movers_descend_into_skipped_siblings` — `nextSibling` descends into a skipped sibling, and steps past a rejected one.
- `tree_walker_previous_sibling_descends_to_last_child` — the backward mover descends to *last* children, which is where a direction-agnostic implementation would silently be wrong.

All verified RED first.
### Verification

Rebased onto current `main` (post-#473, whose test-fixture fix moved `run_page_init()` into `setup_runtime` — that is what cleared the failures the original body called "pre-existing", so the suite is now fully green). Full `obscura-js` suite after rebase: **107 passed / 0 failed**, the 3 added tests included.


SGavrl:
Could you rebase please. Thanks.

PR #513, opened 2026-07-26
fix(mcp): bound the request read with a timeout (slowloris)

### LINKED ISSUE #501: MCP HTTP server has no read timeout (slowloris DoS)
### Summary
The MCP HTTP server has no read timeout, so a slow or stalled client holds the connection (and, because connections are served sequentially, blocks all other MCP clients). The request-line loop, header loop, and `reader.read_exact(&mut body)` all `.await` with no time bound; a client that sends `Content-Length: 16000000` then stalls, or dribbles headers one byte at a time, wedges the server. `MAX_BODY_BYTES` caps allocation but not the stall (slowloris).

### Fix
Wrap the request read (request line, headers, and body) in a bounded `tokio::time::timeout`; on expiry, close the connection (optionally 408) instead of waiting forever. Keep it generous enough not to affect legitimate slow networks.

mnaza:
Fixes #501.

## Problem
The MCP HTTP server's request-line, header, and body reads all awaited with no time bound, so a client that connected and then sent its request slowly (or stalled) held the connection forever; because connections are served sequentially, that blocked every other MCP client (slowloris).

## Fix
Bound the whole request read with a single per-request deadline (`OBSCURA_MCP_READ_TIMEOUT_MS`, default 30s, floored at 50ms); on expiry send 408 and close.

## Tests
New integration test `read_timeout.rs`: a client that stalls mid-request is disconnected within the (shortened) timeout; RED before, GREEN after. obscura-mcp 6/6.

### DIFF
--- a/crates/obscura-mcp/src/http.rs
+++ b/crates/obscura-mcp/src/http.rs
@@ -12,6 +12,20 @@ use crate::{dispatch, BrowserState};
 /// is far above any real JSON-RPC tool call.
 const MAX_BODY_BYTES: usize = 16 * 1024 * 1024;
 
+/// Per-request read deadline. Without one, a client that opens a connection and
+/// then sends its request line/headers/body slowly (or not at all) holds the
+/// connection open forever; because connections are served sequentially, that
+/// stalls every other MCP client (slowloris). Configurable via
+/// `OBSCURA_MCP_READ_TIMEOUT_MS` (default 30s, floored at 50ms).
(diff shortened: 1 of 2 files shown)


mnaza:
Closing as redundant: issue #501 was already fixed upstream by #505 (commit 750ffa6, `fix(mcp): bound HTTP request read time`), which bounds the whole request read with a shared 30s deadline — the same slowloris protection this PR adds. No further change needed.

PR #564, opened 2026-08-04
fix(net): decode Content-Encoding on the stealth client

### LINKED ISSUE #563: --stealth skips Content-Encoding decoding, raw gzip bytes reach the HTML parser
## Summary

With `--stealth`, a response carrying `Content-Encoding: gzip` reaches the HTML parser **undecompressed**. The parser treats the gzip bytes as document text, producing a DOM of `<html><head></head><body>«binary»`. Without `--stealth` the identical response decodes correctly. There is no error, warning, or non-zero exit — the caller receives plausible-looking HTML.

Reproduced on `main` at `f0541f40a487` and on the released 0.1.10 binary. Linux x86_64. Reproduces over plain HTTP on loopback, so TLS/h2 is not involved.

## Minimal Python script to reproduce

```python
import socket, threading, zlib, subprocess, sys

BIN = sys.argv[1] if len(sys.argv) > 1 else "obscura"
PAGE = (b'<!DOCTYPE html><html lang="de-DE"><head><title>hello</title></head>'
        b'<body><p id="MARK">plain text</p></body></html>')
c = zlib.compressobj(9, zlib.DEFLATED, 31)
GZ = c.compress(PAGE) + c.flush()

def handle(conn):
    conn.recv(65536)
    conn.sendall(("HTTP/1.1 200 OK\r\nContent-Type: text/html; charset=utf-8\r\n"
                  "Content-Encoding: gzip\r\nContent-Length: %d\r\n\r\n" % len(GZ)).encode() + GZ)
    conn.close()

srv = socket.socket(); srv.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
srv.bind(("127.0.0.1", 8987)); srv.listen(8)
threading.Thread(target=lambda: [threading.Thread(target=handle, args=(srv.accept()[0],),
                 daemon=True).start() for _ in iter(int, 1)], daemon=True).start()

for flags in ([], ["--stealth"]):
    out = subprocess.run([BIN, "--allow-private-network"] + flags +
                         ["fetch", "--dump", "html", "http://127.0.0.1:8987/"],
                         capture_output=True, timeout=90).stdout
    print("  %-11s -> %s" % (flags or ["(no flags)"],
          "OK, decoded" if b'id="MARK"' in out else "BROKEN: %r" % out[:76]))
```

**Expected:** both invocations print `OK, decoded`.

**Actual:**

```
['(no flags)'] -> OK, decoded
['--stealth']  -> BROKEN: b'<!DOCTYPE html>\n<html><head></head><body>\x1f\xef\xbf\xbd\x08\x02\x03\xef\xbf\xbdQt\xef\xbf\xbdw\x0e...'
```

`1f (8b) 08 … 02 03` is the gzip header of the served body. `8b` and every other invalid byte became U+FFFD; the header's five NUL bytes are gone because HTML5 tree construction ignores U+0000 character tokens in the "in body" insertion mode; `<head></head><body>` is auto-inserted because the first token after `<html>` is `0x1f`.

## Root cause

`stealth` swaps the HTTP client (`crates/obscura-net/Cargo.toml:8`, `stealth = ["wreq", "wreq-util"]`). wreq gates each decoder behind a cargo feature (`gzip`/`brotli`/`deflate`/`zstd`, each pulling a `tower-http` decompression layer), exactly like reqwest. The reqwest path enables them; the wreq path does not:

| declaration | features |
|---|---|
| `Cargo.toml:39` reqwest (default path) | `gzip, brotli, deflate, rustls-tls, socks` |
| `crates/obscura-net/Cargo.toml:42` wreq (stealth, Linux) | `prefix-symbols, socks` |
| `crates/obscura-net/Cargo.toml:45` wreq (stealth, other) | `socks` |

So `StealthHttpClient::with_proxy` (`crates/obscura-net/src/wreq_client.rs:49`) builds a client with no decoder registered, and `resp.bytes()` — `:174` for navigation, `:254` in `send_single` for scripted `fetch()`/XHR — stores the still-compressed body straight into `Response.body`. Meanwhile `wreq_util::Emulation` with `Profile::Chrome145` sends Chrome's real `accept-encoding`, so origins do compress.

Nothing in `crates/obscura-net/src` references gzip or content-encoding: the default path relies entirely on reqwest's implicit feature-driven behaviour, so the stealth port silently lost it. This looks like the same class as #538/#551 (cert store) and #482 (TLS on navigation) — a capability the reqwest path has that the stealth client quietly lacks.

## Fix (built and verified)

```diff
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
-wreq = { version = "=6.0.0-rc.29", optional = true, features = ["prefix-symbols", "socks"] }
+wreq = { version = "=6.0.0-rc.29", optional = true, features = ["prefix-symbols", "socks", "gzip", "brotli", "deflate", "zstd"] }

 [target.'cfg(not(any(target_os = "linux", target_os = "android")))'.dependencies]
-wreq = { version = "=6.0.0-rc.29", optional = true, features = ["socks"] }
+wreq = { version = "=6.0.0-rc.29", optional = true, features = ["socks", "gzip", "brotli", "deflate", "zstd"] }
```

After a 20 s incremental rebuild the test case prints `OK, decoded` for both invocations. `zstd` is included deliberately — the Chrome145 emulation advertises it, so without it zstd responses fail the same way.

## Scope

Each tested with `--stealth` as the only variable:

- **Not size-dependent** — fails at gzip bodies of 9.5 KB, 35 KB, 70 KB, 81 KB, 229 KB; all pass without `--stealth`.
- **Not framing-dependent** — `Content-Length`, `Transfer-Encoding: chunked`, multi-chunk and close-delimited all fail under `--stealth`.
- **Scripted requests too** — `send_single` has the same `resp.bytes()` pattern, so JS `fetch()`/XHR over the stealth client are equally affected.
- **Not TLS/h2** — reproduces over plain HTTP/1.1 on loopback.
- **`--dump original` unaffected** — it bypasses the browser layer.

## Impact

Live `https://www.lidl.de/` (serves `content-encoding: gzip`), same binary, same UA:

| | before fix | after fix |
|---|---|---|
| `--dump html --stealth` output | 515,814 B | 868,357 B |
| U+FFFD in output | 26,779 | 0 |
| `<html lang>` | `en` (latches onto a nested inlined document) | `de-DE` |
| `<title>` | `search` | `Lidl lohnt sich …` |

Over HTTPS the leak is partial rather than total: the DOM keeps most of the document but loses ~40% of it and gains 26,779 replacement characters, with a raw-gzip text node as `document.body.firstChild`. Verified in the live DOM via `--eval`:

```json
{"parentTag":"BODY","isFirstOfBody":true,"len":46,
 "head":["1f","fffd","8","2","3","fffd","51","74","fffd","77"],
 "nextSibling":"#comment","nextData":" empty_kam_esi "}
```

ibukanov:
Fixes #563.

`crates/obscura-net/Cargo.toml` declared the `wreq` dependency without the `gzip`, `brotli`, `deflate` and `zstd` features, so the stealth client had no decoder registered, while the reqwest path (`Cargo.toml:39`) enables all of them. `wreq_client.rs` stores `resp.bytes()` straight into `Response.body`, so a compressed response reached the HTML parser as document text and the DOM came out as `<html><head></head><body>«raw gzip»`. Silently: no error, no warning, no non-zero exit.

The Chrome145 emulation advertises gzip, br, deflate and zstd, so origins do compress. `zstd` is included for that reason; without it zstd responses fail the same way.

## Verification

- New regression test `stealth_client_decodes_gzip_response` in `wreq_client.rs` serves one `Content-Encoding: gzip` response on an ephemeral port and asserts the body is decompressed. It fails on `main` (the assert sees the raw gzip bytes) and passes with this change. The gzip fixture is a hardcoded const, so the test adds no dependency.
- `cargo build --release --features stealth` compiles clean.
- `cargo test -p obscura-net --features stealth` passes.
- Obstacle course against the patched release binary: **33/33**, median latency 12.0ms.
- Live check on a gzip origin (`https://www.lidl.de/`, `--dump html --stealth`): 515,814 bytes with 26,779 U+FFFD before, 868,357 bytes with 0 U+FFFD after, and `<html lang>` plus `<title>` correct again.

Dependency features and one test only, so no engine code path changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -496,6 +496,8 @@ dependencies = [
  "compression-core",
  "flate2",
  "memchr",
+ "zstd",
+ "zstd-safe",
 ]
 
 [[package]]
@@ -3680,6 +3682,7 @@ dependencies = [
  "tokio-btls",
  "tokio-socks",
  "tower",
+ "tower-http",
  "url",
  "webpki-root-certs",
  "wreq-proto",
--- a/crates/obscura-net/Cargo.toml
+++ b/crates/obscura-net/Cargo.toml
@@ -39,7 +39,7 @@ tokio-rustls = { version = "0.26", default-features = false, features = ["ring",
 # Enabling it on other platforms produces unresolved `build_script_main_*`
 # symbols at link time (see issue #39).
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
-wreq = { version = "=6.0.0-rc.29", optional = true, features = ["prefix-symbols", "socks"] }
+wreq = { version = "=6.0.0-rc.29", optional = true, features = ["prefix-symbols", "socks", "gzip", "brotli", "deflate", "zstd"] }
 
 [target.'cfg(not(any(target_os = "linux", target_os = "android")))'.dependencies]
-wreq = { version = "=6.0.0-rc.29", optional = true, features = ["socks"] }
+wreq = { version = "=6.0.0-rc.29", optional = true, features = ["socks", "gzip", "brotli", "deflate", "zstd"] }
--- a/crates/obscura-net/src/wreq_client.rs
+++ b/crates/obscura-net/src/wreq_client.rs
@@ -278,3 +278,65 @@ impl StealthHttpClient {
         self.active_requests() == 0
     }
 }
+
+#[cfg(all(test, feature = "stealth"))]
+mod tests {
+    use std::sync::Arc;
+
+    use tokio::io::{AsyncReadExt, AsyncWriteExt};
+    use url::Url;
+
+    use super::StealthHttpClient;
+    use crate::cookies::CookieJar;
+
+    const PLAIN_BODY: &str = "<!DOCTYPE html><html><body><p id=\"mark\">gzip ok</p></body></html>";
+
+    // gzip (level 9) of PLAIN_BODY, hardcoded so the fixture needs no
+    // compression dependency. A wrong byte fails the assert below.
+    const GZIP_BODY: &[u8] = &[
+        0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x03, 0xb3, 0x51,
+        0x74, 0xf1, 0x77, 0x0e, 0x89, 0x0c, 0x70, 0x55, 0xc8, 0x28, 0xc9, 0xcd,
+        0xb1, 0xb3, 0x81, 0x90, 0x49, 0xf9, 0x29, 0x95, 0x76, 0x36, 0x05, 0x0a,
+        0x99, 0x29, 0xb6, 0x4a, 0xb9, 0x89, 0x45, 0xd9, 0x4a, 0x76, 0xe9, 0x55,
+        0x99, 0x05, 0x0a, 0xf9, 0xd9, 0x36, 0xfa, 0x05, 0x76, 0x36, 0xfa, 0x10,
+        0x69, 0x7d, 0xb0, 0x5a, 0x00, 0x80, 0x3d, 0x1c, 0x5f, 0x41, 0x00, 0x00,
+        0x00,
+    ];
+
+    /// Serve one `Content-Encoding: gzip` response on an ephemeral port.
+    async fn gzip_fixture() -> u16 {
+        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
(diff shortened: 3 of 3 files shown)


dannolan:
I'm very keen to see this one land, my agent reimplemented it this morning!

Issue #570, opened 2026-08-07
serve mode: Target.getTargets empty on fresh browser connection — puppeteer.connect sees no pages


ZhongJiangcheng:
Environment: obscura v0.1.11, Windows x64.

Reproduction:
1. Start: obscura serve --port 9222
2. Connect a fresh WebSocket to ws://127.0.0.1:9222/devtools/browser (exactly what puppeteer.connect does)
3. Send Target.getTargets -> {"targetInfos":[]} — even though GET /json lists page-1 (about:blank)
4. Target.setDiscoverTargets also emits no targetCreated for the existing page
5. Only after calling Target.createTarget (or Target.attachToTarget) within that session does the target appear (attached:true)

Consequence: puppeteer.connect() finds zero pages -> tools fail with 'No page targets available on the attached browser'. The HTTP /json endpoint and the CDP target manager disagree about existing pages.

Expected: existing pages visible via Target.getTargets on a fresh session (Chrome behavior), or targetCreated events for them after setDiscoverTargets.

CLI fetch mode works fine — only serve/CDP mode is affected.

yinnho:
We don't have this disagreement on our bridge — but not because our `getTargets` is smarter. It's because there is no persistent browser for the two surfaces to disagree about, and that difference seems worth spelling out for whoever picks the fix shape here.

Our CDP bridge is per-connection isolated: every `/devtools/*` WebSocket upgrade gets a fresh context (own page set, cookie jar, HTTP client — our `Page`s are `!Send` deno_core state pinned to one thread, so isolation is also the cheap design). Measured live on it, following your repro steps:

- `GET /json/list` → `[]`
- fresh WebSocket to `/devtools/browser/<any-id>`, `Target.getTargets` → `{"targetInfos":[]}`
- `Target.setDiscoverTargets {discover:true}` → ack plus exactly one `Target.targetCreated` (the `browser` target)
- `Target.createTarget` → the page appears in `getTargets` and as a `targetCreated` event at the same time

All three surfaces read the same answer because they read the same per-connection state. Puppeteer `connect()` and Playwright `connectOverCDP()` both drive it — they get zero existing pages and create what they need.

[...]

Issue #543, opened 2026-08-03
Target.getTargets returns empty for new connections — pages not globally visible (CDP spec violation)


javabcde:
## Root cause found — regression from #456

Found the cause while reading the source. `Target.getTargets` itself is correct (iterates `ctx.pages`), but **`CdpContext.pages` is per-connection**. PR #456 (merged 2026-07-22) deliberately gave each WebSocket connection its own default browser context for cookie/UA isolation — and targets live in that per-connection context, so a new connection enumerates only its own (empty) pages.

That also explains why this wasn't reported earlier: it's a regression introduced ~1 week before the 0.1.11 build we tested.

## Suggested fix (Chrome semantics)

- Keep per-context data isolation (cookies/headers/UA) as #456 intended.
- Move the **target registry** to server-level shared state so `getTargets`/setDiscoverTargets see all live pages across connections, each tagged with its `browserContextId` (the code already carries `page.context.id`).
- Route page ops by sessionId → the owning context (already how the dispatcher routes by `{targetId}-session`).
- Serialize access through the existing global `v8_lock`.

This matches Chromium: targets are globally visible, but each belongs to a browser context; a client attaches and operates via sessionId.

javabcde:
## Important follow-up: setAutoAttach is a deeper blocker than getTargets

While investigating I found that even with getTargets fully working (v0.1.10, before #456), a standard Puppeteer connect() + browser.pages() client (omp's browser tool) still cannot discover pages.

```rust
// crates/obscura-cdp/src/domains/target.rs
"setAutoAttach" => Ok(json!({})),
// No multi-target lifecycle to manage: obscura runs one page per session.
// Ack these so Chrome-shaped clients that call them do not warn (issue #340).
```

setAutoAttach is an intentional no-op (one page per session). Puppeteer's connect flow: (1) setDiscoverTargets -> targetCreated events with attached:false; (2) relies on setAutoAttach to receive Target.attachedToTarget events that build browser.pages(); (3) Obscura acks with {} and never emits attachedToTarget -> pages() stays empty -> enumeration clients fail with 'No page targets available'. The Puppeteer quickstart works only because newPage() -> Target.createTarget creates the session inline, bypassing auto-attach.

[...]

PR #595, opened 2026-08-09
fix(cdp): Hermes-driver compatibility (getTargets, getBoxModel ints, insertText)

fepfitra:
Fixes three CDP quirks that break Hermes Agent (and other puppeteer/playwright-style clients) driving obscura:

1. **Target.getTargets returns empty on fresh browser connections** (#543, #570). A browser-level client that connects and immediately calls Target.getTargets saw `[]` because the pages registry was only populated by session-scoped events. Now `NewConnection` creates a page + session mirroring the interception path, so targets are globally visible per the CDP spec.

2. **DOM.getBoxModel emits float coordinates** where the spec requires integers. Clients deserializing coordinates as i64 (Hermes: "invalid type: floating point 32.0, expected i64") fail. All quad coords + width/height now rounded to integers.

3. **Input.insertText not implemented** ("Unknown Input method: insertText"). Implemented via the existing prototype-setter JS path (`__obscura_setFieldValue`), caret-aware with selection replacement, and dispatches a trusted `input` event so React/Vue controlled inputs register typed text.

All three verified live driving the patched build from Hermes: fresh-connection getTargets lists targets, getBoxModel returns integer coords, insertText types into a login form and submits.

Tests added: `get_box_model_returns_integer_coordinates`, `insert_text_types_into_focused_input`, `fresh_connection_sees_page_targets_in_get_targets`. Full workspace: 81 tests pass, release build clean.

### DIFF
--- a/crates/obscura-cdp/src/domains/dom.rs
+++ b/crates/obscura-cdp/src/domains/dom.rs
@@ -87,6 +87,15 @@ fn guess_mime(path: &str) -> &'static str {
     }
 }
 
+// The CDP BoxModel protocol types its coordinates as integers (Quad is an
+// array of 8 integer pairs; x/y/width/height are integers too).
+// getBoundingClientRect returns CSS floats, and Hermes refuses to deserialize
+// them ("invalid type: floating point 32.0, expected i64"), so round every
+// emitted coordinate to the nearest whole pixel.
+fn round_px(v: f64) -> i64 {
+    v.round() as i64
+}
+
 pub async fn handle(
     method: &str,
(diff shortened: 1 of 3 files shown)


fepfitra:
First of a 3-PR stack (595 → 597 → 634). All mergeable/clean. Fixes Hermes browser integration (getTargets, getBoxModel ints, insertText).

fepfitra:
Superseded by #634 — consolidated the 3-PR stack into one PR (single reviewable diff, no merge ordering).

PR #597, opened 2026-08-10
fix(cdp): global target registry — targets visible & attachable across connections (#570)

fepfitra:
fix(cdp): global target registry — targets visible & attachable across connections (#570)

Per-connection target lookup made Target.getTargets list only pages owned
by the caller, so a page created on one connection was invisible to others
and attachToTarget failed with "Target not found" (Hermes repro). Pages are
thread-confined (thread-per-connection #430) so their V8 isolates cannot be
shared; instead mirror their metadata in a process-wide registry.

- registry.rs: TargetRegistry — global page-id counter, upsert/remove/all
- createPage/createTarget register the page; Drop unregisters a dead
  connection's pages; navigation syncs url/title into the registry
- getTargets returns every live target with attached=false for remote ones
- attachToTarget/closeTarget resolve through the registry (browser-global,
  matching Chrome), sessions stay connection-local
- /json/list mirrors the registry instead of a hardcoded about:blank page

### DIFF
--- a/crates/obscura-cdp/src/dispatch.rs
+++ b/crates/obscura-cdp/src/dispatch.rs
@@ -7,6 +7,7 @@ use serde_json::json;
 
 use crate::domains;
 use crate::domains::fetch::FetchInterceptState;
+use crate::registry::{TargetInfo, TargetRegistry};
 use crate::types::{CdpEvent, CdpRequest, CdpResponse};
 
 #[cfg(feature = "render")]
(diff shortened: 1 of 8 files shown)


fepfitra:
Stacked on #595. Merge order: 595 → 597 → 634. All three mergeable/clean. Verified live with Hermes browser tools.

fepfitra:
Superseded by #634 — consolidated the 3-PR stack into one PR (single reviewable diff, no merge ordering).

PR #623, opened 2026-08-10
feat(cdp): report the page's real child frame tree

### LINKED ISSUE #600: Child iframes never get a scripting context: <script> in a child frame never executes, Page.getFrameTree reports no childFrames
### Before you start

- [x] I tested against the latest release **and** `main`.

**obscura version or commit:** `main` @ `458d068` (source build); also reproduces on the official **v0.2.0** release binary
**OS and architecture:** macOS 26.6.1 arm64 (Apple Silicon)
**How are you using obscura?** CLI (`fetch`) — also reproduces via `serve` + CDP (Playwright `connect_over_cdp` and a raw CDP client)
**Affected area:** JavaScript or DOM behavior
**Build variant:** Render-enabled with stealth (`--features render,stealth`)
**Stealth mode:** reproduces both with and without `--stealth`
**Does headless Chrome behave the same?** **No** — headless Chrome behaves correctly and obscura does not. The repro below runs entirely on loopback, so no anti-bot service is involved.

---

## Summary

A child iframe's document is fetched and parsed — `contentDocument` is genuinely populated and its body content is readable — but the frame never receives a scripting context. `<script>` elements inside it sit in the DOM as inert nodes and never execute, so the child can never call `parent.postMessage(...)` either.

`Page.getFrameTree` confirms this at the protocol level: `childFrames` is always `[]`, even while the parent's `document.querySelectorAll('iframe')` reports the frame and `iframe.contentDocument.URL` is correct. The child is a parsed document, not a browsing context.

**This is not #592 / #593.** That pair fixed the globals exposed on the `contentWindow` *object* (so `new iframe.contentWindow.XMLHttpRequest()` works), and I verified the fix is present and effective on `458d068`: globals visible in an iframe realm went from 5/14 on v0.2.0 to 14/14 on `main`. Scripts *inside* the child document still do not run, and the Turnstile symptom below is byte-identical before and after.

## Reproduction

Two files and a static server. No external network, no anti-bot service.

`child.html`:

```html
<!doctype html><html><head><title>BEFORE</title></head><body>
<p>child</p>
<script>
  window.__ran = "YES";
  document.title = "RAN-IN-CHILD";
  try { parent.postMessage("FROM-CHILD", "*"); }
  catch (e) { document.title = "POST-THREW:" + e.message; }
</script>
</body></html>
```

`parent.html`:

```html
<!doctype html><html><head><title>parent</title></head><body>
<script>
  window.__res = {parentGot: [], childRan: null, childTitle: null, iframeLoaded: false};
  window.addEventListener('message', function (e) { window.__res.parentGot.push(String(e.data)); });
  var f = document.createElement('iframe');
  f.onload = function () {
    window.__res.iframeLoaded = true;
    try { window.__res.childTitle = f.contentDocument ? f.contentDocument.title : 'no-doc'; }
    catch (err) { window.__res.childTitle = 'blocked:' + err.name; }
    try { window.__res.childRan = f.contentWindow ? String(f.contentWindow.__ran) : 'no-win'; }
    catch (err) { window.__res.childRan = 'blocked:' + err.name; }
  };
  f.src = '/child.html';
  document.body.appendChild(f);
</script>
</body></html>
```

```bash
python3 -m http.server 8899 --bind 127.0.0.1 &

obscura --allow-private-network fetch http://127.0.0.1:8899/parent.html --wait 3 -q \
  --eval '(function(){var f=document.querySelector("iframe");return JSON.stringify({
    childRan: (function(){try{return String(f.contentWindow.__ran)}catch(e){return "blocked"}})(),
    childTitle: (function(){try{return f.contentDocument.title}catch(e){return "blocked"}})(),
    childBodyLen: (function(){try{return f.contentDocument.body.innerHTML.length}catch(e){return -1}})(),
    parentGot: window.__res ? window.__res.parentGot : "no-res"});})()'
```

## Expected behavior

The child frame runs its own script, as headless Chrome does on the same two files:

```json
{"childRan": "YES", "childTitle": "RAN-IN-CHILD", "parentGot": ["FROM-CHILD"], "iframeLoaded": true}
```

```
frames: 2  ['http://127.0.0.1:8899/parent.html', 'http://127.0.0.1:8899/child.html']
```

## Actual behavior

```json
{"childRan":"undefined","childTitle":"","childBodyLen":201,"parentGot":[]}
```

Side by side, same two files, same server:

| Observation | headless Chrome 145 | obscura `458d068` / v0.2.0 |
|---|---|---|
| `iframe.onload` fires | yes | yes |
| child `contentDocument` populated | yes | **yes** (201 bytes of body HTML) |
| child `<script>` executes (`window.__ran`) | `"YES"` | **`undefined`** |
| child `document.title` after script | `"RAN-IN-CHILD"` | **`""`** |
| child → parent `postMessage` | `["FROM-CHILD"]` | **`[]`** |
| frames / `Page.getFrameTree.childFrames` | 2 frames | **1 frame, `childFrames: []`** |

Note that the child's static `<title>BEFORE</title>` is missing as well, so even the child document's `<head>` is not reflected in `contentDocument.title`, while its `body` content is present.

## Logs

`Page.getFrameTree` taken while the iframe is loaded and its `contentDocument.URL` is correctly `http://127.0.0.1:8899/child.html`:

```json
{
  "frameTree": {
    "frame": {
      "id": "page-1",
      "loaderId": "initial-loader",
      "url": "http://127.0.0.1:8899/parent.html",
      "securityOrigin": "http://127.0.0.1:8899/parent.html",
      "mimeType": "text/html"
    },
    "childFrames": []
  }
}
```

## Why this matters

Any widget that runs its logic in an embedded frame and reports back over `postMessage` cannot work: OAuth and consent flows, payment fields, embedded players, and anti-bot widgets.

Concretely, this is why Cloudflare Turnstile cannot complete. Turnstile's `api.js` does:

```js
var le = document.createElement("iframe"),
    dt = document.createElement("div"),
    ce = dt.attachShadow({mode: "closed"});
```

so the challenge runs in a cross-origin iframe inside a closed shadow root and reports its result outward via `postMessage`. On obscura, `turnstile.render()` logs `render_success`, then nothing happens and api.js times itself out after ~43 s:

```
[ 0.3s] [Turnstile] render_start   {"sitekey":"0x4AAAAAABFlBOxeiHIxst8q"}
[ 0.3s] [Turnstile] render_success {"widgetId":"cf-chl-widget-87zuk"}
[43.6s] Turnstile Widget seem to have hung:  87zuk
[43.6s] [Turnstile] error: {"errorCode":"300030","errorTitle":"Security Check Failed","timeTakenMs":43311}
[43.6s] [Cloudflare Turnstile] Error: 300030.
```

Reproduced on Cloudflare's own diagnostic page `https://browser-compat.turnstile.workers.dev/`, identically on v0.2.0 and on `458d068`. That page's three local checks — `navigator.webdriver`, clock skew, and reachability of `challenges.cloudflare.com` — all **pass**, and Cloudflare returns a `__cf_bm` cookie, so the TLS/fingerprint layer is accepted. The failure is the frame, not detection.

To be explicit about scope: I am reporting the loopback repro above as the engine bug. The Turnstile case is only the real-world symptom that led me to it, and it needs no third-party service to reproduce.

## Related observations

Same investigation, likely the same root cause — happy to split these into separate issues if you prefer:

1. **`<iframe srcdoc>` never fires `onload`** (times out at 6 s), while `src`-based iframes do fire it.
2. **`data:text/html,…` iframes** also yield an empty `contentDocument.title`, and their inline script does not run.
3. **Scripts appended into a child document from the parent stay inert**: `child.contentDocument.createElement('script')` + `textContent` + `appendChild` never executes.
4. **`window.postMessage('x','*')` does not deliver to a `message` listener on the same window.** Control: the same listener does fire when the event is dispatched manually via `window.dispatchEvent(new MessageEvent('message', {data:'x'}))` — so listener registration works and delivery does not.
5. Minor, but it made all of the above harder to diagnose: obscura does not emit `Runtime.consoleAPICalled` or `Runtime.exceptionThrown`, so Playwright's `page.on("console")` and `page.on("pageerror")` never fire. Shimming `console` via `Page.addScriptToEvaluateOnNewDocument` works as a workaround and is how the Turnstile log above was captured.

## Already ruled out

Verified working on `458d068`, to save you time:

- `attachShadow({mode:'open'|'closed'})`, and an iframe **inside** a shadow root does fire `onload` and expose a `contentWindow`
- cross-origin iframe loading (`https://example.com`: `onload` fires, `contentDocument` correctly null)
- reachability of `challenges.cloudflare.com/cdn-cgi/challenge-platform/...` (real opaque 404 in 114 ms)
- `structuredClone` byte fidelity for `Uint8Array`, and `MessagePort` delivery between two ports in one realm (the #389/#390 fixes are present and correct)
- iframe realm globals (#592/#593): 14/14 present on `main`, 5/14 on v0.2.0

## Environment detail

- `main` @ `458d068`, built with `cargo build --release -p obscura-cli --bins --features render,stealth` (Rust 1.97.1, cmake 4.4.2, `LIBCLANG_PATH=/Library/Developer/CommandLineTools/usr/lib`). Build took 1 m 49 s — V8 arrived prebuilt via `deno_core`.
- Official `obscura-aarch64-macos-stealth.tar.gz` v0.2.0.
- Reference browser: Google Chrome 145 headless via Playwright 1.62.0 (`channel="chrome"`).

xrip:
## What changed

Third of three stacked PRs for #600. **Fixes #600** once all three land.

**Depends on #621 and #622** and is stacked on them. GitHub will not let a cross-fork PR use a fork branch as its base, so this targets `main` and therefore *also contains their commits*. Merge #621, then #622, then this. Its own commit is `feat(cdp): report the page's real child frame tree`.

`Page.getFrameTree` returned `"childFrames": []` however many frames a page had built, and no frame lifecycle event was ever emitted, so Playwright and Puppeteer saw a single-frame page and could not address a child at all. That is the protocol half of the issue report.

- `getFrameTree` now carries the live hierarchy, nesting included. A child's protocol id is derived from its page's frame id and stays stable for the life of the frame, which is what lets a client match an attach event to the frame it later sees in the tree.
- Each child is announced with `Page.frameAttached`, `Page.frameNavigated` and `Page.frameStoppedLoading`, and retracted with `Page.frameDetached`. Attach is emitted before navigate, because a client builds its frame from the attach event and treats navigation of a frame it has never seen as a protocol error.
- The events come from a diff drawn after every dispatch, alongside the existing binding-call drain, rather than from the navigation handler. Script can add an iframe at any time, so a frame that first appears on a later command still gets reported, exactly once.

One behavior change worth review attention: **loading a document now includes building the frames in it**, so navigation does that work instead of leaving it to a caller that settles. Without this a CDP client that only navigates was told the page had no frames, because nothing had given them realms yet. Pages with no iframe skip it after a single native `query_selector("iframe")`, so the common case pays one selector query and no event-loop time. The pumping is bounded (8 rounds x 50ms) because a frame can add a frame and an unbounded loop would never finish on a page that adds one every turn.

## Validation

```
cargo nextest run --release --features render --no-fail-fast                                          # 1416/1417, see note
cargo nextest run --release --no-default-features -p obscura-js -p obscura-browser                    # 342/342
cargo build --release -p obscura-cli --bins --no-default-features                                     # clean
CARGO_INCREMENTAL=0 CARGO_BUILD_JOBS=2 cargo build --release -p obscura-cli --bins --features render  # clean
```

New test `crates/obscura-cdp/tests/child_frame_tree.rs` drives real CDP against a local server whose page embeds a child that itself embeds a grandchild. It asserts the nested tree with correct `parentId` links, that `frameAttached` precedes `frameNavigated` for the same frame, and that a second command does not announce the same frame twice.

The one workspace failure is `obscura-cdp::max_connections_cap max_connections_refuses_then_recovers`, which fails identically on an unmodified `main` worktree in this environment.

Obstacle course: 29/33, **identical to unmodified `main` run back to back on the same machine** (`observer-intersection`, `textdecoder`, `charset-shiftjis`, `fingerprint`; the middle two are Windows console codepage mojibake and the last is local timezone). No regression from this change, but I cannot attest to 33/33 from this environment.

## Rendering

Not applicable.

## Performance

Navigation of a page without an iframe adds one native selector query and nothing else. A page with frames pays bounded pumping while its frames load, which is work the page needs done regardless.

## Checklist

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -2849,6 +2849,13 @@ impl Page {
 
         self.lifecycle = LifecycleState::DomContentLoaded;
 
+        // Before any `wait_until` can return, because the frames belong to the
+        // document rather than to one readiness level. Puppeteer and Playwright
+        // send `Page.navigate` with no `waitUntil`, which lands here and returns
+        // on the next line, so building frames further down left every real CDP
+        // client seeing a page with no frames at all.
+        self.build_document_frames().await;
+
         if wait_until == crate::lifecycle::WaitUntil::DomContentLoaded {
             return Ok(());
         }
@@ -2926,8 +2933,6 @@ impl Page {
             }
             self.lifecycle = LifecycleState::NetworkIdle;
         }
-
-        self.build_document_frames().await;
 
         Ok(())
     }
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -1089,6 +1089,18 @@ impl Page {
         true
     }
 
+    /// Moves the frame tree forward by one step: give any fetched frame
+    /// document a realm, then hand on any message waiting for a realm.
+    ///
+    /// Reports whether anything happened, so a caller can pump again. A new
+    /// frame runs scripts that can post, and a message usually causes a reply,
+    /// so neither queue is finished until both are quiet.
+    async fn advance_frames(&mut self) -> bool {
+        let attached = self.attach_pending_frames().await;
+        let delivered = self.deliver_frame_messages();
+        attached || delivered
+    }
+
     /// URLs of the page's live child frames, in creation order.
     pub fn frame_urls(&self) -> Vec<String> {
         self.frames
@@ -2459,11 +2471,7 @@ impl Page {
                     let _ = js.run_event_loop_until_quiescent(remaining, 150).await;
                 }
             }
-            let attached = self.attach_pending_frames().await;
-            // A message usually causes a reply, so keep going while either the
-            // frame tree or the conversation is still moving.
-            let delivered = self.deliver_frame_messages();
-            if !attached && !delivered {
+            if !self.advance_frames().await {
                 break;
             }
         }
@@ -2503,8 +2511,7 @@ impl Page {
         // realms once at the end instead of being interleaved as in `settle`.
         // Their document scripts still run; only their own deferred work is
         // left for a later settle.
-        self.attach_pending_frames().await;
-        self.deliver_frame_messages();
+        self.advance_frames().await;
     }
 
     /// Advance one wake-driven browser task for a continuously owned page.
@@ -2920,7 +2927,48 @@ impl Page {
             self.lifecycle = LifecycleState::NetworkIdle;
         }
 
+        self.build_document_frames().await;
+
         Ok(())
+    }
+
+    /// Builds the child frames of the document that just loaded.
+    ///
+    /// Loading a document includes loading the frames in it, so this belongs to
+    /// navigation rather than to `settle`: a CDP client that only navigates
+    /// would otherwise be told the page has no frames, because nothing had
+    /// given them realms yet.
+    ///
+    /// A frame's document is fetched by page script, so it arrives from the
+    /// event loop rather than being ready the moment parsing ends. Pages
+    /// without an iframe skip the pumping entirely and pay one native selector
+    /// query.
+    async fn build_document_frames(&mut self) {
+        // How many rounds of "attach a frame, let it start its own" to follow.
+        // A frame can add a frame, so this needs a bound rather than a loop
+        // until quiet: a page that adds one on every turn would never finish.
+        const ROUNDS: usize = 8;
+        const ROUND_MS: u64 = 50;
+
+        let has_iframe = self
+            .with_dom(|dom| dom.query_selector("iframe").ok().flatten().is_some())
+            .unwrap_or(false);
+        if !has_iframe {
+            return;
+        }
+
+        for _ in 0..ROUNDS {
+            if let Some(js) = &mut self.js {
+                let _ = tokio::time::timeout(
+                    tokio::time::Duration::from_millis(ROUND_MS),
+                    js.run_event_loop(),
+                )
+                .await;
+            }
+            if !self.advance_frames().await {
+                break;
+            }
+        }
     }
 
     pub fn navigate_blank(&mut self) {
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use base64::{engine::general_purpose::STANDARD as BASE64, Engine as _};
 use obscura_dom::{parse_html, DomTree};
+use obscura_js::frame::FrameRealm;
 use obscura_js::runtime::ObscuraJsRuntime;
 use obscura_net::{
     CallbackRegistry, ObscuraHttpClient, ObscuraNetError, RequestCallback, ResourceRequest,
@@ -218,6 +219,10 @@ pub struct Page {
     pub frame_id: String,
     pub url: Option<Url>,
     pub dom: Option<DomTree>,
+    /// Live child frame realms, in creation order. Declared before `js` on
+    /// purpose: a realm holds a V8 handle into that isolate, and fields drop in
+    /// declaration order, so the frames must go first.
+    pub frames: Vec<FrameRealm>,
     pub js: Option<ObscuraJsRuntime>,
     pub lifecycle: LifecycleState,
     pub http_client: Arc<ObscuraHttpClient>,
@@ -888,6 +893,7 @@ impl Page {
             frame_id,
             url: None,
             dom: None,
+            frames: Vec::new(),
             js: None,
             lifecycle: LifecycleState::Idle,
             http_client,
@@ -948,6 +954,107 @@ impl Page {
             }
         }
         false
+    }
+
+    /// Gives every frame document the page has fetched a realm of its own, and
+    /// runs the scripts that came with it (issue #600).
+    ///
+    /// Building a realm needs the whole runtime, which an op cannot reach, so
+    /// the JS side queues the fetched document and this drains the queue between
+    /// event loop turns. Reports whether anything was attached, so a caller can
+    /// settle and come back for frames that these frames created.
+    async fn attach_pending_frames(&mut self) -> bool {
+        let pending = match self.js.as_ref() {
+            Some(js) => js.take_pending_frames(),
+            None => return false,
+        };
+        if pending.is_empty() {
+            return false;
+        }
+
+        for frame in pending {
+            let realm = match self.js.as_mut().and_then(|js| {
+                FrameRealm::new(js, frame.frame_id, frame.parent_frame_id, &frame.url, &frame.html)
+            }) {
+                Some(realm) => realm,
+                None => {
+                    tracing::warn!("could not build a realm for frame {}", frame.url);
+                    continue;
+                }
+            };
+
+            // A frame's scripts resolve and are fetched against the frame's own
+            // URL, so they need fetching before run_document_scripts, which
+            // resolves sources synchronously.
+            let wanted = match self.js.as_mut() {
+                Some(js) => realm.external_script_urls(js),
+                None => Vec::new(),
+            };
+            let mut sources: std::collections::HashMap<String, String> =
+                std::collections::HashMap::new();
+            for url in wanted {
+                let Ok(parsed) = Url::parse(&url) else { continue };
+                if self.should_block_url(&url) {
+                    continue;
+                }
+                match self.do_fetch(&parsed).await {
+                    Ok(response) => {
+                        sources.insert(url, String::from_utf8_lossy(&response.body).into_owned());
+                    }
+                    Err(error) => {
+                        tracing::warn!("frame script {} failed: {}", url, error);
+                    }
+                }
+            }
+
+            if let Some(js) = self.js.as_mut() {
+                if let Err(error) = realm.set_viewport(
+                    js,
+                    frame.viewport_width as f64,
+                    frame.viewport_height as f64,
+                ) {
+                    tracing::debug!("frame {} viewport setup failed: {error}", frame.url);
+                }
+                // Page.addScriptToEvaluateOnNewDocument applies to every new
+                // document, including child frames. Debug hooks and browser
+                // automation setup must be present before frame scripts run.
+                for source in &self.preload_scripts {
+                    if let Err(error) = realm.execute_script(js, source) {
+                        tracing::debug!("frame {} preload failed: {error}", frame.url);
+                    }
+                }
+                for problem in realm.run_document_scripts(js, |url| sources.get(url).cloned()) {
+                    tracing::debug!("frame {}: {}", frame.url, problem);
+                }
+                // The frame's scripts have run, so its document is loaded. Say
+                // so: everything a widget defers to DOMContentLoaded or load
+                // hangs on this, which is most of its interface.
+                if let Err(error) = realm.dispatch_load_events(js) {
+                    tracing::debug!("frame {} load events failed: {error}", frame.url);
+                }
+            }
+            self.frames.push(realm);
+        }
+        true
+    }
+
+    /// URLs of the page's live child frames, in creation order.
+    pub fn frame_urls(&self) -> Vec<String> {
+        self.frames
+            .iter()
+            .map(|frame| frame.url().to_string())
+            .collect()
+    }
+
+    /// Evaluates an expression inside one of the page's child frames.
+    pub fn evaluate_in_frame(
+        &mut self,
+        index: usize,
+        expression: &str,
+    ) -> Result<serde_json::Value, String> {
+        let realm = self.frames.get(index).ok_or("no such frame")?;
+        let js = self.js.as_mut().ok_or("no runtime")?;
+        realm.evaluate(js, expression)
     }
 
     /// Update the page's CSS viewport. Calling this before navigation makes
@@ -1076,6 +1183,9 @@ impl Page {
         // attacker-controlled state, trigger a navigation, and then
         // run code in the next document's context.
         if self.js.is_some() {
+            // Every frame realm holds a V8 handle into this isolate, so the
+            // frames of the outgoing document must go before the runtime does.
+            self.frames.clear();
             let _ = self.js.take();
         }
 
@@ -2276,17 +2386,30 @@ impl Page {
             return;
         }
         let settle_started = std::time::Instant::now();
-        if let Some(js) = &mut self.js {
-            if std::env::var_os("OBSCURA_STRICT_SETTLE").is_some() {
-                Self::settle_runtime_for_duration(js, max_ms).await;
-            } else {
-                // A deno_core event loop remains "busy" for any future timer,
-                // including analytics intervals and animation loops which do
-                // not make the page more ready. Require a short window without
-                // observable document/network/script activity instead. The
-                // absolute caller budget and V8 watchdog still bound both
-                // asynchronous work and synchronous microtask storms.
-                let _ = js.run_event_loop_until_quiescent(max_ms, 150).await;
+        // Pump, then give any frame document that finished fetching a realm of
+        // its own. Attaching one runs its scripts, which can start timers,
+        // fetches and further frames, so keep alternating until no new frame
+        // appears or the budget is gone (issue #600).
+        loop {
+            let remaining = max_ms.saturating_sub(settle_started.elapsed().as_millis() as u64);
+            if remaining == 0 {
+                break;
+            }
+            if let Some(js) = &mut self.js {
+                if std::env::var_os("OBSCURA_STRICT_SETTLE").is_some() {
+                    Self::settle_runtime_for_duration(js, remaining).await;
+                } else {
+                    // A deno_core event loop remains "busy" for any future timer,
+                    // including analytics intervals and animation loops which do
+                    // not make the page more ready. Require a short window without
+                    // observable document/network/script activity instead. The
+                    // absolute caller budget and V8 watchdog still bound both
+                    // asynchronous work and synchronous microtask storms.
+                    let _ = js.run_event_loop_until_quiescent(remaining, 150).await;
+                }
+            }
+            if !self.attach_pending_frames().await {
+                break;
             }
         }
         #[cfg(feature = "render")]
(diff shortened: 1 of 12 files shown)


wntic:
I built and tested this stack against #600 (I filed that issue). The engine half is a
real fix — thank you. But the protocol half does not reach a real CDP client, and this
PR's own test does not catch that.

**Setup.** Merge-base of the stack is `main` @ `6fa87e1`; PR #624's head `4fd0654`
carries all four commits (`6a4683d`, `3db9c60`, `bbc7d80`, `4fd0654`), so I built that
head as "stack" and `6fa87e1` as "base", both with
`cargo build --release -p obscura-cli --bins --features render,stealth`.

## What works: `obscura fetch`

My loopback repro from #600 flips completely, with and without `--stealth`:

| loopback repro via CLI | base `6fa87e1` | stack `4fd0654` |
|---|---|---|
| child `<script>` runs (`window.__ran`) | `undefined` | **`"YES"`** |
| child `document.title` | `""` | **`"RAN-IN-CHILD"`** |
| child → parent `postMessage` | `[]` | **`["FROM-CHILD"]`** |

It also covers more than I asked for. A **cross-origin** child (a second loopback port)
runs its script and reaches the parent, and so does the exact shape Turnstile uses — a
cross-origin iframe inside a **closed shadow root** — for both insertion orders
(iframe into shadow then host into document, and the reverse):

```
cross-origin child             : parentGot ["FROM-XO-CHILD"]
closed shadow root + x-origin  : got ["FROM-XO-CHILD", "FROM-XO-CHILD"]
```

## What does not: `obscura serve` + CDP

Same fixtures, same binary, through the CDP server instead of the CLI. No change from
base at all — including the `childFrames` this PR is about:

| child frame via `serve` + CDP | base | stack |
|---|---|---|
| child `<script>` runs | `undefined` | `undefined` |
| child → parent `postMessage` | `[]` | `[]` |
| `Page.getFrameTree` → `childFrames` | `[]` | `[]` |
| `Page.frameAttached` events | 0 | 0 |

Controls I ran so this is not an artifact of my client:

- with `--stealth` and without it — identical
- sampled at 0 / 2 / 5 / 9 s after `Page.navigate`, plus a `Runtime.evaluate` settle
  nudge mirroring what `child_frame_tree.rs` does between navigate and `getFrameTree`
- attaching to the **reused** startup target (`Target.createTarget` returns the
  existing `page-1`) *and* to a **freshly created** one (`page-2`) — identical
- only `Page.frameNavigated` (1) and `Page.frameStoppedLoading` (1) are ever emitted;
  no `Page.frameAttached`

Verbatim tree from the stack build:

```json
{"frameTree": {"frame": {"id": "page-1", "loaderId": "initial-loader",
  "url": "http://127.0.0.1:49361/parent.html", "domainAndRegistry": "",
  "securityOrigin": "http://127.0.0.1:49361/parent.html", "mimeType": "text/html",
  "adFrameStatus": {"adFrameType": "none"}}, "childFrames": []}}
```

## Why the test passes anyway

`crates/obscura-cdp/tests/child_frame_tree.rs` passes in-tree (1/1, and
`child_frame_scripts.rs` 5/5). It drives the protocol like this:

```rust
let mut ctx = CdpContext::new();
let page_id = ctx.create_page();
ctx.sessions.insert(session.to_string(), page_id);
// ... dispatch(&CdpRequest { .., session_id: Some(session) }, &mut ctx)
```

so it calls `dispatch()` against a page from `ctx.create_page()` with a hand-inserted
session entry. A real client instead gets its page and session through
`Target.createTarget` → `Target.attachToTarget {flatten: true}` against
`ws://…/devtools/browser`. The frame work appears not to be wired into that path — and
it is the only path Puppeteer/Playwright can take, which is the scenario the test's own
doc comment describes ("a Playwright or Puppeteer client saw a single-frame page").

A regression test that goes through `Target.createTarget` + `Target.attachToTarget`
would close the gap.

## Reproduction

<details>
<summary>Fixtures (two files) and a static server</summary>

`child.html` and `parent.html` are exactly the two files from #600. Serve them on
loopback:

```bash
python3 -m http.server 8899 --bind 127.0.0.1 --directory ./fixtures &
```
</details>

CLI — fixed on the stack, broken on base:

```bash
obscura --allow-private-network fetch http://127.0.0.1:8899/parent.html --wait 3 -q \
  --eval '(function(){var f=document.querySelector("iframe");return JSON.stringify({
    childRan:(function(){try{return String(f.contentWindow.__ran)}catch(e){return "blocked"}})(),
    childTitle:(function(){try{return f.contentDocument.title}catch(e){return "blocked"}})(),
    parentGot: window.__res ? window.__res.parentGot : "no-res"});})()'

# base   6fa87e1 : {"childRan":"undefined","childTitle":"","parentGot":[]}
# stack  4fd0654 : {"childRan":"YES","childTitle":"RAN-IN-CHILD","parentGot":["FROM-CHILD"]}
```

CDP — unchanged on the stack. Start `obscura serve --port 9222 --allow-private-network`,
then take the browser socket from `/json/version`, `Target.createTarget` →
`Target.attachToTarget {flatten:true}`, pass the `sessionId` on every command,
`Page.navigate` to the same `parent.html`, wait ~6 s, `Runtime.evaluate` once to settle,
then read `window.__res` and `Page.getFrameTree`:

```
__res        = {"parentGot":[],"childRan":"undefined","childTitle":"","iframeLoaded":true}
childFrames  = []
frameAttached events = 0
```

## Secondary notes

- **Still broken, exactly as #621 documents** in its known limitations: `srcdoc`
  iframes never fire `onload` (base and stack alike), and `data:text/html` iframes fire
  `onload` but their inline script stays inert. Flagging only so it is on the record —
  not a request to widen this PR.
- **Turnstile is unchanged**: `300030`, no token, hang at ~43 s, on both the CLI and CDP
  paths. A logging CONNECT proxy (`--proxy` is honored — a dead port fails with
  `ProxyConnect`) shows base and stack are network-identical: 2 connections to
  `challenges.cloudflare.com` each. That measures connections, not requests, so it does
  not say whether the challenge frame itself fetched anything.
- For calibration, headless Chrome 145 on the same page makes 5 requests to
  `challenges.cloudflare.com`, including
  `/cdn-cgi/challenge-platform/h/g/orchestrate/jsch/v1` and a second versioned
  `api.js`. Note that run also had no token and an empty status at 50 s, so Chrome is
  not automatically a pass on that page either.
- Correcting two things I put in #600 that turned out to be bad instrumentation, in
  case they mislead anyone here: `document.querySelectorAll('iframe').length === 0` is
  **not** evidence of a missing widget frame (Chrome reports 0 too — the frame is in a
  closed shadow root), and request counts taken from CDP network events are unusable
  against obscura, which does not report `fetch`/XHR traffic.

## Environment

macOS 26.6.1 arm64 (Apple Silicon), Rust 1.97.1, cmake 4.4.2,
`LIBCLANG_PATH=/Library/Developer/CommandLineTools/usr/lib`. Baseline build 2 m 01 s,
stack 26 s incremental. Reference browser: Google Chrome 145 headless.

xrip:
Thank you — this reproduced exactly, and the test really was blind to it. Fixed in `b5fa41b`.

The cause turned out not to be the target path. It is the `waitUntil` default.

`navigate_single` returns early the moment `wait_until` is `DomContentLoaded`:

```rust
self.lifecycle = LifecycleState::DomContentLoaded;

if wait_until == crate::lifecycle::WaitUntil::DomContentLoaded {
    return Ok(());
}
```

and this PR put `build_document_frames()` at the end of that function, past that return. `Page.navigate` deliberately defaults to `DomContentLoaded` rather than `Load` (`domains/page.rs`, to keep JS-heavy sites from pushing navigation past client timeouts), so a client that sends no `waitUntil` — which is what Puppeteer and Playwright do, and what you did — drove the one path that never built the frames. The CLI defaults to `Load`, which is why your `obscura fetch` column flipped and your `serve` column did not.

Your controls were all sound; the variable that mattered was one neither of us was holding. On your own repro against the **unfixed** stack, adding `waitUntil` to the navigate makes it pass:

```
Page.navigate {url}                     -> childFrames [], frameAttached 0, parentGot []
Page.navigate {url, waitUntil: "load"}  -> childFrames [1], frameAttached 1, parentGot ["FROM-CHILD"]
```

The frames are now built before any `waitUntil` can return, since they belong to the document rather than to one readiness level.

## A second defect your setup exposed

With that fixed, `Page.frameAttached` came out carrying `sessionId: page-1-session` while my client's session was `page-1-session-1`. `drain_frame_events` collapsed every session of a page into one arbitrary `HashMap` entry, and a client drops events addressed to a session it does not hold — so the events were emitted and still invisible. Going through `Target.createTarget` is what surfaces this: it opens a session, and the `Target.attachToTarget` after it opens a second one on the same page. The events now go to every session on the page.

So your instinct that the target path was implicated was right, just one layer over: it is not that the frame work is unwired there, it is that the extra session that path creates was swallowing the announcements.

## Verified the way you asked

Real binary, real `/devtools/browser` socket, `Target.createTarget` → `Target.attachToTarget {flatten:true}` → `Page.navigate` with **no** `waitUntil`:

```
childFrames  : [{"frame":{"id":"page-1-frame-1","parentId":"page-1","url":".../child.html", ...},"childFrames":[]}]
__res        : {"parentGot":["FROM-CHILD"],"iframeLoaded":true}
frameAttached: [{"s":"page-1-session"},{"s":"page-1-session-1"}]    my sessionId: page-1-session-1
```

`child_frame_tree.rs` now does the real `Target.createTarget` + `Target.attachToTarget` handshake instead of inserting a session for a hand-made page, lets `waitUntil` default, and requires every session on the page to receive the attach. Both halves fail without their fix:

- without the navigate fix: `no child frame in the tree: ...childFrames:[]` — your output, verbatim
- without the session fan-out: `session page-1-session-1 on the page was never told the child frame attached`

The second one needed care: asserting only "my session got it" passes roughly half the time on `HashMap` ordering, which is the same coin flip the bug itself was.

## Validation

```
cargo nextest run --release --no-default-features -p obscura-js -p obscura-browser -p obscura-cdp   # 457/458
```

The one failure is `obscura-cdp::max_connections_cap max_connections_refuses_then_recovers`, which I re-ran on the unmodified PR head in the same worktree and it fails identically — unchanged from what this PR already noted.

## On your secondary notes

`srcdoc` and `data:text/html` frames are still as #621 documents them; not widened here. Turnstile is a separate matter and I am not claiming this moves it.

Also noted, and being handled as their own PRs rather than widened into this one:

- `drain_binding_calls` has the identical one-arbitrary-session flaw, so `Runtime.bindingCalled` can be delivered to a session the client does not hold. Pre-existing, and it would bite `exposeFunction` through the same `createTarget` path you used.
- `suspend_js` drops the runtime without clearing `self.frames`, while `init_js` only clears them under `if self.js.is_some()`. After a suspend/resume, `FrameRealm` V8 handles into a destroyed isolate survive.

Thanks again for building both binaries and running the controls — the `serve`-versus-`fetch` split in your table is what made this findable.

xrip:
Follow-up: #624 has been rebased onto this PR's fix (`b5fa41b`) and force-pushed, so the stack is linear again and each PR contains the ones below it. Merge order is unchanged: #621, #622, #623, then #624.

While reviewing #624 on top of the corrected #623, one of its own changes turned out to break the case you had specifically confirmed working.

## The shadow-root frame was being torn down

#624 discards a frame realm when its iframe element leaves the document, which it decided with:

```js
[...document.querySelectorAll('iframe')].map(f => f._frameId >>> 0).filter(id => id)
```

That query does not reach into a shadow root. You made exactly this point in your review here — `document.querySelectorAll('iframe').length === 0` is not evidence of a missing frame, because Chrome reports 0 as well. So an iframe inside a shadow root was read as detached and had its realm destroyed moments after it loaded, along with the published window and document that #624 exists to provide.

It is the one shape where the whole four-PR stack matters most, and it is the shape you verified:

```
closed shadow root + x-origin  : got ["FROM-XO-CHILD", "FROM-XO-CHILD"]
```

Measured on the same fixture, same binary, before and after:

```
closed shadow root, before:  childRan "undefined"  childTitle ""              frameObjects []
closed shadow root, after:   childRan "YES"        childTitle "RAN-IN-CHILD"  frameObjects ["1"]
plain iframe, unchanged:     childRan "YES"                                   frameObjects ["1"]
```

Liveness is now asked of the element through `isConnected`, which reports a shadow-contained iframe correctly where the document query returns nothing:

```
iframeIsConnected true,  querySelectorAll('iframe').length 0
```

A failed liveness query now also leaves the frame tree alone instead of reading as "every frame is gone" — holding a frame is bounded by the existing cap, whereas destroying a live one is not recoverable.

Two regression tests cover both directions: a shadow-root frame survives the sweep and stays reachable, and a genuinely removed iframe still has its realm released with every reference the page realm held for it cleared, so nothing is retained.

## On the rest of the stack

Cross-origin frames still stay opaque — the frame loads and can `postMessage` out, while nothing is published and the page reads `undefined` for anything inside it. The three new bootstrap globals are captured by the snapshot hide list; `Object.getOwnPropertyNames(globalThis)` shows none of them on a running page.

Performance, against this PR's head on the same machine:

```
DOM-heavy page, no frames, in-page ms (median of 7):   #623 382     #624 378
navigation wall clock, 0 frames (avg of 5):            #623 1041ms  #624 1038ms
navigation wall clock, 8 frames (avg of 5):            #623 1069ms  #624 1072ms
```

Suite after the rebase: `462/463` on `-p obscura-js -p obscura-browser -p obscura-cdp --no-default-features`, the one failure being `max_connections_cap`, unchanged from an unmodified worktree here.

## Two things on the record, not in this PR

- `srcdoc` and `data:text/html` frames remain as #621 documents them. Still not widened.
- Four globals (`__obscura_viewport_w`, `__obscura_viewport_h`, `__obscura_screen_emulated`, `__obscura_click_target`) are missing from the snapshot hide list and are enumerable on `window`. That is pre-existing — identical on an unmodified build — and unrelated to the frame work, but it is a detection surface and I did not want to leave it unmentioned given how much of this stack is about not adding one.

Separately, and also found while working on this stack: `Runtime.bindingCalled` had the same one-arbitrary-session flaw the frame events had (#632), and `suspend_js` dropped the page's runtime while its frame realms still held V8 handles into it, which aborts the process (#633).

Issue #629, opened 2026-08-11
Request: support --storage-dir in `obscura mcp` for persistent cookies/session


ety001:
### Feature Description

The `obscura mcp` subcommand (both stdio and `--http` transports) does not support `--storage-dir`, so cookies/localStorage live only in process memory and are lost on restart. Please consider adding `--storage-dir` support to `obscura mcp` (or confirm whether it's planned).

### Motivation

MCP is the primary way agents (Claude Desktop, Claude Code, Codex, custom agent stacks) drive Obscura, and `obscura mcp --http` is the documented sidecar deployment pattern. In Docker/container deployments, persistent sessions matter:

1. `docker run` + `obscura mcp --http` with a mounted volume — users expect `--storage-dir` to work like it does for `serve`/`fetch` so a login survives container restarts.
2. The only current workaround is exporting/importing session state through the `browser_storage_state` / `browser_set_storage_state` MCP tools (agent-side JSON persistence), which works but is manual, fragile for large storage, and not what a volume mount implies.
3. `fetch`, `serve`, and `scrape` all accept `--storage-dir` (confirmed in `crates/obscura-cli/src/main.rs`); `obscura_mcp::run()` / `http::run()` simply don't receive the parameter today.

### Proposed Solution

Wire `--storage-dir` into the `mcp` subcommand and pass it through to the browser/cookie store, with the same semantics as `serve`: all sessions in the process read/write one directory (`cookies.json` + `localStorage/<origin>.json`), written on clean exit and after each navigation.

### Alternatives Considered

- Keeping agent-side `browser_storage_state` JSON round-trips (works, but manual and not volume-mount friendly).
- An env var (e.g. `OBSCURA_STORAGE_DIR`) as an alternative to the flag — either is fine; the flag would match the other subcommands.

xrip:
The Playwright persistence audit found an earlier lower-level break as well: the standard storage-state flow failed at Network.clearBrowserCache, so even cookie/localStorage state handling could not complete. The fork now has a passing offline Playwright test covering load, save, restore, cookie filtering, localStorage by origin, and context isolation. The upstream PR will keep the change focused on this compatibility path and will link this issue.

xrip:
[...]

PR #631, opened 2026-08-11
fix(playwright): persist storageState cookies and localStorage

### LINKED ISSUE #678: sessionStorage does not survive a navigation, and a script in a second target clears it
`sessionStorage` in obscura lives exclusively in the JS realm. obscura tears down the realm in two places: on every navigation and on every switch between two targets of the same CDP connection. After that, an entry is gone.

The second place is the unexpected one. A page loses its store without doing anything itself. It is enough for a script to run in another page of the same connection. No navigation is needed for this, a bare `Runtime.evaluate` on the second target suffices.

**On the relationship to PR #631, because otherwise this issue looks like a duplicate.** On `main` the same defect hits `localStorage` just the same. #631 fixes it there by moving the store out of the realm into `OriginStorage`. I built #631 (`947d08f`) as a release build with `--features render` and measured: `localStorage` afterwards behaves like Chrome in both parts. `sessionStorage` remains just as broken as before, because #631 explicitly leaves it realm-local (`globalThis.sessionStorage = _mkStore(false)`, bootstrap.js:10001 on the PR branch). This is consistent with the purpose of #631, since Playwright's `storageState` does not cover `sessionStorage`. But it means this defect will persist even after #631 is merged.

## Before you start

## obscura version or commit

`d9ef40bea2411e42b1f0e8f6da87206611c2a1a0`, the state of `origin/main` at the time of measurement. Additionally measured against `947d08f`, the head of PR #631.

## OS and architecture

Ubuntu 24.04.4 LTS, x86_64, WSL2 (Kernel 6.18.33.2-microsoft-standard-WSL2)

## How are you using obscura?

## Affected area

JavaScript or DOM behavior

## Build variant

Render-enabled (official default archive or --features render)

## Stealth mode

Off

## Reproduction

Node 22 or newer. No packages needed, the script uses the global `WebSocket`. Tested with Node 24.15.0. Start the engine first, then run the script against its port. `STORE` selects the store so that the same run shows both cases.

```
obscura serve --port 9222
ENDPOINT=http://127.0.0.1:9222 STORE=sessionStorage node repro.mjs
ENDPOINT=http://127.0.0.1:9222 STORE=localStorage   node repro.mjs
```

For comparison, the same against Chrome:

```
<path-to-chrome> --headless --remote-debugging-port=9223 --user-data-dir=$(mktemp -d) about:blank
ENDPOINT=http://127.0.0.1:9223 STORE=sessionStorage node repro.mjs
```

`URL_A` and `URL_B` are just two reachable addresses of different origins. Anyone who redirects them to a local server must start obscura with `--allow-private-network`. Without the flag the engine responds with `Network error: Access to private/internal IP address 127.0.0.1 is not allowed`. Chrome needs nothing for this.

```js
// repro.mjs
const ENDPOINT = process.env.ENDPOINT ?? "http://127.0.0.1:9222";
const URL_A = process.env.URL_A ?? "https://example.com/";
const URL_B = process.env.URL_B ?? "https://example.org/";
const STORE = process.env.STORE ?? "sessionStorage";

const { webSocketDebuggerUrl } = await (await fetch(`${ENDPOINT}/json/version`)).json();
const ws = new WebSocket(webSocketDebuggerUrl);
await new Promise((r) => (ws.onopen = r));

let id = 0;
const pending = new Map();
ws.onmessage = (e) => {
  const m = JSON.parse(e.data);
  const p = pending.get(m.id);
  if (!p) return;
  pending.delete(m.id);
  m.error ? p.rej(new Error(m.error.message)) : p.res(m.result);
};
const send = (method, params = {}, sessionId) =>
  new Promise((res, rej) => {
    const i = ++id;
    pending.set(i, { res, rej });
    ws.send(JSON.stringify({ id: i, method, params, ...(sessionId ? { sessionId } : {}) }));
  });

async function open() {
  const { targetId } = await send("Target.createTarget", { url: "about:blank" });
  const { sessionId } = await send("Target.attachToTarget", { targetId, flatten: true });
  await send("Page.enable", {}, sessionId);
  await send("Runtime.enable", {}, sessionId);
  return sessionId;
}

const evaluate = async (sessionId, expression) =>
  (await send("Runtime.evaluate", { expression, returnByValue: true }, sessionId)).result?.value;

async function goto(sessionId, url) {
  const nav = await send("Page.navigate", { url }, sessionId);
  const want = new URL(url).origin;
  let last = null;
  for (let i = 0; i < 200; i++) {
    last = JSON.parse(await evaluate(sessionId, `JSON.stringify([location.origin, document.readyState])`));
    if (last[0] === want && last[1] === "complete") return;
    await new Promise((r) => setTimeout(r, 100));
  }
  // Do not give up silently. Anyone who continues here reads from a
  // foreign document and takes the result for a finding. The response
  // to Page.navigate comes with it, it carries the reason more often
  // than the state afterwards.
  throw new Error(
    `navigation to ${url} did not settle: origin=${last?.[0]} readyState=${last?.[1]} navigate=${JSON.stringify(nav)}`,
  );
}

// Every read also returns the origin, the full key list, and
// a canary check. The canary check rules out that setItem or
// getItem are simply broken. It says nothing about whether the realm
// is the same.
const read = (key) => `(() => {
  const c = "canary_" + Date.now();
  ${STORE}.setItem(c, "x");
  const canary = ${STORE}.getItem(c) === "x";
  ${STORE}.removeItem(c);
  return JSON.stringify({ origin: location.origin, value: ${STORE}.getItem(${JSON.stringify(key)}), keys: Object.keys(${STORE}), canary });
})()`;

const write = (key, value) => `(() => {
  ${STORE}.setItem(${JSON.stringify(key)}, ${JSON.stringify(value)});
  return JSON.stringify({ origin: location.origin, readback: ${STORE}.getItem(${JSON.stringify(key)}) });
})()`;

console.log(`store = ${STORE}\n`);

console.log("Part 1: does the entry survive a navigation?");
const a = await open();
await goto(a, URL_A);
console.log("  written            ", await evaluate(a, write("k", "value")));
console.log("  read, no navigation", await evaluate(a, read("k")));
await goto(a, URL_A);
console.log("  read, after nav    ", await evaluate(a, read("k")));

// Own key, own value. Part 1 wrote to the same origin.
// A store shared across all pages of one origin would still hold the entry
// from Part 1 here. The same key would then leave open which of the two
// you just read.
console.log("\nPart 2: does a second target wipe the first?");
const t1 = await open();
await goto(t1, URL_A);
console.log("  T1 written         ", await evaluate(t1, write("t1_k", "t1_value")));
const t2 = await open();
console.log("  T1 after T2 created", await evaluate(t1, read("t1_k")));
await goto(t2, URL_B);
console.log("  T1 after T2 nav    ", await evaluate(t1, read("t1_k")));
// T3 is deliberately on THE SAME origin as T1. This rules out the second
// explanation, that the connection maintains a store that follows the
// origin of the last-addressed target. Under this explanation T1 would
// have to keep its entry, since the origin does not change.
const t3 = await open();
await goto(t3, URL_A);
// From here on, nothing navigates. Between the write and the read on T1
// stands exactly one Runtime.evaluate on T3.
console.log("  T1 re-written      ", await evaluate(t1, write("t1_k", "t1_value")));
console.log("  T1 before T3 eval  ", await evaluate(t1, read("t1_k")));
console.log("  T3 evaluated       ", await evaluate(t3, `String(1 + 1)`));
console.log("  T1 after T3 eval   ", await evaluate(t1, read("t1_k")));

ws.close();
process.exit(0);
```

## Expected behavior

`sessionStorage` belongs to a browsing context and an origin. An entry survives a navigation to the same origin in the same tab. A page's store remains unaffected when another page navigates or runs a script.

Chrome for Testing 152.0.7977.8 with `STORE=sessionStorage`:

```
store = sessionStorage

Part 1: does the entry survive a navigation?
  written             {"origin":"https://example.com","readback":"value"}
  read, no navigation {"origin":"https://example.com","value":"value","keys":["k"],"canary":true}
  read, after nav     {"origin":"https://example.com","value":"value","keys":["k"],"canary":true}

Part 2: does a second target wipe the first?
  T1 written          {"origin":"https://example.com","readback":"t1_value"}
  T1 after T2 created {"origin":"https://example.com","value":"t1_value","keys":["t1_k"],"canary":true}
  T1 after T2 nav     {"origin":"https://example.com","value":"t1_value","keys":["t1_k"],"canary":true}
  T1 re-written       {"origin":"https://example.com","readback":"t1_value"}
  T1 before T3 eval   {"origin":"https://example.com","value":"t1_value","keys":["t1_k"],"canary":true}
  T3 evaluated        2
  T1 after T3 eval    {"origin":"https://example.com","value":"t1_value","keys":["t1_k"],"canary":true}
```

## Actual behavior

obscura on `d9ef40b` with `STORE=sessionStorage`:

```
store = sessionStorage

Part 1: does the entry survive a navigation?
  written             {"origin":"https://example.com","readback":"value"}
  read, no navigation {"origin":"https://example.com","value":"value","keys":["k"],"canary":true}
  read, after nav     {"origin":"https://example.com","value":null,"keys":[],"canary":true}

Part 2: does a second target wipe the first?
  T1 written          {"origin":"https://example.com","readback":"t1_value"}
  T1 after T2 created {"origin":"https://example.com","value":"t1_value","keys":["t1_k"],"canary":true}
  T1 after T2 nav     {"origin":"https://example.com","value":null,"keys":[],"canary":true}
  T1 re-written       {"origin":"https://example.com","readback":"t1_value"}
  T1 before T3 eval   {"origin":"https://example.com","value":"t1_value","keys":["t1_k"],"canary":true}
  T3 evaluated        2
  T1 after T3 eval    {"origin":"https://example.com","value":null,"keys":[],"canary":true}
```

The output for `STORE=localStorage` is the same on `d9ef40b`, except for the header line `store =`. On the branch of #631 both parts pass for `localStorage`, while `sessionStorage` keeps the above output verbatim.

| Step | main `d9ef40b` | PR #631 `947d08f` | Chrome 152.0.7977.8 |
|---|---|---|---|
| `localStorage` after navigation, key `k` | `null` | `"value"` | `"value"` |
| `localStorage` after a `Runtime.evaluate` on a second target, key `t1_k` | `null` | `"t1_value"` | `"t1_value"` |
| `sessionStorage` after navigation, key `k` | `null` | `null` | `"value"` |
| `sessionStorage` after a `Runtime.evaluate` on a second target, key `t1_k` | `null` | `null` | `"t1_value"` |

**With multiple pages, not a single one holds anything.** Three pages of the same origin, each writes its own key and reads it back immediately. All three reads return the value. A second read pass over all three finds **not a single one** still filled in obscura, not even the last one written. Reading the first page already clears the realm of the third. Chrome holds all three. The procedure is the same as above, only with three pages instead of one: per page `Target.createTarget`, `Page.navigate` to the same address, one `setItem` and one `getItem`, then a second read pass over all three.

`947d08f` is 44 commits behind `d9ef40b`. The column therefore proves what #631 does **not** change about `sessionStorage`, and is not a comparison of two equally current states.

### The null proves something

A missing value and a broken store look the same from the outside. That is why every read returns three additional pieces of information. `origin` shows that the read happened under the same origin. `keys` shows the entire store and not just the sought key. `keys` is empty. `canary` writes a fresh key and reads it back immediately, every read line above shows `canary: true`. This rules out that `setItem` or `getItem` are simply broken.

The canary check runs in the same `Runtime.evaluate` as the read, i.e. in the new realm. The canary check therefore does not distinguish a cleared store from a freshly created one. This is exactly the case here, the next section shows why.

**In the second part** obscura sends no `Runtime.executionContextsCleared` for the losing target between the write and the lost read. This measurement is not included here. The client thus sees no context switch. The page's store is gone nonetheless. In the first part obscura does send the event, since there the target navigates itself.

### The cause in the code

The store lives in the realm. `crates/obscura-js/js/bootstrap.js:10064` on `d9ef40b` sets `globalThis.sessionStorage = _mkStore()`, and the created object holds its data in a JS object in the same realm.

The realm is torn down in two places.

**First, on every navigation.** `Page::init_js` in `crates/obscura-browser/src/page.rs:1387` discards an existing runtime and creates a new one. The comment from line 1393 names a good reason. The rebuild came with commit `9edde41` as a security fix: a page on http could set `window.onload`, navigate to a `file:` address, and continue running there in the state of the previous document. Anyone who pulls the store out of the realm must preserve this guarantee.

**Second, on addressing another target.** `CdpContext::get_session_page_mut` in `crates/obscura-cdp/src/dispatch.rs:321` holds at most one live isolate per connection. The function suspends the runtime of another page as soon as the requested page does not have one. After that `get_session_page_mut` wakes the target page:

```rust
if !target_has_js {
    for page in &mut self.pages {
        if page.id != page_id && page.has_js() {
            page.suspend_js();
            break;
        }
    }
    if let Some(target) = self.pages.iter_mut().find(|p| p.id == page_id) {
        target.resume_js();
    }
}
```

`Page::suspend_js` (`page.rs:3940`) sets `self.js = None`. `Page::resume_js` (`page.rs:3961`) calls `init_js` and thus builds a fresh realm. The wake-up is not a restore.

`suspend_js` saves two things across the switch. The function pulls the DOM out with `js.take_dom()` and sets aside the `started_script_ids`. `sessionStorage` needs the same treatment and is the only one of the three not to get it.

**The trigger is not the navigation.** Two places clear foreign realms. One clears one page, the other all of them.

`get_session_page_mut` suspends **one** other page as soon as the requested page has no realm. `Runtime.evaluate` takes this path, `Page.navigate` does too. `is_v8_free_method` (`dispatch.rs:352`) lists 60 methods that do not take it. The list is not an inversion of this rule: `Audits.enable`, which Puppeteer sends on connect according to the comment at `dispatch.rs:509`, is also not in it and still does not touch any page (`dispatch.rs:512`).

`process_with_interception` (`server.rs:1151`) suspends **every** other page with a realm (`server.rs:1202`). Every `Page.navigate` runs through there (`server.rs:880`). A single navigation thus costs the store of all other pages.

The last four lines of the reproduction show the case without navigation. Between `T1 before T3 eval` and `T1 after T3 eval` stands exactly one `Runtime.evaluate` on T3, and the value changes from `"t1_value"` to `null`. **T3 is on the same origin as T1.** This rules out the obvious second explanation, that the connection maintains a store that follows the origin of the last-addressed target. Under this explanation T1 would have to keep its entry. The branch of #631 does manage `localStorage` by origin, and the same line shows `"t1_value"` there.

Merely creating a second target does nothing on its own. The reason is a different one. The handler for `Target.createTarget` takes `ctx.get_page_mut` (`target.rs:91`), and `get_page_mut` is just a `find` over the page list (`dispatch.rs:286`). This handler never reaches `get_session_page_mut`. `Target.attachToTarget`, `Page.enable`, and `Runtime.enable` are all in the list as well. The path through `Target.createTarget` thus suspends nothing. The new page is created via `navigate_blank`, which sets `self.js = None` (`page.rs:3139`), and thus remains without a realm. The first method that does go through `get_session_page_mut` for this target finds it without a realm and takes the realm away from the other page.

### A note on the implementation

`OriginStorage` from #631 is attached to the `BrowserContext` and shares the data between all pages of the same context. This is correct for `localStorage` and too broad for `sessionStorage`: there an entry must not leave the tab. The fitting scope would be the same store, but held per page instead of per context, and still separated by origin.

The path through `suspend_js` and `resume_js` would be the smaller change. `suspend_js` already saves the DOM across the switch. Placing a snapshot of `sessionStorage` alongside it and playing it back in `resume_js` only fixes the second part, however. The first part, the loss on navigation, needs a store outside the realm, because `init_js` rebuilds on every navigation.

I am happy to build the patch. I wanted to record the finding separately first, so that it is not considered done once #631 is merged.

## Logs

obscura does not report the loss on any channel. In the four obscura runs of the reproduction above, stderr without `--verbose` was empty. The engine wrote 1669 lines with `--verbose`. None of them contain `storage`, `suspend`, `resume`, `realm`, or `isolate`. A client sees an empty store and gets no indication that it was ever filled.

xrip:
## What changed

Adds Playwright `BrowserContext.storageState` compatibility for cookies and origin-scoped localStorage. Storage is shared by pages in one context and isolated between contexts. Adds `Network.clearBrowserCache` as a protocol-compatible no-op and `Storage.clearCookies`, which Playwright uses while loading state. Adds an offline Playwright regression test and documents save/restore usage.

Also fixes #678: `sessionStorage` previously lived in the JS realm and died on every navigation (`init_js` rebuilds the realm, the 9edde41 security fix) and on every CDP target switch (`suspend_js`/`resume_js`). It now lives in a page-owned `OriginStorage` keyed by origin, the same shape localStorage got at the context level here. An entry survives same-origin navigation in the same tab, is untouched when another target navigates or runs a script, never leaves the tab, and keeps the 5MB/origin quota. Realm teardown on navigation is unchanged.

Closes #678. Related to #629.

## Validation

- `cargo build --release -p obscura-cli --bins --features render`
- `cargo nextest run --release --features render --no-fail-fast` (1473 passed, 4 skipped)
- `node crates/obscura-cdp/tests/playwright_storage_state.mjs` (pass with Playwright 1.62.1)
- `node crates/obscura-cdp/tests/session_storage.mjs` (pass: survives same-origin navigation, survives other targets' navigation and evaluate, fresh tab starts empty)
- The exact repro from issue #678 (`STORE=sessionStorage` and `STORE=localStorage`) now matches the Chrome 152 column of the issue on both parts
- Obstacle course: 29/33 on this host, and the 4 failures (`observer-intersection` per #671, `textdecoder`, `fingerprint`, `charset-shiftjis`) reproduce identically on an unmodified `origin/main` build on the same host, so they are pre-existing environment failures, not regressions
- Leak probe: 60 tabs x 512 KiB of sessionStorage take RSS from 19 to 65 MB; after `Target.closeTarget` on all of them RSS returns to 26 MB and a fresh tab on the same origin starts with an empty store

## Rendering

Not applicable.

## Performance

No expected hot-path impact. localStorage and sessionStorage have bounded per-origin, total-byte, and origin-count limits.

## Checklist

- [x] Existing relevant tests pass in the no-render configuration.
- [x] Public API and user-facing behavior changes are documented.


xrip:
## Added: sessionStorage persistence (issue #678)

This branch was rebased onto main (the `ops.rs` conflict was a context merge of the new `op_local_storage` with a comment main added above `op_navigate`; both kept) and extended so the PR now also closes #678.

### What was added

- `sessionStorage` now lives outside the JS realm. The `Page` owns an `Arc<OriginStorage>` keyed by origin (page-scoped, not context-scoped, the same shape localStorage got at the context level in this PR). The realm is still torn down on every navigation as before (the 9edde41 security fix is untouched); `init_js` just re-attaches the same store to the new realm.
- One new op, `op_session_storage`; `op_local_storage` and the new op share a single `origin_storage_command` dispatch, so the command handling is not duplicated.
- `bootstrap.js`: both stores go through one `_storageOp(slot)` path. The old realm-local fallback map (`slot.data`) is deleted, so there is exactly one implementation of each store.
- New regression test `crates/obscura-cdp/tests/session_storage.mjs` (raw CDP, no dependencies) covering all three scenarios from the issue:
  1. an entry survives a same-origin navigation in the same tab;
  2. another target navigating (any origin) or running a bare `Runtime.evaluate` on a same-origin target does not wipe the first tab's store;
  3. a fresh tab on the same origin starts with an empty store (tabs stay isolated).

### Iframe storage scoped to the calling realm (commit 4d5cf1f)

While reviewing the branch I found the pre-existing frame gap is the opposite of what it looks like from `frame.rs`: the storage ops resolved the page-level state, so every frame realm read and wrote the top document's origin bucket. A cross-origin iframe could read and overwrite the top origin's `localStorage` and `sessionStorage`. Same-origin frames were already sharing through that same path, which is why no per-frame store was ever observable.

- `op_local_storage` / `op_session_storage` now resolve `realm_state`, the same per-realm routing `op_get_cookies` and `op_navigate` use, so a frame lands in the bucket keyed by its own URL.
- `share_resources_with` copies the page's local and session storage Arcs into the frame state (same trust class as the cookie jar it already shares), so same-origin sharing is unchanged and the frame state's `url` selects the correct origin.
- Tests in `obscura-browser`: `same_origin_frames_share_the_page_stores` (both directions, fails without the Arc sharing) and `cross_origin_frames_get_their_own_storage_buckets` (full isolation, both stores).
- Live CDP check: a same-origin frame's writes are visible to the parent and vice versa; a cross-origin frame's writes are invisible to the parent and the parent's entries invisible to the frame. Matches Chrome.

### Verification

- The exact repro from the issue, `STORE=sessionStorage` and `STORE=localStorage`, now matches the Chrome 152 column in both parts.
- `cargo nextest run --release --features render --no-fail-fast`: 1475 passed, 4 skipped.
- `node crates/obscura-cdp/tests/playwright_storage_state.mjs`: pass (Playwright 1.62.1).
- Obstacle course: 29/33 on the test host, and all 4 failures (`observer-intersection` per #671 plus 3 host-environment stages) reproduce identically on an unmodified `origin/main` build on the same host, so none are regressions.
- Leak probe: 60 tabs x 512 KiB of sessionStorage take RSS from 19 to 65 MB; after `Target.closeTarget` on all of them RSS returns to 26 MB and a fresh tab on the same origin is empty, so closed tabs' stores are dropped with the page. Growth is bounded by the same quota as localStorage (5 MB/origin, 32 MB total, 256 origins).

### Deliberately not addressed

- The `tech()` skip rule and font priority walking (belongs to #667).

PR #697, opened 2026-08-23
fix(cli): await promise-returning --eval expressions

### LINKED ISSUE #693: fetch --eval does not await returned Promises — async evals serialize as {}
### Version / platform
- obscura 0.2.0 (official aarch64-linux release binary)
- Ubuntu 22.04 proot on Termux (Android aarch64), kernel-shared loopback

### Steps to reproduce
```bash
obscura fetch "http://127.0.0.1:3000/auth" \
  --allow-private-network \
  --eval "(async () => {
    const r = await fetch('/api/version');
    return JSON.stringify({ ok: r.ok });
  })()" \
  --wait-until networkidle0 --timeout 30 --quiet
```

### Actual
Prints `{}` — the Promise returned by the async IIFE is serialized like a plain object before it settles.

### Expected
Either:
- `Runtime.evaluate` should pass `awaitPromise: true` so resolved values are printed, or
- an explicit opt-in flag (e.g. `--eval-await`) that awaits async results.

### Why it matters
Single-invocation E2E scripting is impossible without this: any flow that needs network I/O inside the page (e.g. "call the app's own login helper, then inspect localStorage") requires promises. Current workaround is synchronous XHR inside the eval, which is fragile and can't express async app APIs (`window.__isoLogin(...)` returns a Promise).

### Environment details
- Obscura version: 0.2.0 (release binary)
- OS: Ubuntu 22.04 (proot) on Android/Termux aarch64

Suydev:
Fixes #693.

## Problem

`obscura fetch --eval <expr>` used the synchronous `Page::evaluate` path, which serializes an unresolved Promise as `{}`. Any async IIFE printed `{}` instead of its result:

```bash
obscura fetch URL --eval "(async () => { const r = await fetch('/api'); return JSON.stringify({ok: r.ok}); })()"
# → {}
```

This makes single-invocation E2E flows impossible without a CDP client — e.g. "call the app's own async login helper, then inspect localStorage".

## Fix

Route `WorkerCommand::Evaluate` through `Page::evaluate_for_cdp_with_timeout(expression, return_by_value=true, await_promise=true)` with the same 30s cap the CDP path uses:

- resolved promises serialize their actual value (`v8_to_json` → `info_from_json`)
- plain values behave exactly as before
- a never-settling promise cannot hang the worker (30s cap)

The JS runtime has implemented this await path for CDP clients all along — the CLI simply wasn't using it.

## Testing

Not built locally (aarch64 phone; V8 from source is out of reach) — relying on this repo's PR CI to validate the build across all four feature configs. Behavior verified by reading the runtime paths: `evaluate_for_cdp_with_timeout` with `await_promise && return_by_value` returns via `info_from_json(&json_val)`, so non-promise expressions keep byte-identical output through the same `info_from_json` helper the old sync path used.

### DIFF
--- a/crates/obscura-cli/src/worker.rs
+++ b/crates/obscura-cli/src/worker.rs
@@ -106,7 +106,22 @@ async fn main() {
                 }
             }
             WorkerCommand::Evaluate { expression } => {
-                let result = page.evaluate(&expression);
+                // Await promise-returning expressions so async IIFEs resolve
+                // before serialization. Previously the sync path serialized an
+                // unresolved Promise as `{}`, making single-invocation flows
+                // that call async app APIs impossible (issue #693). A 30s cap
+                // matches the CDP await timeout so a never-settling promise
+                // cannot hang the worker.
+                let result = match page
+                    .evaluate_for_cdp_with_timeout(&expression, true, true, 30_000)
+                    .await
+                {
+                    Ok(info) => match info.value {
+                        Some(v) => v,
+                        None => serde_json::Value::String(info.description),
+                    },
+                    Err(_) => serde_json::Value::Null,
+                };
                 WorkerResponse::success(result)
             }
             WorkerCommand::Title => {


Suydev:
Heads up for maintainers 👋 The CI run on this branch shows `action_required` — since this is my first contribution here, the workflow needs your "Approve and run" click before it can build. Happy to adjust anything if the build surfaces issues. Thanks!

SGavrl (approved):

PR #712, opened 2026-08-24
fix(cdp): escape control characters in Runtime.getProperties objectId embed

### LINKED ISSUE #709: fix(cdp): Runtime.getProperties objectId escaping omits control characters
## Summary

The inline objectId escaping in `Runtime.getProperties` handles `\` and `'` but not control characters (newline 0x0A, carriage return 0x0D, NUL). A literal newline in an objectId breaks the generated single-quoted JS string.

## Location

`crates/obscura-cdp/src/domains/runtime.rs` (~L242):

```rust
let escaped_oid = oid.replace('\\', "\\\\").replace('\'', "\\'");
// interpolated as '{oid}' (=escaped_oid via named arg) into __obscura_objects['{oid}']
```

## Assessment

Not an escalation — CDP clients already have full `Runtime.evaluate` access, so this cannot grant capability a caller lacks. It is an inconsistency: other objectId sites use robust escaping (`escape_object_id` in dom.rs) or `serde_json::to_string` (input.rs), which cover control characters.

## Fix direction

Embed the objectId via `serde_json::to_string(oid)` (a JSON double-quoted string literal that escapes control characters), replacing the manual single-quote splice at both interpolation sites.

## Severity

LOW — robustness/consistency; no privilege escalation.

mnaza:
The inline objectId escaping in `Runtime.getProperties` handled only backslash and single-quote, leaving control characters (newline, CR, NUL) raw — a literal newline in an objectId broke the generated single-quoted JS string. Not an escalation (CDP clients already have full `Runtime.evaluate`), but inconsistent with the robust escaping used at the other objectId sites.

Routes both interpolation sites through an `object_id_js_literal` helper using `serde_json::to_string` (a double-quoted JS/JSON literal with every control char escaped). Normal objectIds embed identically to before.

TDD: RED left raw newline/CR/tab, GREEN escapes them and round-trips as JSON. Full 126-test obscura-cdp suite green.

Closes #709

### DIFF
--- a/crates/obscura-cdp/src/domains/runtime.rs
+++ b/crates/obscura-cdp/src/domains/runtime.rs
@@ -17,6 +17,16 @@ fn is_valid_binding_name(name: &str) -> bool {
         && !name.chars().next().unwrap_or('0').is_ascii_digit()
 }
 
+/// Render a CDP objectId as a self-contained JS string literal, safe to
+/// interpolate into generated code (e.g. `__obscura_objects[<here>]`). Escapes
+/// every character that could terminate or break the literal — including
+/// control characters like newline — which a plain `\\`/`'` replacement misses.
+fn object_id_js_literal(oid: &str) -> String {
+    // serde_json produces a double-quoted JS/JSON string literal with every
+    // control character escaped, so the id can never terminate the literal.
+    serde_json::to_string(oid).unwrap_or_else(|_| "\"\"".to_string())
+}
+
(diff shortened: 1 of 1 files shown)


mnaza:
Superseded by #723 (`fix/getproperties-objectid-json-literal`), which landed the same fix and more broadly: main now embeds the objectId via a shared `crate::util::object_id_literal` = `serde_json::to_string(oid)` at every lookup site, identical to this PR's `object_id_js_literal`. Closing as redundant — no change needed here. Thanks @alkaz-nodemaven.

Issue #716, opened 2026-08-25
Binary fetch and interception bodies are corrupted by UTF-8 conversion


enginespot:
### Before you start

- [x] I searched existing issues and this is not a duplicate.
- [x] I tested against the latest release or `main`.
- [x] This is not a security vulnerability.

### obscura version or commit

`f449e6fb3183138eb3e80f11fe44af31cefe0fae` (`main`)

### OS and architecture

macOS arm64

### How are you using obscura?

Embedded Rust library and CDP server.

### Affected area

JavaScript fetch/XHR, networking, and CDP request interception.

### Build variant

Render-enabled. The stealth transport is affected by the same string conversion.

### Stealth mode

Both off and on.

### Reproduction

Send a request body containing bytes that are not valid UTF-8:

```js
fetch(url, {
  method: "POST",
  body: new Uint8Array([0, 128, 255, 16]),
});
```

The same problem is observable with `Blob`, `ArrayBuffer`, sliced typed-array views, binary `File` values in `FormData`, and `XMLHttpRequest.send()`.

For the CDP fulfill path, resolve an intercepted request with:

```json
{
  "method": "Fetch.fulfillRequest",
  "params": {
    "responseCode": 200,
    "body": "AP+Awyg="
  }
}
```

`AP+Awyg=` represents the bytes `[0, 255, 128, 195, 40]`.

### Expected behavior

Request and fulfilled response bodies remain byte-for-byte identical across the JavaScript, Rust transport, public interception API, and CDP boundaries.

### Actual behavior

Binary request bodies pass through JavaScript/Rust strings and are UTF-8 encoded. For example, `[0, 128, 255, 16]` reaches the server as `[0, 194, 128, 195, 191, 16]`. A `Request` object's inherited body is also dropped when `fetch(request)` has no explicit `init.body`.

The CDP fulfill bridge decodes base64 and then uses lossy UTF-8 conversion, replacing non-UTF-8 bytes. CDP `continueRequest.postData` is also modeled as a string instead of raw bytes.

### Does headless Chrome behave the same?

No. Chrome preserves the original request and fulfilled response bytes.

### Proposed split

1. Preserve binary JavaScript fetch/XHR bodies through the normal and stealth transports without changing the public interception API.
2. Change interception request/response bodies to `Vec<u8>` and decode CDP base64 directly into bytes.

enginespot:
[...]

PR #718, opened 2026-08-25
fix(js)!: preserve byte-native interception bodies

enginespot:
## What changed

Based on upstream main at `89964b3ef9b4132f679558e6b6685e086648b955`.

Interception body overrides still used text even after #717 fixed normal request transport. This changes `Continue.body` to `Option<Vec<u8>>` and replaces `Fulfill.body` plus `Fulfill.body_base64` with a single `Vec<u8>`.

CDP request and fulfilled-response bodies are decoded from Base64 before resolving the paused request. Invalid Base64 or a non-string body returns `-32602` and leaves the request available for a corrected command. The JavaScript response retains both its text view and exact `bodyBase64` representation, preserving the upstream #912 fix.

This addresses the interception portion of #716. #717 is already merged; XHR response handling and response-body retention remain separate concerns. No new issue or dependent branch is required.

This is a breaking Rust API change: callers supply raw bytes and no longer provide a separate `body_base64` field.

## Validation

- Release render CDP library and public interception tests: 137 passed.
- No-render check for `obscura-cdp` and `obscura`: passed.
- No-render focused Continue, Fulfill, invalid-input and public interception tests: 5 passed.
- Retained #912 binary fulfilled-response regression: passed.
- Full release render suite: 1,710 passed, 2 failed, 4 skipped. Both failures (`a_miss_created_inside_an_awaited_expression_loads_during_the_wait` and `timers_inside_a_fixed_wait_observe_bytes_that_land_during_it`) reproduce with the same font-width assertions on unmodified main at the baseline above.
- Release CLI build with render: passed.
- Obstacle course: 32/33 passed. The `observer-intersection` stage also fails on unmodified main with the same result (expected `io:50`, received an empty string).
- `git diff --check`: passed.

## Rendering

Not applicable.

## Performance

The wire format is unchanged. Rust interception uses one byte buffer instead of caller-maintained text and Base64 copies. No additional network requests or dependencies are introduced.

### DIFF
--- a/crates/obscura-cdp/src/server.rs
+++ b/crates/obscura-cdp/src/server.rs
@@ -3,6 +3,7 @@ use std::net::SocketAddr;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::Arc;
 
+use base64::Engine as _;
 use futures_util::{SinkExt, StreamExt};
 use serde_json::json;
 use tokio::net::TcpStream;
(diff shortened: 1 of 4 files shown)


yinnho:
We hit the same need from the other direction: we shipped our CDP Fetch domain with String bodies first (`from_utf8_lossy` on the decoded base64), and a synthetic binary fulfillment corrupted every non-ASCII byte before the page ever saw it — noticed when an `application/octet-stream` fulfill round-tripped through `arrayBuffer()` with 0x80+ bytes mangled. Moving the boundary to bytes fixed it: decode the wire base64 straight to `Vec<u8>` and let the synthetic body ride the same base64 envelope channel real network responses use; the JS side re-derives text through the response's charset.

One semantic detail worth pinning: an invalid-base64 resolution errors as a protocol error but does not consume the pause — the same requestId stays answerable afterwards. We have a regression test asserting exactly that (bad `fulfillRequest`, then a good one on the same requestId, page sees the good body), which I think matches your "-32602 without resolving the paused request" choice.

Byte-exactness invariant, for the test suite here: fulfill with `[0, 1, 2, 0x7f, 0x80, 0xfe, 0xff, 0x0a, 0x00]`, read `new Uint8Array(await response.arrayBuffer())` in the page, compare equal. That's the check the String pass failed.

enginespot:
@SGavrl  This PR is now ready for review and has been updated against the latest main. It addresses the remaining interception-body portion of #716, preserves the #912 fix, and includes byte-exact regression tests. This changes the public interception API to use Vec<u8>; could you review whether this API direction fits the project? Full validation results are in the PR description.

Issue #557, opened 2026-08-04
fix(cdp): resolve_node_id objectId path does not escape backslashes


mnaza:
### Summary
`resolve_node_id`'s `objectId` path escapes `'` but not `\` when interpolating the client-supplied objectId into evaluated JS, so an objectId containing a backslash produces malformed JS and the node fails to resolve. The sibling `describeNode` handler in the same file escapes both.

### Details
crates/obscura-cdp/src/domains/dom.rs:
```rust
oid.replace('\'', "\\'")                       // resolve_node_id (~line 22)
oid.replace('\\', "\\\\").replace('\'', "\\'") // describeNode (~line 143)
```
For an objectId ending in a backslash, e.g. `x\`, the first form yields `['x\']` where `\'` escapes the closing quote → a JS syntax error → the eval returns null → `DOM.scrollIntoViewIfNeeded` / `focus` / `getBoxModel` (anything routing through `resolve_node_id`) errors with "could not be resolved to a node".

This is **not** an injection: all `'` are still escaped, so a crafted id cannot break out of the string to execute code — the worst case is an unterminated string literal → syntax error → clean `Err`. Server-generated objectIds never contain backslashes, so it is practically unreached; this is a robustness/consistency fix.

### Fix
Escape backslashes before single quotes, matching `describeNode`:
```rust
oid.replace('\\', "\\\\").replace('\'', "\\'")
```

PR #723, opened 2026-08-25
fix(cdp): embed objectId as a JSON literal at every lookup site

### LINKED ISSUE #709: fix(cdp): Runtime.getProperties objectId escaping omits control characters

alkaz-nodemaven:
## What changed

`Runtime.getProperties` escaped the objectId with `oid.replace('\\', "\\\\").replace('\'', "\\'")` and spliced the result into `__obscura_objects['{oid}']`. That pair covers the two characters that close a single-quoted literal and leaves every C0 control alone, and a raw newline ends a JS string literal exactly as a stray quote does. The generated snippet is then a syntax error, `page.evaluate` returns no array, and the handler falls through to `result: []`. Nothing reports an error, so a client walking a nested object sees an object that has no properties rather than a failure.

Both interpolation sites switch to a JSON literal: the `__obscura_objects[...]` lookup and the `childOid = oid + '::' + k` construction. Switching only the first silently reintroduces the break on nested properties, as @yinnho noted on the issue.

Two things I found while writing the fix that the issue does not cover:

**The ids are not all client-supplied.** `getProperties` mints child ids as `parent + "::" + key`, and `key` is a property name off a page object. The page, not the client, decides which characters land inside that literal, and a client cannot sanitise them away because it never sees the key until an objectId has already been built out of it.

**`escape_object_id` in `dom.rs` had the same hole.** The issue cites it as an example of robust escaping, but it was the same two `replace` calls. It matters because the page-minted id does not stay inside `Runtime`: Puppeteer's `$$` flow is evaluate -> getProperties -> asElement, and asElement handles go on to `DOM.describeNode` / `DOM.resolveNode`. There the fallback is worse than an empty list. `describeNode`'s `unwrap_or(0)` answers with node 0, so a handle that failed to resolve comes back described as `#document` instead of erroring, and the client is handed the wrong element. The included test asserts that case directly.

Both files now go through `util::object_id_literal`, which is `serde_json::to_string`. JSON string syntax is a subset of JavaScript's, so one call covers the quote and backslash the manual pair handled plus every control it did not. Sharing one helper is what stops the two domains drifting apart again, which is how `dom.rs` ended up with its own copy of the bug. This is the same approach the project already took for `Input.insertText` in 524f2a9.

Removing the per-domain helper also removes its unit test; the assertions it made were carried over into the `object_id_literal` tests, including the trailing-backslash case from #557.

## Validation

New integration file `crates/obscura-cdp/tests/runtime_get_properties_objectid_escaping.rs`, two tests. Keys are assembled with `String.fromCharCode` so the characters under test are exactly LF, CR, NUL and `'`+`\`, not an artefact of how the Rust literal was written. The quote/backslash pair the old code did handle is included so a regression in either direction shows up in one run.

Both fail on `main` and pass with the change. On `main`:

```
get_properties_walks_into_a_child_whose_key_holds_a_control_character
  walking into the child keyed "lf\nx" must reach its properties, not fall
  through to an empty list; got Object {"result": Array [], "internalProperties": Array []}
  left: None   right: Some("lf")

describe_node_resolves_a_child_handle_whose_key_holds_a_control_character
  the handle keyed "lf\nx" must describe its own element
  left: Some("#document")   right: Some("P")
```

Commands run:

```
cargo nextest run --features render -p obscura-cdp --no-fail-fast    160 passed, 3 skipped
cargo nextest run -p obscura-cdp --no-default-features --no-fail-fast  129 passed, 3 skipped
cargo build --release -p obscura-cli --bins --features render         clean
```

Obstacle course, `--runs 1 --warmup 0`, base and candidate on the same host: **29/33 both**, the same four stages failing either way (`observer-intersection`, `textdecoder`, `fingerprint`, `charset-shiftjis`). Those four are a Windows-host artefact rather than anything this change touches, and the point of running both binaries was to show the delta is zero. I have not reproduced 33/33 locally, so if the Linux CI run disagrees I would rather see its output than argue from mine.

## Rendering

Not applicable.

## Performance

No expected impact. The change is one `serde_json::to_string` on an objectId in place of two `String::replace` allocations, on the `getProperties` / `describeNode` / `resolveNode` paths only. It runs once per call, not per property, and the replaced code also allocated.

## Checklist

Fixes #709.

### DIFF
--- a/crates/obscura-cdp/src/domains/dom.rs
+++ b/crates/obscura-cdp/src/domains/dom.rs
@@ -4,17 +4,6 @@ use serde_json::{json, Value};
 
 use crate::dispatch::CdpContext;
 
-/// Escape a client-supplied objectId for interpolation into a single-quoted JS
-/// string literal (`__obscura_objects['<here>']`). Backslashes must be escaped
-/// before single quotes, otherwise an id ending in `\` turns the closing quote
-/// into `\'` and produces a syntax error. All three objectId lookup sites route
-/// through this so they cannot diverge; not an injection vector (every `'` stays
-/// escaped, so the worst case is an unterminated string -> clean resolution
-/// failure), but a robustness fix.
-fn escape_object_id(oid: &str) -> String {
-    oid.replace('\\', "\\\\").replace('\'', "\\'")
-}
-
 /// Resolve a DOM `nodeId` from CDP params. Honors `nodeId`, `backendNodeId`,
 /// and `objectId` in that order. Playwright commonly passes only `objectId`
 /// (returned by a prior `DOM.resolveNode`); without this fallback those
@@ -28,9 +17,9 @@ fn resolve_node_id(page: &mut Page, params: &Value) -> Result<u64, String> {
     }
     if let Some(oid) = params.get("objectId").and_then(|v| v.as_str()) {
         let code = format!(
-            "(function() {{ var o = globalThis.__obscura_objects && globalThis.__obscura_objects['{}']; \
+            "(function() {{ var o = globalThis.__obscura_objects && globalThis.__obscura_objects[{}]; \
              return (o && typeof o._nid === 'number') ? o._nid : -1; }})()",
-            escape_object_id(oid)
+            crate::util::object_id_literal(oid)
         );
         let result = page.evaluate(&code);
         let nid = result.as_f64().map(|n| n as i64).unwrap_or(-1);
@@ -151,10 +140,9 @@ pub async fn handle(
             {
                 nid
             } else if let Some(oid) = params.get("objectId").and_then(|v| v.as_str()) {
-                let escaped_oid = escape_object_id(oid);
                 let code = format!(
-                    "(function() {{ var o = globalThis.__obscura_objects['{}']; if (!o) return -1; return (typeof o._nid === 'number') ? o._nid : -1; }})()",
-                    escaped_oid
+                    "(function() {{ var o = globalThis.__obscura_objects[{}]; if (!o) return -1; return (typeof o._nid === 'number') ? o._nid : -1; }})()",
+                    crate::util::object_id_literal(oid)
                 );
                 let result = page.evaluate(&code);
                 result.as_f64().map(|n| n as u64).unwrap_or(0)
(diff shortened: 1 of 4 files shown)


SGavrl (approved):

yinnho:
Both bycatch findings applied verbatim to our lineage (diting, in aginxbrowser) — thanks for extending the trail past the original note.

- `escape_object_id` in our `dom.rs` was the same two `replace` calls you found upstream. The page-minted-id observation is the part that makes this a correctness class rather than an escaping nit: once the id is `"parent::" + <page property name>`, no client-side discipline matters, so the only sound fix is making the interpolation survive arbitrary characters. We switched every dom.rs lookup site to the JSON double-quoted literal (same as the Runtime.getProperties fix), commit [0a1f366](https://github.com/yinnho/aginxbrowser/commit/0a1f366).

- `DOM.describeNode` (and `resolveNode`) had the `unwrap_or(0)` stale-handle lie too — a gone handle came back described as `#document`. Both now route through a resolver that turns the probe's -1 into a CDP error, matching Chrome. Verified on the wire: a newline-keyed handle resolves to `BODY`, a stale one returns `objectId … could not be resolved to a node`.

The `describeNode` fallback being *worse than an empty list* is a nice phrasing worth keeping in mind for any `unwrap_or(0)` on a node path — it converts a failure into a confident wrong answer at exactly the layer (handle → node) where clients stop double-checking.

PR #727, opened 2026-08-26
fix(render): give flattened inline wrappers real DOM geometry

### LINKED ISSUE #722: getBoundingClientRect returns 0x0 for inline elements in a run that cannot fold (replaced, inline-block, block, float or out-of-flow sibling)
### Before you start

### obscura version or commit

### OS and architecture

Linux x86_64

CDP server with Puppeteer or Playwright (also reproduces via CLI `--eval`)

### Affected area

JavaScript or DOM behavior

### Build variant

### Stealth mode

Off

### Reproduction

```
obscura fetch 'data:text/html,<div><label id=l>toggle</label><input type=checkbox></div>' --eval "(function(){ var r = document.getElementById('l').getBoundingClientRect(); return r.width + 'x' + r.height; })()"
```

The trigger is the sibling, not the element: `<div><label>toggle</label></div>` measures correctly. Any atomic inline in the same run reproduces it (`input`, `button`, `img`, `textarea`, a text sibling plus an input, `span` instead of `label`); all verified 0x0 against Chrome on a 10-case matrix.

### Expected behavior

About `48x17` (headless Chrome 145). The label should be a coordinate-click hit target.

### Actual behavior

`0x0`, so Puppeteer/Playwright `handle.click()` fails with "not clickable". Root cause as far as I can tell: a run containing a replaced element fails `try_build_run` and takes the mixed-run path in `build_mixed_block`, where `is_flattenable_inline` splices the wrapper's children directly into the parent's Taffy children. The wrapper never gets a layout box, so nothing records a rect for it; `synthesize_ordinary_inline_fragments` skips it because it requires an existing `rects` entry. ~~Painting is unaffected; this is DOM geometry only.~~ **Correction:** that is wrong. Flattening also drops the wrapper's own horizontal margin and padding from layout, so the following content is painted in the wrong place. The trigger set is also wider than a replaced sibling, and there is a second code path with the same symptom. See the follow-up comment below for the measurements: https://github.com/h4ckf0r0day/obscura/issues/722#issuecomment-5601628752

Combined with #721 this makes the common `<label>text</label><input>` checkbox pattern fully inoperable: the label has no activation behaviour and no hit-target either.

No. Chrome reports the real fragment geometry.

### Additional context

I have a fix that synthesizes the flattened wrapper's rect after layout as the union of its rendered descendants' element rects and per-word text runs, excluding positioned inlines (a positioned inline is a containing block for absolute descendants; installing a rect changes how those resolve, and broke `positioned_counter_inside_nested_inline_code_uses_the_code_gutter` until excluded): https://github.com/Vrejf/obscura/commit/8285e0b

The same pass has to avoid three traps, all verified against Chrome: an
out-of-flow descendant must not stretch the union (a `position:absolute` child at
`left:900px` otherwise turns a 28px span into a 918px one), fragments must be
merged per line rather than per word or `getClientRects()` returns one rect per
word, and a padded inline should keep a real box instead of being flattened at all
(`is_flattenable_inline` excludes borders and backgrounds today but not padding).
The descendant walk is depth-capped, matching the `descendants()` guard in
`tree.rs`.

Verification: the geometry matrices match Chrome within font-metric noise;
`cargo nextest run --release --features render --no-fail-fast` 1493/1493; the
no-render build and its tests (449/449); obstacle course unchanged at 32/33 with
the one failure also failing on stock; all 69 `render-repros` fixtures produce
metrics byte-identical to the pre-change build, so nothing repaints differently;
interleaved perf runs show no regression (58 vs 55 ms median on a fixture,
1358 vs 1351 ms on a synthetic page with 3000 flattened wrappers).

Known residuals I left alone rather than install guessed geometry: an empty inline
reports `0x0` where Chrome reports `0x<line height>`, and an inline wrapping only a
replaced element gets the replaced element's height rather than the line box's
(`8x8` vs Chrome's `8x16`). Both need the inline strut, which the existing
`synthesize_ordinary_inline_fragments` path derives from the text engine.

Happy to open a PR.

Vrejf:
Fixes #722.

A plain inline wrapper whose run holds a replaced element (`<label>x</label><input>`, `<span>x</span>`) fails `try_build_run`, so `build_mixed_block` takes the mixed-run path and `is_flattenable_inline` splices the wrapper's children into the parent's taffy children. The wrapper owns no box, nothing records a rect for it, and `getBoundingClientRect()` returned `0x0`, so Puppeteer/Playwright `handle.click()` failed with "not clickable".

The rect is now synthesized after layout from the wrapper's rendered descendants, grouped per line. Out-of-flow and floated descendants are excluded, since they are not part of the inline box; without that an absolutely positioned child at `left:900px` turned a 28px span into a 918px one. Positioned inlines keep their own box because they are a containing block for absolute descendants. A padded inline now keeps a real box instead of being flattened, matching how a bordered inline is already handled.

The walk is iterative and bounded by a node budget rather than by depth, and wrappers are resolved deepest-first so each subtree is visited once and a nested wrapper contributes its own line fragments rather than a collapsed union.

## Verification

Measured against headless Chrome 145 unless stated.

| probe | Chrome | before | after |
|---|---|---|---|
| `<div><label>toggle</label><input></div>` | 48x17 | 0x0 | 47x18 |
| absolute descendant inside the wrapper | 28px wide | 918px | 30px |
| wrapped inline, `getClientRects()` | 2 | 4 (one per word) | 2 |
| padded inline width | 43px | 27px | 47px |
| 70 nested inline wrappers, outermost | 39x19 | 0x0 | 38x18 |

- `cargo nextest run --release --features render --no-fail-fast`: 1488 run, 1486 passed. The two failures are `intersection_observer_*` tests that fail on unmodified `main` as well, at 2/10 and 3/10 over ten runs each on both trees.
- Obstacle course: 32/33, failing `observer-intersection` identically on `main`.
- All 69 `render-repros` fixtures produce metrics byte-identical to the pre-change build, so nothing repaints differently.
- Interleaved performance, same build and fixtures: 58 vs 55 ms median on a fixture, 1358 vs 1351 ms on a synthetic page with 3000 flattened wrappers. Inside the noise floor.

No `render-repros` fixture is included: the change is DOM geometry and produces no pixel difference, so a screenshot fixture would assert nothing. Coverage is in `layout_test.rs`.

## Known limitations

- An inline wrapping a block child reports one client rect where Chromium reports one per fragment. The bounding box matches.
- An empty inline reports `0x0` where Chromium reports `0x<line height>`, and an inline wrapping only a replaced element gets that element's height rather than the line box's. Both need the inline strut that `synthesize_ordinary_inline_fragments` derives from the text engine.

### DIFF
--- a/crates/obscura-render/src/dom.rs
+++ b/crates/obscura-render/src/dom.rs
@@ -7088,6 +7088,8 @@ fn layout_dom_once(
         }
     }
 
+    synthesize_flattened_inline_rects(tree, &mut rects, &text_runs, &styles, &mut inline_fragments);
+
     let generated_boxes = ifc_items
         .generated
         .iter()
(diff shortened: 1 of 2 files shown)


Vrejf:
Rebased onto main. The only conflict was an adjacency one in `layout_test.rs`, where main's new `deeply_nested_wrappers_cascade_without_stack_overflow` and this PR's tests were both appended; both are kept and both pass.

Vrejf:
Pushed one more commit and re-measured the whole thing against headless Chrome 149 on a 94-case fixture (one probe element per isolated container; `getBoundingClientRect`, `getClientRects().length`, `offsetWidth`/`offsetHeight`, and `document.elementFromPoint` at the box centre, compared per container so one case's height cannot cascade into the next).

**New commit: margin joins padding as a reason not to flatten.**

`is_flattenable_inline` excluded border and background but not the wrapper's own horizontal edges, so flattening dropped them from *layout*, not just from DOM geometry — the content after the wrapper was painted in the wrong place. Measuring `left` of the element following the wrapper:

| markup | Chrome | before | after |
| --- | --- | --- | --- |
| `<span style="margin-left:50px">A</span><input>` | 60 | 10 | 60 |
| `<span style="margin-right:50px">A</span><input>` | 60 | 10 | 60 |
| `<span style="padding-left:50px">A</span><input>` | 60 | 10 | 60 |
| `<span><span style="margin-left:50px">A</span></span><input>` | 60 | 10 | 60 |
| `<span style="margin-left:50px">A</span><b>B</b>` (not flattened) | 60 | 60 | 60 |

**Coverage on the 94-case matrix**

| | matches Chrome | 0x0 | geometry differs | rect count differs |
| --- | --- | --- | --- | --- |
| `main` (727cc46) | 9 | 85 | 0 | 0 |
| this branch | 66 | 11 | 13 | 4 |

The 9 that already worked on `main` are the controls with no trigger.

**What this branch does not fix, deliberately**

*A second code path with the same symptom (the 11 remaining 0x0).* `<div><span id=t>toggle</span><div>blk</div></div>` is still 0x0. `flatten_boxless_inline_children` removes the wrapper from the child list before `build_mixed_block` segments it; if the resulting run then folds, `collect_node_spans` never sees the wrapper, no `InlineOwnerBox` is recorded, and there is no per-node `text_runs` entry either — so this pass has nothing to reconstruct from. That also covers `<iframe>`, `<canvas>`, `<svg>`, `<video>`, `<object>` and `<embed>` siblings, which obscura lays out as block-level and which therefore land on the same path. Fixing it means changing how the run is built, not how geometry is recovered afterwards, so it belongs in its own change. Details and instrumented output in #722.

*Empty and whitespace-only wrappers.* `<span></span>`, `<span> </span>` and `<span>&nbsp;</span>` stay 0x0 where Chrome reports `0x<line height>` / `10x19`. These need the inline strut, which the existing `synthesize_ordinary_inline_fragments` path derives from shaping provenance this pass does not have.

*Approximations in the synthesized box, all over-estimates rather than under-estimates, so centre-point hit testing still lands on the content:*

| case | Chrome | this branch |
| --- | --- | --- |
| `<span style="line-height:40px">` — height comes from the line box, not the font content area | 58x19 | 58x40 |
| inline wrapping only a replaced element — replaced height, not line height | 8x19 | 8x8 |
| `<label><input type=checkbox> text</label>` — an atomic child's margins are outside the union | 97x19 | 83x19 |
| `getClientRects()` on a `<br>`-split or partly relatively-offset inline — merged per line, so fewer fragments | n=3..5 | n=2 |

The remaining differences in the matrix are pre-existing and unrelated to this change: `<small>` does not apply its font size, `<q>` does not generate UA quotes, `transform` is applied to a non-replaced inline (CSS says it should not be), text next to a taller atomic inline is top-aligned rather than baseline-aligned, and `text-align:right` places the line 8px off.

**Verification**

- `cargo nextest run --release --features render --no-fail-fast`: 1649 passed, 4 skipped
- `cargo nextest run -p obscura-render --features paint`: 590 passed, 1 skipped
- All 64 `render-repros` fixtures render byte-identically before and after the margin commit, so nothing repaints differently.
- Geometry survives incremental relayout: mutating `textContent`, inserting a sibling, resizing the viewport, and scrolling all keep the synthesized rect in step with Chrome, and `scrollIntoView` plus a real Puppeteer `click()` on a below-the-fold label both work.

PR #728, opened 2026-08-26
fix(js,cdp): implement label activation behaviour

### LINKED ISSUE #721: Label activation is not implemented: clicking a <label> never toggles its control (element.click() and CDP)
### Before you start

- [x] I searched existing issues and this is not a duplicate. (#676 described this but was closed by its author as unreproduced; it does reproduce, details below.)

### obscura version or commit

### OS and architecture

Linux x86_64

CDP server with Puppeteer or Playwright (also reproduces via CLI `--eval`)

### Affected area

JavaScript or DOM behavior

### Build variant

### Stealth mode

Off

### Reproduction

```
obscura fetch 'data:text/html,<label for=c>toggle</label><input type=checkbox id=c>' --eval "(function(){ document.querySelector('label').click(); return document.getElementById('c').checked; })()"
```

Same result when the click is delivered as CDP `Input.dispatchMouseEvent` at the label's coordinates (puppeteer-core `handle.click()`).

### Expected behavior

`true`. Per the HTML spec, activating a label runs a synthetic click on its labeled control (the `for`-referenced element, else the first labelable descendant).

### Actual behavior

`false`. Neither `HTMLElement.click()` in `bootstrap.js` nor the click path in `obscura-cdp/src/domains/input.rs` implements label activation. Practical impact: checkbox/radio groups built as `<label for=...>` (Webflow's standard checkbox markup, most styled-checkbox patterns) cannot be operated by automation at all.

Edge cases that need to hold (all verified against headless Chrome 145):

- `for=""` present but empty associates nothing; it must not fall back to a nested control
- dangling `for` activates nothing
- `for` wins over a nested control when both exist
- a disabled control has no activation behaviour
- a click on a deep descendant of the label activates via the nearest label ancestor
- clicking a nested control directly toggles once, no double-fire through the label
- a control's click handler that clicks its own label must not re-toggle (Chrome guards this recursion)
- checkbox/radio pre-click activation: state flips before the click event and reverts if the event is cancelled; `input` and `change` fire after

No. Chrome returns `true` and passes all edge cases above.

### Additional context

Two traps worth naming, because adding activation behaviour introduces them if
they are not handled. Neither is broken in v0.2.1 today: with no activation
behaviour at all, a disabled control cannot be toggled and `indeterminate` is
never touched. Both regressed the moment I implemented activation, and I only
caught them by differential testing:

- a disabled control has no activation behaviour and dispatches no click event at
  all, including when disabled by an ancestor `<fieldset>` (its first `<legend>`
  exempt). Adding pre-click activation without this check makes
  `disabledCheckbox.click()` toggle, which stock correctly does not do.
- pre-click activation is also specified to clear `indeterminate` and restore it
  when the event is cancelled. I left that out: `HTMLInputElement.indeterminate`
  is not implemented at all today (`'indeterminate' in checkbox` is `false`), so
  setting it from the activation steps would only create an enumerable expando on
  the element rather than drive real state. It belongs with implementing the
  property, not here.

I have a fix covering all of the above through both `element.click()` and CDP
mouse events (16 cases x 2 dispatch modes x 2 DOM shapes, exact match with Chrome;
stock fails 11-12 of 16), plus six Rust regression tests that fail against
pre-fix sources: https://github.com/Vrejf/obscura/commits/fix/label-activation

Verification: `cargo nextest run --release --features render --no-fail-fast` 1493/1493; the no-render build and its tests (449/449); obstacle course unchanged
at 32/33, with the one failure (`observer-intersection`, `expected 'io:50', got ''`)
failing identically on stock 0.2.1; all 69 `render-repros` fixtures produce metrics byte-identical to the
pre-change build; interleaved perf runs show no regression (58 vs 55 ms median on
a fixture, 1358 vs 1351 ms on a synthetic 3000-wrapper page, both inside the noise
floor).

Known gaps I did not attempt, since they are pre-existing and wider than this bug:
`element.form` ignores the `form="id"` attribute so radio grouping can be wrong
across form-attribute association; `ownerDocument` always resolves to the global
document, so `for` association is not shadow-tree scoped; form-associated custom
elements are not in the labelable set; and a label activation delivered by CDP
produces an untrusted click on the control, where Chrome reports it as trusted.

Happy to open a PR if the approach looks right.

Vrejf:
Fixes #721.

Activating a `<label>` runs a synthetic click on its labeled control: the element referenced by `for`, else the first labelable descendant. Neither `HTMLElement.click()` nor the CDP click path did this, so a checkbox or radio built as a styled label could not be operated at all. That is the whole of Webflow's `.w-checkbox` markup and most styled-checkbox patterns.

Checkbox and radio pre-click activation runs with it, since without it a click event dispatched but nothing ever toggled: state flips before the click event so listeners observe the new value, reverts if the event is cancelled, and `input` then `change` fire after an uncancelled toggle.

The association and activation rules that go with it, each verified against headless Chrome: an empty `for` associates nothing rather than falling back to a descendant; a dangling `for` activates nothing; `for` wins over a nested control; a disabled control has no activation behaviour, including under a disabled `<fieldset>` outside its first `<legend>` child; a click on interactive content inside a label keeps that content's own behaviour, while `meter`, `output` and `progress` are labelable but not interactive and still forward; a control clicking its own label cannot bounce back.

The CDP path calls shared helpers for association, disabled state, and interactive content rather than restating those rules, so the two click paths cannot drift apart on them. Forwarded events inherit the trustedness of the click that caused them, so a physical label click produces a trusted control click as it does in a real browser.

## Verification

16 spec edge cases, each through both `element.click()` and CDP `Input.dispatchMouseEvent`, on two DOM shapes: 64/64 exact match with headless Chrome 145. Stock fails 11-12 of 16.

Cases that were wrong in my own first attempt and are now covered by tests, all measured against Chrome:

| probe | Chrome | naive implementation | now |
|---|---|---|---|
| `disabledCheckbox.click()` | no toggle, no event | toggles | no toggle, no event |
| `<div disabled>.click()` | dispatches | suppressed | dispatches |
| `<a>` inside `<fieldset disabled>` | dispatches | suppressed | dispatches |
| click `<a href>` inside a label | no activation | activates | no activation |
| click `<output>` inside a label | activates | no activation | activates |
| `<legend>` wrapped in a `<div>` | disabled | enabled | disabled |
| inner fieldset legend, outer fieldset disabled | disabled | enabled | disabled |
| CDP click on a disabled control | no toggle, no event | toggles | no toggle, no event |

- `cargo nextest run --release --features render --no-fail-fast`: 1497 run, 1494 passed. All three failures pass individually and also fail on unmodified `main`; they are timing-sensitive `intersection_observer` and budget tests, measured at 2-3/10 on both trees.
- Obstacle course: 32/33, failing `observer-intersection` identically on `main`.
- New coverage: six runtime tests and an end-to-end CDP test (`input_mouse_label_activation`) asserting event type, order, count and `isTrusted` for explicit, implicit and deep-descendant labels, plus a disabled control.
- On a real page built with this markup, three option clicks now produce the same selected state as Chrome (0 to 3), where stock produces none.

## Notes for review

- `HTMLInputElement.indeterminate` is not implemented in obscura (`'indeterminate' in checkbox` is `false`), so the pre-click step that clears and restores it is deliberately omitted rather than faked with an expando. It belongs with implementing the property.
- `__obscura_activateLabel` can be called by page script to request a trusted forwarded click. The token itself is closure-private, but the helper supplies it on request, exactly as the existing `__obscura_markTrusted` does for any event. No new capability, but worth knowing.
- The forwarded control event carries trustedness but not the originating pointer's coordinates, `detail` or modifiers.
- A bare inline `<label>` next to an `<input>` has no box until #722 lands, so coordinate clicks on that shape need this and #722 together. Labels with their own box, which is the common styled case, work with this change alone.

### DIFF
--- a/crates/obscura-cdp/src/domains/input.rs
+++ b/crates/obscura-cdp/src/domains/input.rs
@@ -157,6 +157,7 @@ pub async fn handle(
                             if (!clickTarget) return;\
                             var tag = clickTarget.tagName;\
                             var type = (clickTarget.getAttribute && clickTarget.getAttribute('type') || '').toLowerCase();\
+                            if (globalThis.__obscura_isDisabled(clickTarget)) return;\
                             var checkable = tag === 'INPUT' && (type === 'checkbox' || type === 'radio');\
                             var oldChecked = checkable ? !!clickTarget.checked : false;\
                             var radioStates = null;\
@@ -188,6 +189,12 @@ pub async fn handle(
                                 try {{ clickTarget.dispatchEvent(globalThis.__obscura_markTrusted(new Event('input', {{bubbles:true}}))); }} catch(e) {{}}\
                                 try {{ clickTarget.dispatchEvent(globalThis.__obscura_markTrusted(new Event('change', {{bubbles:true}}))); }} catch(e) {{}}\
                                 return;\
+                            }}\
+                            var labelHost = tag === 'LABEL' ? clickTarget : (clickTarget.closest ? clickTarget.closest('label') : null);\
+                            var interactiveHost = globalThis.__obscura_interactiveHost(clickTarget);\
+                            if (labelHost && !(interactiveHost && labelHost.contains(interactiveHost))) {{\
+                                var ctl = globalThis.__obscura_labeledControl(labelHost);\
+                                if (ctl && ctl !== clickTarget && globalThis.__obscura_activateLabel(labelHost, ctl, true)) {{ return; }}\
                             }}\
                             var link = clickTarget.closest ? clickTarget.closest('a[href]') : null;\
                             if (!link && tag === 'A' && clickTarget.getAttribute('href')) link = clickTarget;\
--- a/crates/obscura-cdp/tests/input_mouse_label_activation.rs
+++ b/crates/obscura-cdp/tests/input_mouse_label_activation.rs
@@ -1,0 +1,158 @@
+#![cfg(feature = "render")]
+
+use obscura_cdp::dispatch::{dispatch, CdpContext};
+use obscura_cdp::types::CdpRequest;
+use serde_json::{json, Value};
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::TcpListener;
+
+async fn serve_fixture() -> String {
+    let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let addr = listener.local_addr().unwrap();
+    tokio::spawn(async move {
+        let (mut socket, _) = listener.accept().await.unwrap();
+        let mut buf = [0u8; 2048];
+        let _ = socket.read(&mut buf).await.unwrap();
+        // The labels carry their own boxes so a coordinate click has a target
+        // regardless of inline-wrapper geometry.
+        let body = r#"<!doctype html><html><head><style>
+            html, body { margin: 0; font: 16px monospace }
+            label { display: block; width: 200px; height: 40px }
+        </style></head><body>
+          <label id="explicit" for="boxa">a</label><input id="boxa" type="checkbox">
+          <label id="implicit">b <input id="boxb" type="checkbox"></label>
+          <label id="deep" for="boxc"><span><b id="deep-text">c</b></span></label>
+          <input id="boxc" type="checkbox">
+          <label id="off" for="boxd">d</label><input id="boxd" type="checkbox" disabled>
+          <script>
+            window.events = [];
+            for (const id of ['boxa','boxb','boxc','boxd']) {
+              const el = document.getElementById(id);
+              for (const t of ['click','input','change']) {
+                el.addEventListener(t, e => window.events.push(id + ':' + t + ':' + e.isTrusted));
+              }
+            }
+          </script>
+        </body></html>"#;
+        let response = format!(
+            "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\nContent-Length: {}\r\nConnection: close\r\n\r\n{body}",
+            body.len()
+        );
+        let _ = socket.write_all(response.as_bytes()).await;
+    });
+    format!("http://{addr}/")
+}
+
+async fn cdp(ctx: &mut CdpContext, id: u64, method: &str, params: Value, sid: &str) -> Value {
+    let response = dispatch(
+        &CdpRequest {
+            id,
+            method: method.to_string(),
+            params,
+            session_id: Some(sid.to_string()),
+        },
+        ctx,
+    )
+    .await;
+    assert!(response.error.is_none(), "CDP {method} failed: {:?}", response.error);
+    response.result.unwrap_or_else(|| json!({}))
+}
+
+async fn evaluate(ctx: &mut CdpContext, id: u64, expression: &str, sid: &str) -> Value {
+    cdp(
+        ctx,
+        id,
+        "Runtime.evaluate",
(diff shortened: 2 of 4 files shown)


yinnho:
Reviewed the three PRs together (#727/#728/#729) — the two-level guard composes correctly, matching what we traced through the expanded matrix on our build.

**Own-label bounce, implicit label** (checkbox C inside label L, C's handler calls L.click()):
1. `L.click()`: L not in `_clickInProgress` → dispatch L → forwarding: `_labeledControl(L)`=C, `_forwardingLabels.has(L)`=false → add L → `C.click()`
2. `C.click()`: C not in `_clickInProgress` → flip C → dispatch C → handler calls `L.click()`
3. `L.click()`: L still not in `_clickInProgress` (label-level set is separate) → dispatch L → forwarding: `_forwardingLabels.has(L)`=**true** → skip → returns

So the element-level WeakSet (#729) stops direct self-reentry, and the closure-private `_forwardingLabels` stops the label→control bounce — the control flips exactly once, and a cancelled dispatch restores via the existing `_oldChecked` path. The element-level flag sitting above the pre-activation flip (not below it) is what keeps state from being stuck post-flip on the suppressed entry.

One residual difference on our side: we went ahead and implemented `indeterminate` as a real IDL property (prototype getter/setter over a per-element store, since the checkbox activation path clears it — see https://github.com/yinnho/aginxbrowser/commit/88b8496). Obscura has no such property so there's nothing to clear here; flagging only so the option is on the table if `'indeterminate' in checkbox` ever matters for a page.

Vrejf:
Thanks for tracing it. I merged all three locally and the composed behaviour is what you describe, but one step of the mechanism is not:

At step 3 `L` **is** already in `_clickInProgress`. Step 1 added it before dispatching and only removes it in `finally`, so the nested `L.click()` is suppressed by the element-level flag, not by `_forwardingLabels`. The two sets are separate, but they are keyed on different elements in the same call stack rather than being independent of each other.

`_forwardingLabels` is still load-bearing, just on the other path. A CDP click never calls `L.click()`: it dispatches its own trusted event and calls `__obscura_activateLabel` directly, so `_clickInProgress` never holds the label and the label-level set is the only thing stopping the bounce. Measured on the composed tree, both paths match headless Chrome:

| path | Chrome | composed |
|---|---|---|
| `L.click()`, implicit label, control handler clicks `L` | checked=true, change=1, control dispatches=1 | same |
| CDP mouse click on `L`, same handler | checked=true, change=1, control dispatches=1 | same |

So both guards are needed, and the ordering point in #729 holds: the element flag has to sit above the pre-activation flip. The merge is not clean, incidentally: #728 and #729 both edit `click()` and conflict, and resolving it by taking either side alone silently drops one of the two guards.

On `indeterminate`, agreed that implementing the IDL property is the right call. I left it out of #728 deliberately rather than write the activation step against a property that does not exist, since setting it would only create an expando. Worth its own issue upstream.

yinnho:
You're right — my step-3 claim was wrong. `L` was added to `_clickInProgress` at step 1 and only removed in `finally`, so the JS-path bounce is stopped by the element-level flag, not the label set. The label-level `_forwardingLabels` is what carries the CDP path (which never re-enters `L.click()`), so the two guards cover different call paths — that's the cleaner way to put it, and it matches your measured table. Good catch on the merge conflict too: taking either #728 or #729 alone silently drops one guard.

On `indeterminate` — opening the upstream issue now.

SGavrl (approved):

PR #729, opened 2026-08-26
fix(js): honor the click in progress flag

### LINKED ISSUE #726: HTMLElement.click() does not honor the click in progress flag
### Before you start

- [x] I searched existing issues and this is not a duplicate. (The closed stack-overflow issues are deep-DOM recursive traversal; this is re-entrant event dispatch.)

### obscura version or commit

### OS and architecture

Linux x86_64

### Affected area

JavaScript or DOM behavior

### Build variant

### Stealth mode

Off

### Reproduction

```
obscura fetch 'data:text/html,<button id=x>b</button>' --eval "(function(){ var n=0; var x=document.getElementById('x'); x.onclick=function(){ n++; if(n<5000) x.click(); }; x.click(); return n; })()"
```

`5000` is above the stack limit observed here, so the repro still reaches the swallowed `RangeError`, and it also caps execution on a build with a larger stack. Without the guard, recursion continues until stack exhaustion.

### Expected behavior

`1`. Per HTML, `click()` first returns for a disabled form control, then returns if the element's click in progress flag is set. Otherwise it sets the flag, fires a synthetic pointer event named `click`, and unsets the flag, so a handler that clicks its own element cannot re-enter. Headless Chrome returns `1`.

### Actual behavior

The command returned `2017` on this host on both v0.2.1 and `main`. The depth is stack-size dependent; without the guard, dispatch continues until stack exhaustion.

`click()` has no click in progress flag, so each dispatch re-enters until the stack runs out. The resulting `RangeError` is raised inside the listener, and `dispatchEvent` wraps every listener in `try { ... } catch(e) { console.error(e) }` (`bootstrap.js`, the inline-handler and registered-handler loops), so it is logged to stderr and does not propagate. The caller sees `click()` return normally: a `try`/`catch` around the outer `click()` catches nothing, and no `error` event or `window.onerror` fires. So one `click()` recurses to stack exhaustion, and the overflow surfaces only as a console trace.

No. Chrome dispatches once.

### Additional context

Proposed fix and regression test: https://github.com/Vrejf/obscura/commit/495fadb

A `WeakSet` of elements with a click in progress, checked and set around the dispatch in `try`/`finally`. Behaviour verified against headless Chrome on the cases the guard could plausibly break:

| case | Chrome | stock | fixed |
|---|---|---|---|
| handler clicks its own element | 1 | to stack exhaustion | 1 |
| three sequential clicks | 3 | 3 | 3 |
| flag clears after a listener throws | 2 | 2 | 2 |
| handler clicks a different element | fires both | fires both | fires both |
| mutual recursion, A clicks B clicks A | 2 | to stack exhaustion | 2 |

Also `cargo nextest run --release --features render --no-fail-fast` at 1493/1493. The obstacle course scores 32/33 on both the base and fixed commits, failing `observer-intersection` identically (`expected 'io:50', got ''`), so it is unrelated to this change.

The branch also carries fixes for #721 and #722. This commit touches only `bootstrap.js` and its test, so it is independent of both.

Vrejf:
Fixes #726.

`HTMLElement.click()` did not implement the HTML click in progress flag, so a handler that clicks its own element re-entered until the V8 stack was exhausted. `dispatchEvent` wraps every listener in `try { ... } catch(e) { console.error(e) }`, so the resulting `RangeError` was logged and did not propagate: the outer `click()` returned normally and the overflow surfaced only as a console trace.

```
obscura fetch 'data:text/html,<button id=x>b</button>' --eval "(function(){ var n=0; var x=document.getElementById('x'); x.onclick=function(){ n++; if(n<5000) x.click(); }; x.click(); return n; })()"
```

Returned `2017` on this host on v0.2.1 and `main`; the depth is stack-size dependent. Chrome returns `1`.

The flag is tracked per element in a closure-private `WeakSet` and cleared in `finally`, so it survives a listener that throws and adds nothing page-observable.

## Verification

| probe | Chrome | before | after |
|---|---|---|---|
| handler clicks its own element | 1 | to stack exhaustion | 1 |
| three sequential clicks | 3 | 3 | 3 |
| cycle: A clicks B clicks A | 2 | to stack exhaustion | 2 |
| flag clears after a listener throws | 2 | 2 | 2 |
| handler clicks a different element | fires both | fires both | fires both |

- `cargo nextest run --release --features render --no-fail-fast`: 1486 run, 1485 passed. The single failure is a timing-sensitive test that also fails on unmodified `main`.
- Obstacle course unchanged at 32/33, failing `observer-intersection` identically on `main`.

## Note on merge order

This is independent of #721 and #722 and applies to `main` on its own. If it merges alongside #721, the flag check must stay after the disabled early return and before pre-click activation, otherwise a suppressed re-entrant call still flips `checked` before it returns. That ordering is not observable on `main` alone, because pre-click activation does not exist yet.

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -2576,6 +2576,10 @@ function _htmlAttrName(el, n) {
 // A submit button per the HTML spec: a <button> whose type is submit — the
 // default, including when the type attribute is missing or invalid — or an
 // <input> of type submit/image. Used to validate requestSubmit's submitter.
+// Elements with a click currently dispatching, so a handler cannot re-enter
+// click() on its own element.
+const _clickInProgress = new WeakSet();
+
 function _isSubmitButton(el) {
   if (!el || typeof el.localName !== "string") return false;
   const type = ((el.getAttribute && el.getAttribute("type")) || "").toLowerCase();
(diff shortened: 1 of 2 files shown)


Vrejf:
Merge note, since it is only visible from the other side: this and #728 both edit `click()` and conflict textually. Resolving by taking either side alone silently drops one of two guards that cover different call paths.

Verified on a locally merged tree: the element-level flag from this PR stops a handler that re-enters `click()` on its own element, while #728's label-level set is the only thing stopping the same bounce over CDP, because a CDP click never calls `label.click()` and so never puts the label in this PR's set. With both, each path matches headless Chrome (control toggles once, one `change`, one dispatch); with either alone, one path regresses.

If both are taken, the correct resolution keeps this PR's flag check after the disabled early return and before #728's pre-click activation. Happy to push a pre-resolved branch if that is easier than resolving at merge time.

Vrejf:
Rebased onto main now that #728 has landed. The conflict was the one flagged above: #728 rewrote `click()`, and resolving by taking either side alone would have dropped a guard.

Resolved by keeping main's `click()` and placing the flag check after the disabled early return and before pre-click activation. Verified on the rebased branch against headless Chrome:

| probe | Chrome | rebased |
|---|---|---|
| handler clicks its own element | 1 | 1 |
| three sequential clicks | 3 | 3 |
| label bounce: control handler clicks its own label | checked, 1 change | checked, 1 change |
| disabled control still inert | no toggle | no toggle |

`cargo nextest run --release --features render --no-fail-fast`: 1646 run, 1644 passed. The two failures are `intersection_observer` tests that also fail on unmodified main.

PR #741, opened 2026-08-29
fix(js): implement HTMLInputElement.indeterminate

### LINKED ISSUE #732: HTMLInputElement.indeterminate is missing (and checkbox activation can't clear it)
### Environment

- Surface: JS runtime (`HTMLElement` / checkbox)
- Build variant: independent of rendering

### Summary

`HTMLInputElement.indeterminate` is not implemented. `'indeterminate' in checkbox` is `false`, and there is no prototype getter/setter to back it. Beyond missing the property itself, the checkbox activation path can't clear it: per the HTML spec, a click on an indeterminate checkbox clears `indeterminate` before flipping `checked`, and any script that set it then clicked would leave the flag stuck.

### Repro

```js
const el = document.createElement('input');
el.type = 'checkbox';
'id' in el;            // true
'indeterminate' in el; // false — should be true
```

Activation interplay: an indeterminate checkbox clicked once should report `checked=false, indeterminate=false` (indeterminate cleared, state not flipped on the first click).

### Reference implementation

We hit this during the label-activation work and implemented it as a real IDL property backed by a per-element store (prototype getter/setter, so `'in'` and fresh elements behave, and the activation step clears the store): https://github.com/yinnho/aginxbrowser/commit/88b8496

ntdatt812:
Fixes #732.

## What changed

`indeterminate` was not implemented anywhere in the engine. `grep -r indeterminate crates/` returned nothing: no prototype accessor to back it, `'indeterminate' in checkbox` was `false`, and a script could neither set nor read the flag.

**`crates/obscura-js/js/bootstrap.js`**

* `get`/`set indeterminate` on the element prototype, backed by a node-keyed `_formIndeterminate` store. Node-keyed rather than per-instance for the same reason `checked` is: element wrappers are rebuilt on every lookup, so a field on the instance would not survive `getElementById` handing back a new wrapper. Being on the prototype is also what makes `'indeterminate' in el` true for a freshly created element, which is the check in the issue's repro.
* `Element#click()` clears `indeterminate` in the pre-click step and restores the old value when the click is canceled, alongside the existing `checked` handling.

**`crates/obscura-cdp/src/domains/input.rs`**

The pre-click activation logic is duplicated in the CDP mouse snippet, which is what real dispatched input goes through. Without the same three lines there, `Input.dispatchMouseEvent` on a checkbox would leave the flag stuck even once the property exists. This is the duplication the comment in `bootstrap.js` already points at.

Radio inputs are unaffected: HTML's pre-activation for radio sets checkedness to true and says nothing about indeterminateness, so the restore path there writes back the value it read.

<details>
<summary><b>One correction to the issue, which changes what the fix does</b></summary>

The issue also says an indeterminate checkbox clicked once should report `checked=false`, describing the state as "not flipped on the first click". That is not what HTML specifies. Legacy-pre-activation behavior is:

> If this element's `type` attribute is in the Checkbox state, then set this element's checkedness to its opposite value (i.e. true if it is false, false if it is true) **and** set this element's indeterminateness to false.

Clearing `indeterminate` is an extra step, not a replacement for the toggle. An unchecked indeterminate checkbox reports `checked=true, indeterminate=false` after one click, which is what Chrome and Firefox do, and it is what this PR implements.

I have gone with the spec rather than the linked reference implementation because the spec is the thing the rest of the file follows. If you would rather match the reference, it is two lines in each of the two places, and I will send it as a follow-up rather than hold this one.
</details>

## Validation

New test `test_checkbox_indeterminate_is_idl_only_and_cleared_by_activation` in `crates/obscura-js/src/runtime.rs`, covering the four things that were broken or at risk: `'indeterminate' in el` is true on a fresh element and reads `false`; setting it round-trips; clicking an indeterminate checkbox yields `checked=true, indeterminate=false`; a canceled click (`preventDefault`) restores both.

I ran the four-configuration sequence from CONTRIBUTING rather than only the crate I touched, since this changes shared DOM and CDP code. All six steps in a `rust:latest` container, so the result matches CI rather than my Windows host:

| Step | Result |
| --- | --- |
| `cargo build --release -p obscura-cli --bins --features render` | ok, 1m40s |

The focused test, for the log:

```
PASS [0.017s] obscura-js runtime::tests::test_checkbox_indeterminate_is_idl_only_and_cleared_by_activation
```

I have not run `cargo fmt` over the tree, and the diff is the three files the fix needs.

## Rendering

Not applicable. No layout, paint, screenshot, screencast or PDF code is touched, and the render-feature suite above is unchanged at 1507 passing.

## Performance

The two accessors are only reached when a page reads or writes `.indeterminate`, which nothing did before this PR. The cost added to the common path is one object write per activation of a checkable input, next to the `_formChecked` write already there. Nothing is added to layout, paint, or the network path.

## Checklist

### DIFF
--- a/crates/obscura-cdp/src/domains/input.rs
+++ b/crates/obscura-cdp/src/domains/input.rs
@@ -160,6 +160,7 @@ pub async fn handle(
                             if (globalThis.__obscura_isDisabled(clickTarget)) return;\
                             var checkable = tag === 'INPUT' && (type === 'checkbox' || type === 'radio');\
                             var oldChecked = checkable ? !!clickTarget.checked : false;\
+                            var oldIndeterminate = checkable ? !!clickTarget.indeterminate : false;\
                             var radioStates = null;\
                             if (checkable && type === 'radio') {{\
                                 var radioName = clickTarget.getAttribute('name') || '';\
@@ -176,13 +177,14 @@ pub async fn handle(
                                 clickTarget.checked = true;\
                             }} else if (checkable) {{\
                                 clickTarget.checked = !oldChecked;\
+                                clickTarget.indeterminate = false;\
                             }}\
                             var click = globalThis.__obscura_markTrusted(new MouseEvent('click', {{bubbles:true,cancelable:true,view:globalThis,clientX:{x},clientY:{y},button:0,buttons:0,detail:{click_count},altKey:{alt_key},ctrlKey:{ctrl_key},metaKey:{meta_key},shiftKey:{shift_key}}}));\
                             var cancelled = !clickTarget.dispatchEvent(click);\
                             if (cancelled) {{\
                                 if (radioStates) {{\
                                     for (var rr = 0; rr < radioStates.length; rr++) radioStates[rr][0].checked = radioStates[rr][1];\
-                                }} else if (checkable) clickTarget.checked = oldChecked;\
+                                }} else if (checkable) {{ clickTarget.checked = oldChecked; clickTarget.indeterminate = oldIndeterminate; }}\
                                 return;\
                             }}\
                             if (checkable && clickTarget.checked !== oldChecked) {{\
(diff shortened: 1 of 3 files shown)


ntdatt812:

| --- | --- | --- |
| this branch | 32/33 | 3016.1ms |

```
```

SGavrl (approved):

PR #742, opened 2026-08-29
fix(js): pin the ICU default locale to navigator.language

### LINKED ISSUE #734: Intl default locale is en-AU while navigator.language and Accept-Language are en-US
### Before you start

- [x] I searched existing issues and this is not a duplicate. The closest are #268 (OS-matched profiles) and #228 (configurable timezone and GPS), neither of which covers the language surfaces.
- [x] Reproduced on the v0.2.0 render+stealth binary and verified at source on `main` (`f449e6f`). I did not rebuild v0.2.1 to re-run it - see "Version" below for exactly what was checked where.

### obscura version or commit

v0.2.0, official render+stealth archive, is where the output below was measured. On `main` (`f449e6f`) the three sources that produce it are unchanged: `crates/obscura-js/js/bootstrap.js:6490-6491`, `crates/obscura-net/src/client.rs:1056`, and no default ICU locale set anywhere under `crates/`. So the same result is expected on v0.2.1, but I am reporting it as source-checked rather than re-measured.

### OS and architecture

Windows 10 x64.

CLI (`fetch`) and the CDP server with Playwright. Both show it.

### Affected area

JavaScript or DOM behavior.

### Build variant

Render-enabled with stealth. Reproduces with `--stealth` and without it - the repro below does not pass the flag.

### Reproduction

No network needed:

```
obscura fetch "data:text/html,<html><body>x</body></html>" --eval "JSON.stringify({intl: Intl.DateTimeFormat().resolvedOptions().locale, navLang: navigator.language, langs: navigator.languages, fmt: new Intl.DateTimeFormat().format(new Date(2026,0,31))})"
```

### Expected behavior

The locale surfaces a page can read agree with each other, the way they do in a real browser. The specific value matters less than the agreement.

### Actual behavior

```json
{"intl":"en-AU","navLang":"en-US","langs":["en-US","en"],"fmt":"31/01/2026"}
```

Every `Intl` constructor resolves to `en-AU`:

```
Intl.DateTimeFormat().resolvedOptions().locale  en-AU
Intl.NumberFormat().resolvedOptions().locale    en-AU
Intl.Collator().resolvedOptions().locale        en-AU
```

while `navigator.language` is `en-US`, `navigator.languages` is `["en-US", "en"]`, and the request carries `Accept-Language: en-US,en;q=0.9`. Two of the three say `en-US` and the third says `en-AU`.

The user-visible half is that `new Intl.DateTimeFormat().format(...)` is day-first, so a page that formats a date without an explicit locale renders `31/01/2026` where an `en-US` browser renders `1/31/2026`.

No. Stock headless Chromium 148.0.7778.96 on the same host, same page:

```json
{"intl":"en-GB","navLang":"en-GB","langs":["en-GB"],"fmt":"31/01/2026"}
```

Different value, but all three agree, which is the part that differs from obscura.

### Where it comes from

- `crates/obscura-js/js/bootstrap.js:6490-6491` hardcodes `navigator.language` to `"en-US"` and `navigator.languages` to `["en-US", "en"]`.
- `crates/obscura-net/src/client.rs:1056` defaults `Accept-Language` to `"en-US,en;q=0.9"`.
- Nothing under `crates/` sets a default ICU locale, so `Intl` falls back to whatever the bundled ICU in the V8 build resolves to. That is `en-AU` on this host.

I have not established whether `en-AU` is a constant of the V8/ICU build or is derived from the host in some way I did not find - this machine's system locale is Russian, so it is not simply inherited from the OS. If it turns out to be host-derived, the mismatch changes shape rather than going away, since the other two are pinned constants.

### Why it seems worth fixing

#268 argued that internally inconsistent identity surfaces are the detectable thing, and the stealth docs state the goal as "the surfaces a site fingerprints agree with each other". This is one line to check and it disagrees. `OBSCURA_TIMEZONE` already exists for the same class of problem on the time axis.

Direction is your call. The two that occurred to me: set the ICU default locale at V8 init from the same constant that feeds `navigator.language`, or drive all three from the profile so that the language axis gets an override the way `OBSCURA_TIMEZONE` gives the time axis one. I did not send a patch because I do not know which of those fits the project, and CONTRIBUTING asks for an issue first.

ntdatt812:
Fixes #734.

## What changed

Three surfaces name a locale, and only two of them were pinned:

* `navigator.language` is `"en-US"` and `navigator.languages` is `["en-US", "en"]`, both hardcoded in `js/bootstrap.js`.
* `obscura-net` sends `Accept-Language: en-US,en;q=0.9`.
* `Intl` resolves its default through ICU, and nothing under `crates/` ever set one. `grep -rn "set_default_locale" crates/` was empty.

So the third value was a property of whichever V8 build you happened to be running, not of the browser identity Obscura presents. The report measured `en-AU` from `Intl` against `en-US` from `navigator.language` on a Windows release build. A page that reads two of the three sees a browser disagreeing with itself, which is the kind of internal inconsistency identity checks look for, so this sits in the stealth priority the README states.

The disagreement being one-sided is what decides the fix: two surfaces are already pinned, one is inherited, so agreeing means pinning the third rather than loosening the other two. That also keeps what scripts read consistent with what actually goes out on the wire.

One constant and one call, in `crates/obscura-js/src/runtime.rs`:

* `DEFAULT_LOCALE` names the value and, in its doc comment, the two other places that have to stay equal to it.
* `pin_default_locale()` calls `v8::icu::set_default_locale` inside a `Once`.

It is called from `with_base_url_and_proxy` **before** `JsRuntime::new`, inside the existing `ISOLATE_CREATE_LOCK` block. Both parts matter: ICU is consulted when a page first constructs an `Intl` object, so an isolate built ahead of the call would answer from the host default, and the lock keeps the write from racing an isolate being constructed on another connection thread. That is the only `JsRuntime::new` site in the crate, so CLI and CDP both go through it.

<details>
<summary><b>A note on the constant, which does not need an answer before merging</b></summary>

`DEFAULT_LOCALE` duplicates a string that also lives in `bootstrap.js` and in `obscura-net`. I kept it that way because threading one constant across a Rust crate boundary and a JS bundle is a larger change than this bug warrants, and the doc comment records the coupling so the next person to change one has a pointer to the others. If you would rather have a single source, tell me where you want it to live and I will send that separately.
</details>

## Validation

The interesting part of this one is that the obvious test does not work.

A test that just asserts the three surfaces agree **passes on `main`, unfixed**, because ICU on a Linux runner already defaults to `en-US`. The reported default was a Windows build's. A host default is not reproducible across machines, so asserting agreement proves nothing about whether the engine pins anything.

`test_intl_default_locale_is_pinned_not_inherited_from_the_host` therefore forces a different default first and asserts the engine still names its own. That is safe because nextest runs each test in its own process, so the global never reaches a sibling test.

Red on this branch with the `pin_default_locale()` call commented out, which is the exact shape of the reported bug, three inherited values next to two pinned ones:

```
assertion `left == right` failed
  left: ["fr-FR", "fr-FR", "fr-FR", "en-US", "en-US"]
 right: ["en-US", "en-US", "en-US", "en-US", "en-US"]
```

Green with the call restored.

@yinnho raised a V8 subtlety on the issue that the test now encodes: V8 resolves the ICU default locale once per isolate and caches it at the first `Intl` use, so a pin landing after a page has formatted anything silently does nothing. Their literal suggestion, format then change the configured language then format again, does not translate directly because Obscura has no runtime language switch. The half that does translate is that the pin has to survive the host default moving underneath a *warmed* isolate, so the test now formats, moves the host default a second time, and formats again in the same isolate.

That second block is load-bearing, and what it reports without the pin is a direct measurement of the caching they described. Host default forced to `fr-FR`, isolate built and warmed, host default then moved to `de-DE`:

```
  left: ["fr-FR", "1 234,5"]
 right: ["en-US", "1,234.5"]
```

The warmed isolate answers **`fr-FR`**, not `de-DE`, and formats with French narrow-no-break-space grouping. It never saw the second change. That is the cache, and it is why `pin_default_locale` runs before `JsRuntime::new` rather than anywhere later. With the pin in place both surfaces read `en-US` and the number formats as `1,234.5`.

Since this changes shared JS-runtime code, I ran the whole four-configuration sequence from CONTRIBUTING rather than only `obscura-js`, in a `rust:latest` container so the result matches CI rather than my Windows host:

| Step | Result |
| --- | --- |

I have not run `cargo fmt` over the tree, per CONTRIBUTING; the added lines are rustfmt-clean on their own.

## Rendering

Not applicable. No layout, paint, screenshot, screencast or PDF code is touched.

## Performance

The call is inside a `std::sync::Once`, on the isolate-construction path, already under `ISOLATE_CREATE_LOCK`. It runs once per process and adds nothing per request, per page, or per script. There is no change to any hot path.

## Checklist

### DIFF
--- a/crates/obscura-js/src/runtime.rs
+++ b/crates/obscura-js/src/runtime.rs
@@ -3214,6 +3214,29 @@ mod tests {
             result,
             serde_json::json!(["en-US", "en-US", "en-US", "en-US", "en-US"])
         );
+
+        // The isolate has now used `Intl`, which is the state that matters:
+        // V8 resolves the default locale once per isolate and caches it, so a
+        // pin that lands after a page has formatted anything silently does
+        // nothing. That is why `pin_default_locale` runs before
+        // `JsRuntime::new` rather than anywhere later. Move the host default a
+        // second time, underneath the warmed isolate, and it must still be
+        // reading its own locale rather than picking the new one up.
+        deno_core::v8::icu::set_default_locale("de-DE");
+        let after_warmup = rt
+            .evaluate(
+                r#"
+            return [
+                Intl.DateTimeFormat().resolvedOptions().locale,
+                Intl.NumberFormat().format(1234.5),
+            ];
+        "#,
+            )
+            .unwrap();
+        // 1,234.5 is en-US grouping. de-DE would say 1.234,5 and fr-FR
+        // 1 234,5, so this asserts formatting behaviour and not just the name
+        // the locale reports for itself.
+        assert_eq!(after_warmup, serde_json::json!(["en-US", "1,234.5"]));
     }
 
     fn setup_runtime(html: &str) -> ObscuraJsRuntime {
--- a/crates/obscura-js/src/runtime.rs
+++ b/crates/obscura-js/src/runtime.rs
@@ -50,6 +50,32 @@ static SNAPSHOT: &[u8] = include_bytes!(env!("OBSCURA_SNAPSHOT_PATH"));
 /// parallel, each isolate on its own thread with no shared lock.
 static ISOLATE_CREATE_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
 
+/// The locale every locale-bearing surface reports.
+///
+/// Kept equal to the `navigator.language` and `navigator.languages` values in
+/// `js/bootstrap.js` and to the `Accept-Language` default in `obscura-net`.
+/// Those two were already pinned; `Intl` was not, so it reported whatever
+/// locale the V8 build happened to carry (#734 measured `en-AU` against a
+/// `navigator.language` of `en-US` on a Windows release build). A page that
+/// reads two of the three surfaces sees a browser disagreeing with itself,
+/// which is exactly the kind of inconsistency identity checks look for.
+const DEFAULT_LOCALE: &str = "en-US";
+
+static ICU_LOCALE_INIT: std::sync::Once = std::sync::Once::new();
+
+/// Pin ICU's default locale once, before the first isolate exists.
+///
+/// `Intl` resolves its default through ICU, not through anything the engine
+/// hands it, so leaving this unset makes the value a property of the build
+/// host rather than of the browser identity Obscura presents. Called under
+/// [`ISOLATE_CREATE_LOCK`] so the write cannot race an isolate being built on
+/// another connection thread.
+fn pin_default_locale() {
+    ICU_LOCALE_INIT.call_once(|| {
+        deno_core::v8::icu::set_default_locale(DEFAULT_LOCALE);
+    });
+}
+
 const DEFAULT_CDP_AWAIT_TIMEOUT_MS: u64 = 30_000;
 const HEAP_LIMIT_RECOVERY_HEADROOM_BYTES: usize = 64 * 1024 * 1024;
 
@@ -304,6 +330,11 @@ impl ObscuraJsRuntime {
             let _create_guard = ISOLATE_CREATE_LOCK
                 .lock()
                 .unwrap_or_else(|poisoned| poisoned.into_inner());
+
+            // Before the isolate, not after: ICU is consulted when a page
+            // first constructs an Intl object, and an isolate built ahead of
+            // this call would answer from the host default.
+            pin_default_locale();
 
             let mut runtime = JsRuntime::new(RuntimeOptions {
                 extensions: vec![build_extension()],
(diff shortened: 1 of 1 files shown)


ntdatt812:

| --- | --- | --- |
| this branch | 32/33 | 3016.3ms |

```
```

ntdatt812:
Superseded by 6f19abc — closing.

Your fix is the same pin in the same place, and before closing I checked whether my version still had anything worth salvaging. It does not, and the reason is worth writing down: **your test is mutation-proof on both properties that matter.** I ran the two mutations against `intl_locale_matches_navigator_language_regardless_of_host` on a Linux container:

| Mutation | Result |
|---|---|
| remove `set_default_locale("en-US")` entirely | **FAIL** — `left: "de-DE\|en-US"` |
| keep it, but move it *below* `JsRuntime::new(...)` | **FAIL** — `left: "de-DE\|en-US"` |

The second one is the interesting result. V8 resolves the ICU default once per isolate and caches it at first `Intl` use, so a pin that lands after the isolate is built is a silent no-op — and your test catches that placement regression, not just the missing call. Your comment says the create-lock placement "guarantees it lands before the first isolate exists"; that is now checked rather than only asserted, and someone tidying that line downward will get a red test rather than a quiet locale leak.

I had assumed `std::env::set_var("LC_ALL", ...)` after process start would be too late to reach ICU, which would have made the test pass regardless of the fix. That assumption was wrong — it reaches ICU fine under nextest's process-per-test, exactly as your comment says. My own version forced the default via `v8::icu::set_default_locale("fr-FR")` instead, which is more direct but guards nothing yours misses, so there is no follow-up here worth your review time.

Thanks for taking #734 — closing this as duplicate.

PR #743, opened 2026-08-29
fix(render): gate the webfont pseudo-content test on paint

### LINKED ISSUE #628: Build: obscura-render test target fails to compile without --features render (E0609 word_ifc_items)
### Before you start

### obscura version or commit

`6fa87e16520aaf13d6e45fa62e9e23ee81d28340` (current `main`)

### OS and architecture

Windows 11 x64

Other (building and testing from source as a contributor)

### Affected area

Build, packaging, or installation

### Build variant

No-render (official no-render archive or source build without render)

### Stealth mode

Not applicable

### Reproduction

From a clean checkout of `main`, without the `render` feature:

```
cargo check --tests
# or
cargo nextest run
```

Both fail before any test executes. Adding the feature makes it compile:

```
cargo check --tests --features render   # Finished, no errors
```

### Expected behavior

`obscura-render`'s test target compiles on default features, so a whole-workspace
test run is possible without `--features render`.

### Actual behavior

Compilation fails:

```
error[E0609]: no field `word_ifc_items` on type `dom::DomLayout`
    --> crates\obscura-render\src\dom.rs:19581:20
     = note: available fields are: `rects`, `inline_fragments`, `styles`,
       `custom_properties`, `clip_rects` ... and 4 others
```

`DomLayout::word_ifc_items` is declared `#[cfg(feature = "paint")]`
(`dom.rs:350`), but `dom::tests::generated_pseudo_content_shapes_with_the_loaded_webfont`
reads it unconditionally inside a plain `#[cfg(test)] mod tests`. `paint` is off
by default, so the test target cannot build. This entered with 4f48557
("fix(render): render webfont icons and color emoji"), part of #602 for #598.

Because this is a compile error in one crate's test target, it aborts the build
for the whole workspace, so no crate's tests run.

**A `#[cfg(feature = "paint")]` on that one test is not sufficient.** I applied
exactly that locally to see what was behind it, and once the crate compiles,
28 of 397 `obscura-render` tests fail on default features:

```
Summary [1.460s] 397 tests run: 369 passed, 28 failed, 1 skipped
```

They fail for one shared reason. The non-`paint` `inline` stub in `lib.rs`
discards web fonts (`fn new_with_web_fonts(_fonts: &[WebFont]) -> Self { TextEngine }`),
so a loaded face cannot change any geometry and every "the loaded face must drive
this measurement" assertion compares a value against itself:

```
thread 'dom::tests::canonical_inline_fragments_shape_with_the_loaded_webfont' panicked at
crates\obscura-render\src\dom.rs:19537:9:
assertion `left != right` failed: the loaded face's advances must drive fallback inline geometry
  left: 176.0
 right: 176.0
```

So the suite appears to be paint-only in practice, and the missing gate is one
instance of that rather than the whole story. Whether the right fix is gating the
affected tests, or giving the stub enough behavior to satisfy them, is a call for
whoever owns this area — I have not sent a patch for that reason.

Worth noting this is invisible to the documented workflow: CONTRIBUTING specifies
`cargo nextest run --release --features render`, and that path is unaffected. It
only bites a plain `cargo nextest run` or a no-render source build.

### Logs

```shell
$ cargo check --tests
error[E0609]: no field `word_ifc_items` on type `dom::DomLayout`
     --> crates\obscura-render\src\dom.rs:19581:20
      |
19581 |             loaded.word_ifc_items.contains_key(&token),
      |                    ^^^^^^^^^^^^^^ unknown field
      |
      = note: available fields are: `rects`, `inline_fragments`, `styles`, `custom_properties`, `clip_rects` ... and 4 others

For more information about this error, try `rustc --explain E0609`.
error: could not compile `obscura-render` (lib test) due to 1 previous error
```

I did not check (build-time issue, not a runtime behavior difference)

ntdatt812:
Fixes #628.

## What changed

`generated_pseudo_content_shapes_with_the_loaded_webfont` reads `DomLayout::word_ifc_items` from a plain `#[cfg(test)] mod tests`, but that field is declared `#[cfg(feature = "paint")]`. So `cargo check --tests -p obscura-render` on default features stopped before any test ran:

```
crates/obscura-render/src/dom.rs:19710:20: error[E0609]: no field `word_ifc_items` on type `dom::DomLayout`
error: could not compile `obscura-render` (lib test) due to 1 previous error
```

Exactly one error. The line has moved since the report was filed (`:19710`, not `:19581`); same test, same field.

The fix is `#[cfg(feature = "paint")]` on the test, plus a doc comment recording why it is the whole test rather than the one field access. Seven lines.

<details>
<summary><b>Why the whole test, and not just the field access</b></summary>

My first attempt gated only the `word_ifc_items` assertion, on the theory that the advance-width comparison beside it was pure layout and worth keeping alive in a no-render build. It compiles. It also fails:

```
FAIL obscura-render dom::tests::generated_pseudo_content_shapes_with_the_loaded_webfont
```

Without `paint`, `TextEngine::new_with_web_fonts` ignores the fonts it is handed. The build says so itself, in warnings that are easy to read past while chasing the error:

```
warning: associated function `new_with_web_fonts` is never used
warning: fields `data`, `family`, `weight` and `italic` are never read
```

So the "loaded" face is the fallback face, both sides of the `assert_ne!` measure the same font, and the widths are equal. Both halves of the test depend on `paint`, so the test does. Gating only the field would have traded a compile error for a failing test, which is the worse of the two: a compile error stops you, a red test looks like a regression and costs someone an investigation. That reasoning is in the doc comment above the gate so the next person does not repeat the attempt.
</details>

## Validation

```
# 1. the reported failure, default features
cargo check --tests -p obscura-render            -> Finished

# 2. paint still runs the test
cargo nextest run --release -p obscura-render --features paint generated_pseudo_content_shapes
   PASS generated_pseudo_content_shapes_with_the_loaded_webfont

# 3. paint suite unaffected
cargo nextest run --release -p obscura-render --features paint
   583 passed, 1 skipped
```

And the four-configuration sequence from CONTRIBUTING, since this is a feature-gate change, run in a `rust:latest` container:

| Step | Result |
| --- | --- |
| `cargo nextest run --release --features render --no-fail-fast` | **1506 passed**, 4 skipped (the count on `main`; this PR adds no test) |

## Rendering

Not applicable in the sense the template means: no fixture is needed and no output changes. The only edit is a `cfg` attribute and a comment on an existing test, and the paint-feature suite that owns the rendering behaviour is unchanged at 583 passing.

## Performance

No expected impact. The change is a test attribute; no code that ships in any binary is touched.

## Something this uncovers, which is deliberately not in this PR

The issue asks for "a whole-workspace test run without `--features render`". This change delivers the compile half, and the run then surfaces something that was invisible while compilation failed first:

```
cargo nextest run --release -p obscura-render --no-fail-fast
399 tests run: 370 passed, 29 failed, 1 skipped
```

29 failures, all one kind: tests that assert real text measurements (`auto_width_*`, `fixed_table_layout_*`, `inline_*`, `canonical_inline_fragments_shape_with_the_loaded_webfont`, and two `layout_test` replaced-element cases). Same root cause as the one above, that font shaping lives behind `paint`, so a no-render build measures text with a stand-in.

**None of this blocks merging.** Your CI's no-render step runs `--workspace --exclude obscura --exclude obscura-render` and then only `cargo check` on `obscura-render`, so those 29 are outside the gate either way; what was inside the gate was the `cargo check`, and this fixes it.

Which direction to take afterwards is a design question, not a bug fix: either those tests get the same gate, or a no-render build needs metrics good enough for them to hold. I would rather land the compile fix and let you pick than hand you 29 gated tests. Happy to open a separate issue with the full list, or to do the work once you have chosen.

## Checklist

### DIFF
--- a/crates/obscura-render/src/dom.rs
+++ b/crates/obscura-render/src/dom.rs
@@ -19678,6 +19678,13 @@ mod tests {
         );
     }
 
+    /// Gated on `paint`: without it `TextEngine::new_with_web_fonts` ignores
+    /// the fonts it is handed, so neither half of this test has anything to
+    /// observe. `word_ifc_items` does not exist to read, and the advance-width
+    /// comparison sees the same fallback face on both sides and finds the two
+    /// widths equal. Gating only the field access would leave an assertion
+    /// that compiles and then fails.
+    #[cfg(feature = "paint")]
     #[test]
     fn generated_pseudo_content_shapes_with_the_loaded_webfont() {
         let tree = parse_html(


ntdatt812:

| --- | --- | --- |
| this branch | 32/33 | 3015.5ms |

```
```

SGavrl (approved):

Issue #671, opened 2026-08-16
Obstacle course stage `observer-intersection` fails on main, headless Chrome fails it identically, and CI does not catch it


aech:
## Before you start

- [x] I searched existing issues and this is not a duplicate.
- [x] I tested against the latest release or `main`.
- [x] This is not a security vulnerability (those go through SECURITY.md, not a public issue).

## obscura version or commit

d9ef40b (main)

## OS and architecture

Ubuntu 24.04.4 LTS x86_64 (WSL2, Kernel 6.18.33.2)

## How are you using obscura?

CLI (fetch / serve / scrape / mcp)

## Affected area

Other

## Build variant

Render-enabled (official default archive or --features render)

## Stealth mode

Off

## Reproduction

AGENTS.md calls the obstacle course "the authoritative behavioral gate" and requires 33 of 33. CONTRIBUTING.md repeats the requirement. On `main` the full run reaches 32 of 33.

```bash
OBSCURA_BIN=./target/release/obscura python3 obstacle-course/run.py --runs 1 --warmup 0
```

```
correctness: 32/33 stages passed   |   median latency: 3027.9ms across stages
```

The only failing stage is `observer-intersection`:

```
observer-intersection  modern-web  FAIL  3061.0  3061.0  expected 'io:50', got ''
```

The fixture is `obstacle-course/fixtures/observer-intersection.html` from `obscura-benchmark` at 6ebac82. An IntersectionObserver observes a sentinel at the end of the feed. Per delivery the callback appends ten cards. At 50 cards it disconnects and sets `window.__obstacle`.

The same case without the fixture, against a local server:

```bash
obscura fetch http://127.0.0.1:PORT/fixtures/observer-intersection.html --allow-private-network --quiet --wait 3 --eval "JSON.stringify(document.querySelectorAll('.card').length + '|' + String(window.__obstacle||''))"
```

## Expected behavior

Two expectations contradict each other. That is what this issue is about.

The documentation expects the stage to pass. On `main` it does not. CI does not flag this.

The specification expects the opposite. An IntersectionObserver delivers an entry only when `isIntersecting` has flipped or the threshold index has changed. The fixture's sentinel is not replaced after the appends. It stays visible. Nothing changes, so no second entry is created.

## Actual behavior

obscura delivers ten cards. The callback fires exactly once.

[...]

PR #744, opened 2026-08-29
test(browser): cover the load-delaying pump's page-error path

### LINKED ISSUE #699: 0.2.1: an unhandled JSON.parse in a page script kills the load-delaying dynamic script event loop, so pending scripts are never processed (Chrome throws the same and is unaffected)
### Before you start

### obscura version or commit

**0.2.1** (official release binary `obscura-aarch64-linux-stealth.tar.gz`, sha256 `77704cf1…2b86828a`), run **without** `--stealth`.

### OS and architecture

Linux aarch64 — Raspberry Pi 5, Debian GNU/Linux 12 (bookworm), Node v22.23.1, puppeteer-core 24.37.5 over CDP.

### Stealth mode

Reproduces without `--stealth`. (I deliberately do not use `--stealth` here: on 0.2.x it also does tracker blocking, which blocks the very anti-fraud scripts under test.)

---

## First — thank you, #592 is fixed and verified on real hardware

0.2.1 landed #593 and iframe realms are now populated. Verified with the one-liner from #592:

```json
{"Object":"function","Function":"function","Error":"function","Promise":"function","XMLHttpRequest":"function","fetch":"function"}
```

All twelve constructors I probe are present across all three iframe flavours (no `src`, `src="about:blank"`, `srcdoc`), `new f.contentWindow.XMLHttpRequest()` works, and the realm stays correctly separate. The downstream effect on the real site is large: the page went from **20 requests to 281**, and resources that never loaded before now do — `cdn/ca/lwsa.html` (the iframe document), `cdn/cd/gpk` now with the `dtsi` parameter like Chrome, `gwf?e=v&v=latest` (the correct resource), and `sifo` `login.js` + `POST pageFeatures`. The old `a0dr is not a constructor` and `There is no supported http request object` errors are **gone**.

The site still does not work, but for a new and unrelated reason.

## Summary

On a page with many dynamically inserted scripts, an **unhandled exception thrown from page script during the load-delaying dynamic script pump kills that event loop**, and the still-pending scripts are then never processed:

```
WARN obscura_browser::page: load-delaying dynamic script event loop failed:
  Event loop error: SyntaxError: Unexpected token 'l', "l5uBRa3EJ3"... is not valid JSON
    at JSON.parse (<anonymous>)
    at K (https://onlinebanking.bancogalicia.com.ar/login:233:252223)
    at https://onlinebanking.bancogalicia.com.ar/login:233:251708
    at eventLoopTick (ext:core/01_core.js:178:7)
WARN obscura_browser::page: script deadline reached with load-delaying dynamic scripts still pending
```

**The key point: that `JSON.parse` throw is normal and benign.** Chromium, on the same page, throws the *same* class of error — I instrumented `JSON.parse` in both engines and Chromium records **12+ identical failures** (`JSON.parse("Window_fdfbaa87-f2bc-4a8e-910a-3746c4618ba3")` from `we-stats/ad1a29c5.js`) — and simply carries on. The obfuscated bundle uses throwing `JSON.parse` as an idiom. Chrome absorbs it; obscura loses the whole pump.

Consequence: the site's device-fingerprint SDK (`uppopi-*-sb.min.js`) is never inserted, so the `cdSNum` cookie is never set.

## Measured comparison (same page, same machine, minutes apart)

| signal | obscura 0.2.1 | Chromium 150 |
|---|---|---|
| `<script>` elements inserted | **9** | 35 |
| `hero-app.js` present as a `<script>` | **null** (it *is* fetched over the network) | ✅ |
| `uppopi-*-sb.min.js` requested | **false** | **true** |
| scripts belonging to the `cd` SDK | **`[]`** | several |
| `cdSNum` cookie | **absent** | present |
| `bmuid` / `cdContextId` cookies | present | present |
| `JSON.parse` failures (instrumented) | 1, **fatal to the pump** | 12+, **harmless** |
| `DOMContentLoaded` / `window.load` | both fire | both fire |
| errors visible to CDP (`pageerror`, `console`) | **none** | 1 (a site-authored message, identical in both) |

Nothing surfaces through CDP at all — the two `WARN` lines above only appear on the server's own stderr, which is what made this hard to find.

## What I could NOT reproduce (so you don't waste time)

I tried to reduce it to a synthetic case on `example.com` and **failed** — I am reporting the real-site evidence rather than a clean repro. Both attempts behave *identically to Chromium* on 0.2.1, with B and C still executing:

1. A `data:` script that executes and throws synchronously (`JSON.parse('l5uBRa3EJ3')`), with a sibling script inserted in the same turn and a third chained from its `onload`. Result: `aExec/aThrew/bExec/cExec` all true. obscura reports `Dynamic script error (…)` and **contains it correctly**.
2. The same, but throwing **asynchronously** (a rejected microtask and a throwing `setTimeout`), plus a network-loaded load-delaying script. Result: again all flags true. Here obscura does log `WARN autonomous page task failed: Event loop error: …` — the same *shape* of message, but from `obscura_cdp::server`, and the page keeps working.

So the failing path is specifically `obscura_browser::page`'s **load-delaying dynamic script** pump, and the trigger needs a condition my toy cases don't hit — presumably an exception arriving while several load-delaying scripts are genuinely in flight, after which they hang until `OBSCURA_SCRIPT_DEADLINE_MS` expires. The real page reaches it every single time.

The stack ends in `eventLoopTick`, so the throw reaches the loop from an async callback rather than from top-level script evaluation.

## Expected behavior

An unhandled exception from page script — sync or async — should be reported and contained, exactly as in #358, and must not stop the remaining dynamically inserted scripts from being fetched, executed, or chained. Chrome reports such exceptions to `window.onerror`/`pageerror` and keeps processing.

## Actual behavior

The load-delaying dynamic script event loop fails, the pending scripts are abandoned, and the run ends with `script deadline reached with load-delaying dynamic scripts still pending`. Raising the budgets (`OBSCURA_NAV_TIMEOUT_MS=120000`, `OBSCURA_SCRIPT_DEADLINE_MS=90000`, `OBSCURA_MODULE_BUDGET_MS=15000`, `OBSCURA_CDP_COMMAND_TIMEOUT_MS=180000`) raises the request count from 158 to 281 but does **not** change the outcome — it is not a matter of time.

No. Chromium 150.0.7871.181 on the same machine throws the same `JSON.parse` errors, more often, and loads the SDK anyway.

## Real-world impact

Same production banking SPA as #255 / #518 / #520 / #592 (Banco Galicia online banking). With the SDK never bootstrapping, the login is rejected with valid credentials that Chromium accepts. This is the fifth distinct engine gap on this one page; the previous four (#485 SVG namespace, #519 `referrer`/nav-timing, #520 dynamic script chaining, #592 iframe realms) are all fixed and confirmed fixed on real hardware — each fix moved the SDK measurably further along.

## Secondary observation (tell me if you'd rather have this as its own issue)

`frame.evaluate()` against a **cross-origin child frame** never resolves — it times out after 30s, for both child frames, while the main frame evaluates fine:

```
frameCount: 3   (obscura)          frameCount: 8   (Chromium)
  https://onlinebanking.bancogalicia.com.ar/login   -> evaluates OK (26 scripts)
  https://logo.prismasystems.com.ar/…/logogalicia.html -> Timed out after waiting 30000ms
  https://portals3.galicia.ar/cdn/ca/lwsa.html      -> Timed out after waiting 30000ms
```

Chromium evaluates in all eight and reports `readyState: complete` with 2 and 1 scripts respectively. Since 0.2.1 makes child frames run their own scripts and report a real frame tree, being unable to evaluate into them makes the new capability hard to observe or drive.

ntdatt812:
**Rescoped.** ad29745 fixed #699 upstream while this was open, and your fix is broader than mine was — it covers all four pumps rather than only the load-delaying one, and `is_fatal_event_loop_error` is a better seam than the local error counter I had. I have dropped my implementation entirely and reduced this PR to the one piece your commit did not carry: **a test at the layer where the bug was reported.**

## Why this layer

ad29745 changes `Page::drive_load_delaying_scripts` in `obscura-browser` and adds no test there. The tests that exist do not reach the new branch:

| Test | Layer | Covers the error path? |
|---|---|---|
| `unhandled_rejection_does_not_starve_later_tasks` (yours, ad29745) | `obscura-js`, `run_event_loop_bounded` | yes, but not this pump |
| `load_delaying_script_driver_respects_absolute_deadline` (existing) | `obscura-browser`, this pump | no — deadline path only |

So the `is_fatal_event_loop_error` branch in `page.rs` is currently unguarded, and #699 was reported as *load-delaying scripts going missing*, which is this function.

## Mutation-checked

Reverting your branch to the unconditional bail it replaced:

```rust
Ok(Err(error)) => {
    tracing::warn!("load-delaying dynamic script event loop failed: {error}");
    return false;
}
```

```
FAIL  load_delaying_script_driver_survives_a_page_script_exception
      the throw must not end the pump while a script is still pending
PASS  load_delaying_script_driver_respects_absolute_deadline
```

The deadline test passing under the same mutation is the point: the existing suite does not notice the regression, and this test does. Restored, both pass:

```
2 tests run: 2 passed, 65 skipped
```

`cargo nextest run --release --features render -p obscura-browser load_delaying_script_driver`, Linux container, rustc 1.98.

## The test

A load-delaying classic script is left pending against a 150 ms server while a page callback throws. The throw is deferred through a timer so it lands on a pump tick — thrown during the installing script's own microtask drain it would clear the pending-script bookkeeping before the fetch starts, which is a different path. The assertions are that the pump reports completion, that the pending script was still fetched, and that the pending set drained.

No production code in this PR. `rustfmt --check` is clean across the inserted range (the file carries unrelated pre-existing diffs elsewhere).

Closes nothing on its own — #699 is already fixed by ad29745.

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -5648,6 +5648,73 @@ mod tests {
     }
 
     #[tokio::test(flavor = "current_thread")]
+    async fn load_delaying_script_driver_survives_a_page_script_exception() {
+        // An exception from an async page callback reaches the pump as an
+        // event-loop error, and the pump used to answer it by abandoning every
+        // still-pending load-delaying script. The error is transient: measured
+        // against this runtime, the tick carrying it fails and the following
+        // ticks poll clean, so the pending script below was dropped over a
+        // condition that had already cleared, and dropped silently, since the
+        // script simply never arrives.
+        //
+        // The throw is deferred through a timer so it lands on a pump tick.
+        // Thrown during the installing script's own microtask drain it would
+        // clear the pending-script bookkeeping before the fetch even starts,
+        // which is a different bug from the one under test here.
+        let (base, requests) = spawn_delayed_classic_script_server(
+            std::time::Duration::from_millis(150),
+            "globalThis.__lateDynamicRan = true;",
+        );
+        let mut page = import_map_test_page(
+            "load-delayer-throws",
+            "http://127.0.0.1:9",
+            "<html><head></head><body></body></html>",
+        );
+        page.js
+            .as_mut()
+            .unwrap()
+            .execute_script(
+                "install-load-delayer",
+                &format!(
+                    "globalThis.__documentReadyState__ = 'loading'; \
+                     const script = document.createElement('script'); \
+                     script.src = '{base}/slow-dynamic.js'; \
+                     document.head.appendChild(script); \
+                     setTimeout(() => {{ \
+                         queueMicrotask(() => {{ throw new Error('page script boom'); }}); \
+                     }}, 0);",
+                ),
+            )
+            .unwrap();
+        assert!(page
+            .js
+            .as_mut()
+            .unwrap()
+            .has_pending_load_delaying_scripts());
+
+        let deadline = tokio::time::Instant::now() + std::time::Duration::from_secs(5);
+        let completed =
+            super::Page::drive_load_delaying_scripts(page.js.as_mut().unwrap(), deadline).await;
+
+        assert!(
+            completed,
+            "the throw must not end the pump while a script is still pending"
+        );
+        assert_eq!(
+            requests
+                .recv_timeout(std::time::Duration::from_secs(1))
+                .unwrap(),
+            "/slow-dynamic.js",
+            "the pending script must still be fetched"
+        );
+        assert!(!page
+            .js
+            .as_mut()
+            .unwrap()
+            .has_pending_load_delaying_scripts());
+    }
+
+    #[tokio::test(flavor = "current_thread")]
     async fn post_load_dynamic_script_waits_only_when_caller_requests_settle() {
         let (base, requests) = spawn_delayed_classic_script_server(
             std::time::Duration::from_millis(400),


yinnho:
Reviewed the patch — the shape is right.

The deadline-as-only-bound tradeoff is the correct call: on a page that throws every tick, the pump now burns its remaining budget instead of abandoning still-pending work over an error that a fresh tick would clear, and that is bounded (the absolute deadline) rather than unbounded. `yield_now()` in the error arm keeps a pathological page from spinning the loop hot between polls, and warn-once/debug-after keeps a noisy page from flooding the log. The regression test matches the measurement table from #699 (throw deferred through a timer so it lands on a pump tick).

One question, from the test comment: *"Thrown during the installing script's own microtask drain it would clear the pending-script bookkeeping before the fetch even starts, which is a different bug from the one under test here."* Is that second failure mode tracked anywhere? We would like to check our tree for it — if it turns out to be real on our side too, it deserves its own issue rather than riding on this PR.

ntdatt812:

| --- | --- | --- |
| this branch | 32/33 | 3016.0ms |

```
```

yinnho:
That stage is #671, and the back-to-back table here (no delta between `main` and this branch) is the expected shape of it — nothing in this PR touches the observer path. Two notes that might save the next person the chase:

- The failure is spec-correct withholding, not a missing computation step. The fixture observes the sentinel exactly once (its comment claims a fresh observation per batch, but `io.observe()` runs once), and nothing flips `isIntersecting` or the threshold index after the appends — so an engine following the spec delivers one entry and stops. Headless Chromium 151 produces the identical `expected 'io:50', got ''` through the same runner; measurement tables and the aff0172 → `10b8492` history are in #671.
- For completeness: our obscura-lineage engine (diting) never returns empty on this stage either, because its IO is still the pre-`10b8492` geometry-naive emulation (always-intersecting, mutation re-fire, burst schedule) — which is what the fixture was originally written against. So the stage currently measures engine history rather than browser behavior from both directions: spec-following engines (including Chrome) fail on the missing re-observation, and the fixture's comment describes an emulation that no longer matches `main`.

Until the fixture re-observes or toggles the sentinel, I'd treat the stage as known-fail per #671 rather than a regression signal for PRs like this one.

SGavrl (approved):

PR #745, opened 2026-08-29
fix(mcp): drain a synthesized navigation before the tool replies

### LINKED ISSUE #618: MCP `browser_click` on a submit button changes `location.href` without issuing any request
### Before you start

### obscura version or commit

6fa87e1

### OS and architecture

Manjaro x86

### Affected area

CLI or MCP

### Build variant

Render-enabled (official default archive or --features render)

### Stealth mode

None

### Reproduction

Minimal page served at `http://127.0.0.1:8940/`:

```html
<!doctype html><meta charset="utf-8">
<form id="f" method="POST" action="/submitted">
  <input name="q" id="q" value="">
  <button type="submit" id="go">Envoyer</button>
</form>
```

The server logs every request it receives (method, path, body).

### Path A — MCP (`obscura --allow-private-network mcp`, stdio)

```
browser_navigate {"url": "http://127.0.0.1:8940/"}
browser_fill     {"selector": "#q", "value": "hello"}
browser_click    {"selector": "#go"}      -> "Clicked '#go'"
browser_evaluate {"expression": "location.href"}
                                          -> "http://127.0.0.1:8940/submitted"
```

Requests seen by the server:

```
GET / body=''
```

That is all. No request for `/submitted`.

### Path B — CDP (`obscura --allow-private-network serve`, Puppeteer `connectOverCDP`)

```js
await page.goto('http://127.0.0.1:8940/');
await page.type('#q', 'hello');
await page.click('#go');
```

Requests seen by the server:

```
GET / body=''
POST /submitted body='q=hello'
```

Same binary, same page, same button. Only the driving layer differs.

### Expected behavior

Clicking a submit button through the MCP tool `browser_click` reports success and updates
`location.href` to the form's `action` then a HTTP reques is sent.

### Actual behavior

Clicking a submit button through the MCP tool `browser_click` reports success and updates
`location.href` to the form's `action` but **no HTTP request is ever sent**. The same click driven through CDP (`Input.dispatchMouseEvent`, e.g. Puppeteer's `page.click()`) correctly issues the `POST`.

The result is a silent false success: an agent sees `Clicked '#go'`, reads back the expected URL, and concludes the form was submitted — while the server never received anything.

### Rendering details

_No response_

### Logs

```shell
INFO obscura_browser::page: script execution phase complete phase="script-execution-total" elapsed_ms=0 budget_ms=30000
```

### LINKED ISSUE #640: MCP: the page task queue is never pumped, so timers, fetch and queued navigations never run
### Before you start

### obscura version or commit

0.2.0 (release archive `obscura-aarch64-macos.tar.gz`)

### OS and architecture

macOS 26 (Darwin 25.5.0), arm64

### Affected area

CLI or MCP

### Build variant

Render-enabled (official default archive or --features render)

### Stealth mode

Off

### Reproduction

Over the MCP transport, the synchronous part of an operation runs and the queued
part never does. Microtasks drain, nothing else does. I think this single problem
explains several separate-looking symptoms, including #618.

Probe on `data:text/html,<p id=log>idle</p>`:

```
browser_evaluate (() => { globalThis.t = "no"; globalThis.raf = "no";
  globalThis.m = "no"; globalThis.f = "no";
  setTimeout(() => { globalThis.t = "yes"; }, 0);
  requestAnimationFrame(() => { globalThis.raf = "yes"; });
  queueMicrotask(() => { globalThis.m = "yes"; });
  fetch("https://example.com/").then(() => { globalThis.f = "yes"; })
    .catch(e => { globalThis.f = "err:" + e.message; });
  return "armed"; })()

browser_evaluate JSON.stringify({timer: globalThis.t, raf: globalThis.raf,
                                 microtask: globalThis.m, fetch: globalThis.f})
-> {"timer":"no","raf":"no","microtask":"yes","fetch":"no"}
```

Only the microtask ran. The timer and the animation frame never fired. The fetch
promise never settled, not even to an error.

**1. The wait tools cannot wait.** They only see DOM that is already there. A
500 ms timer that inserts `#late`:

```
browser_evaluate (() => { setTimeout(() => { document.body.insertAdjacentHTML(
  "beforeend", "<div id=late>late</div>"); }, 500); return "armed"; })()
browser_wait_for {"selector": "#late", "timeout": 5}
-> Error: Timeout waiting for '#late'
```

`browser_scroll`, `browser_screenshot` and the wait tools pump nothing either, so
`browser_scroll {"direction": "bottom"}` cannot trigger an infinite-scroll
loader, although its tool description says it can.

**2. Clicks and form submits change the URL and load nothing.** This is #618, and
it is not only about submit buttons:

```
browser_navigate https://quotes.toscrape.com/page/1/
browser_interactive_elements            -> ref=e53  a  "Next →"
browser_click {"ref": "e53"}            -> "Clicked ..."
browser_evaluate location.href          -> https://quotes.toscrape.com/page/2/
browser_evaluate document.querySelector("span.text").textContent
                                        -> still page 1's first quote
browser_reload                          -> "Reloaded .../page/1/"
```

`location.assign()`, `location.replace()`, `history.pushState`,
`<meta http-equiv=refresh>` and `window.open()` (returns null) behave the same.
`location` is set immediately, the load never happens. This is the worst symptom
for automation, because nothing fails: you read the new URL and scrape the old
page.

**3. Only `networkidle0` settles a JS-built page, and reload undoes it.**
https://quotes.toscrape.com/scroll builds its list from `/api/quotes`:

```
browser_navigate {"url": ".../scroll"}                           -> count div.quote = 0
browser_navigate {"url": ".../scroll", "waitUntil": "networkidle0"} -> count = 10
browser_reload                                                   -> count = 0
```

So a default `load` navigate does not run the XHR, and `browser_reload` throws
away content that was there.

**The engine is fine, the MCP layer is not.** Same binary over CDP
(`obscura serve --port 9223`, browser endpoint, `Target.createTarget` ->
`Target.attachToTarget {flatten: true}` -> sessionId-scoped commands), against a
local page with a link and a form:

```
Runtime.evaluate document.getElementById('lnk').click()
  -> location /target.html, title "TargetPage"    (server log: GET /target.html)
Runtime.evaluate document.getElementById('f').submit()
  -> location /submitted                          (server log: POST /submitted)
```

Both of those are no-ops over MCP. #618 already noticed the CDP/MCP difference
for the form case. The task queue explains that, and also explains the dead
timers and the dead fetch.

### Expected behavior

The MCP server drives the page task queue between tool calls, so timers, network
completions and queued navigations run. At minimum the `browser_wait_for` and
`browser_wait_for_text` tools should pump while they wait, otherwise they can
only ever confirm what was already in the DOM.

### Actual behavior

Anything that needs a task-queue turn is silently dropped. Timers and animation
frames never fire, `fetch` and XHR never settle, and a navigation started by a
click, a `location` assignment or a form submit updates `location` without
loading the document.

### Rendering details

_No response_

### Logs

```shell
Not applicable, nothing is logged. Every call above returns success.
```

ntdatt812:
Fixes #618.

## What changed

`settle_synthetic_navigation()` on `BrowserState`: consume the pending navigation, and if one landed, clear the ref table (it names elements in a document that is gone) and give the new document one event-loop slice, so a tool that reads the URL or the text next sees the landed page rather than an empty one. It is a no-op when nothing is pending.

It is called from `browser_click`, `browser_fill`, `browser_type`, `browser_press_key` and `browser_evaluate`, not from `browser_click` alone. The invariant @yinnho describes is the useful thing here: *every tool that can synthesize a click, a keypress, or run a script drains pending navigation before it replies*. `browser_evaluate` running `location.href = ...` or `form.submit()`, and an Enter through `browser_press_key`, land in the same pending state, so fixing only the reported tool would leave the same bug one call to the left. Those five tools become `async`; two existing tests that called them directly gained `.await`.

@yinnho's framing in the issue is right and holds here exactly: **pumping is not draining.** A submit click runs the page's own form glue and leaves the navigation as pending state. It becomes a request only when a driving layer converts it. The CDP path converts it, in two places: `crates/obscura-cdp/src/domains/input.rs:227` after a dispatched mouse event, and `crates/obscura-cdp/src/domains/runtime.rs:32` after `Runtime.evaluate`. MCP had no equivalent, which is the whole of your Path A versus Path B.

<details>
<summary><b>Why the existing pump does not already cover it</b></summary>

`BrowserState::advance_active_page_tasks` already pumps and drains, and it is armed after every dispatch. It still never gets a turn here, for two independent reasons:

```rust
tokio::select! {
    biased;
    read = reader.read_line(&mut line) => Some(read?),
    pump_result = state.advance_active_page_tasks() => { ... }
}
```

The select is `biased` with `read_line` first, so a client that already has its next tool call queued, which an agent always does, wins that race every time. And the response is written before the loop comes back around at all, so even an unbiased select would reply first.

That is also where the misleading part of the report comes from. `location.href` reports the destination because the pending navigation has already updated the page's URL state. The agent is told the submit happened while no request has left the process, which is worse than an outright failure.
</details>

## Validation

New test `click_on_a_submit_button_issues_the_request_before_replying`, built on your repro page: a recording server serving the form, `browser_fill` on `#q`, `browser_click` on `#go`. The assertion is at the wire, not on `location.href`, because the URL was the thing that lied.

Red with the drain removed from `tool_click`, green with it:

```
FAIL obscura-mcp tests::click_on_a_submit_button_issues_the_request_before_replying
PASS obscura-mcp tests::click_on_a_submit_button_issues_the_request_before_replying
```

Full four-configuration sequence from CONTRIBUTING, in a `rust:latest` container so the result matches CI rather than my Windows host:

| Step | Result |
| --- | --- |

## Rendering

Not applicable. No layout, paint, screenshot, screencast or PDF code is touched. `browser_screenshot` and `browser_pdf` are not among the five tools changed.

## Performance

The added work is one `process_pending_navigation()` per call to those five tools, which returns immediately when nothing is pending, so the cost on the common path is a check. When a navigation *is* pending the tool now waits for one event-loop slice before replying, which is work the client was going to pay for on its next call anyway, moved to where it is correct. The other MCP tools are untouched.

## Relationship to the other issues on this

#640's framing, that the transport never pumps the page task queue, is not what breaks this: the pump exists and is armed. It loses a race and, more fundamentally, runs after the reply. So this fix is orthogonal to any pump change and **does not close #640**.

I have not touched `location.href` reporting a destination for a navigation that has not been converted. After this change the two agree for these tools, but the underlying "URL state moves before the request" behaviour is still reachable from anywhere that does not drain, and it deserves its own issue rather than a rider here.

## Checklist

### DIFF
--- a/crates/obscura-mcp/src/lib.rs
+++ b/crates/obscura-mcp/src/lib.rs
@@ -169,6 +169,42 @@ impl BrowserState {
         Ok(reached_idle && !navigated)
     }
 
+    /// Consume a navigation that a synthesized interaction just queued,
+    /// before the tool replies.
+    ///
+    /// A click on a submit button, an Enter keypress, or an evaluated
+    /// `location.href` assignment does not issue a request by itself. It runs
+    /// the page's own glue and leaves the navigation as pending state, which
+    /// becomes a request only when a driving layer converts it. The CDP path
+    /// already converts it, in `Input.dispatchMouseEvent` and after
+    /// `Runtime.evaluate`, which is why the same click POSTs over CDP and
+    /// does nothing over MCP (#618).
+    ///
+    /// [`Self::advance_active_page_tasks`] cannot cover this. It is armed
+    /// after every dispatch, but it sits in a `biased` select behind
+    /// `read_line`, so a client that sends its next tool call immediately,
+    /// which an agent does, wins that race every time. The reply is also
+    /// written before any pump turn could run, so the tool would answer
+    /// "Clicked" while the request has not left the process.
+    async fn settle_synthetic_navigation(&mut self) -> Result<(), String> {
+        let navigated = self
+            .page_mut()
+            .process_pending_navigation()
+            .await
+            .map_err(|error| error.to_string())?;
+        if navigated {
+            // The ref table names elements in a document that has gone away.
+            self.interactive_refs.clear();
+            // One slice on the landed document, so a tool that reads the URL
+            // or the text next sees the new page rather than an empty one.
+            self.page_mut()
+                .run_autonomous_event_loop_turn()
+                .await
+                .map_err(|error| error.to_string())?;
+        }
+        Ok(())
+    }
+
     /// Resolve `ref=eN` to a CSS selector that uniquely targets the
     /// element. Snapshot writes `data-obscura-ref="eN"` onto every
     /// interactable, so the attribute survives across calls as long as
@@ -719,12 +755,12 @@ async fn handle_tool_call(id: Value, params: &Value, state: &mut BrowserState) -
     let result = match name {
         "browser_navigate" => tool_navigate(args, state).await,
         "browser_snapshot" => tool_snapshot(args, state),
-        "browser_click" => tool_click(args, state),
-        "browser_fill" => tool_fill(args, state),
-        "browser_type" => tool_type(args, state),
-        "browser_press_key" => tool_press_key(args, state),
+        "browser_click" => tool_click(args, state).await,
+        "browser_fill" => tool_fill(args, state).await,
+        "browser_type" => tool_type(args, state).await,
+        "browser_press_key" => tool_press_key(args, state).await,
         "browser_select_option" => tool_select_option(args, state),
-        "browser_evaluate" => tool_evaluate(args, state),
+        "browser_evaluate" => tool_evaluate(args, state).await,
         "browser_wait_for" => tool_wait_for(args, state).await,
         "browser_network_requests" => tool_network_requests(state),
         "browser_console_messages" => tool_console_messages(state),
@@ -942,7 +978,7 @@ fn tool_snapshot(args: &Value, state: &mut BrowserState) -> Result<String, Strin
     Ok(format!("URL: {url}\nTitle: {title}\n\n{body}{refs_summary}"))
 }
 
-fn tool_click(args: &Value, state: &mut BrowserState) -> Result<String, String> {
+async fn tool_click(args: &Value, state: &mut BrowserState) -> Result<String, String> {
     let selector = resolve_target(args, state)?;
 
     let js = format!(
@@ -962,11 +998,12 @@ fn tool_click(args: &Value, state: &mut BrowserState) -> Result<String, String> 
         // A click can navigate or rewrite the DOM; the old ref table may
         // no longer match. Conservative: invalidate. Next snapshot rebuilds.
         state.interactive_refs.clear();
+        state.settle_synthetic_navigation().await?;
         Ok(format!("Clicked '{selector}'"))
     }
 }
 
-fn tool_fill(args: &Value, state: &mut BrowserState) -> Result<String, String> {
+async fn tool_fill(args: &Value, state: &mut BrowserState) -> Result<String, String> {
     let selector = resolve_target(args, state)?;
     let value = args.get("value").and_then(Value::as_str)
         .ok_or("Missing value parameter")?;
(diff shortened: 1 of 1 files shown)


ntdatt812:

| --- | --- | --- |
| this branch | 32/33 | 3016.0ms |

```
```

SGavrl (approved):

PR #747, opened 2026-08-29
fix(cdp): report a thrown or rejected value through exceptionDetails

### LINKED ISSUE #746: Runtime.evaluate reports a rejected promise as a protocol error instead of exceptionDetails
## Before you start

- [x] I tested against `main` (`c138019`).

## obscura version or commit

`c138019` (main)

## OS and architecture

Debian x86_64 (`rust:latest` container), release build

## How are you using obscura?

## Affected area

CDP, Puppeteer, or Playwright compatibility

## Build variant

No-render and render both; the path is in `obscura-js` and `obscura-cdp`

## Reproduction

`Runtime.evaluate` with `awaitPromise: true` on an expression that rejects, driven through the CDP handler:

```
Promise.reject(new Error('boom'))               =>  ERR  Promise rejected: boom
(async () => { throw new Error('boom') })()     =>  ERR  Promise rejected: boom
undefined_variable_xyz  (awaitPromise: true)    =>  ERR  Promise rejected: undefined_variable_xyz is not defined
```

Each of those is a CDP **protocol error**: the command fails, and the reply carries no `result` and no `exceptionDetails`.

The resolving forms are all correct on `main`, so this is narrowly the rejection path:

| expression | `main` |
| --- | --- |
| `Promise.resolve(2)` | `{"type":"number","value":2.0}` ✅ |
| `(async () => ({ok:true}))()` | `{"type":"object","value":{"ok":true}}` ✅ |
| `new Promise(r=>setTimeout(()=>r('timer'),150))` | `{"type":"string","value":"timer"}` ✅ |
| `Promise.reject(new Error('boom'))` | protocol error ❌ |

## Expected behavior

The protocol defines `Runtime.evaluate` as returning `result` plus an optional `exceptionDetails`, and a thrown or rejected value is reported *through* that field on a **successful** command, not by failing the command. `Runtime.exceptionDetails` carries `exceptionId`, `text`, `lineNumber`, `columnNumber` and an `exception` RemoteObject.

That field is what Puppeteer reads to rebuild the page's own error and re-throw it out of `page.evaluate`. With a protocol error instead, the client sees a transport-shaped failure carrying a stringified message rather than the page's `Error`, and cannot recover `name`, the stack, or a non-`Error` rejection value at all. A rejection with a non-Error value degrades the worst: `Promise.reject({code: 42})` currently stringifies through `String(obj.message || obj)`.

## Actual behavior

`exceptionDetails` is never emitted anywhere. On `main`:

```
$ grep -rn "exceptionDetails" crates/
crates/obscura-cdp/src/domains/runtime.rs:423:  "getExceptionDetails" => Ok(json!({ "exceptionDetails": null })),
```

The one occurrence is an unrelated stub. `evaluate_for_cdp_with_timeout` in `crates/obscura-js/src/runtime.rs` already captures the rejected value into the object store and sets `__obscura_await_rejected`, then converts both into `Err(format!("Promise rejected: {}", ...))` at `runtime.rs:1611`, which the dispatcher turns into the protocol error. The information needed for `exceptionDetails` is already in hand at that point; only the shape it is reported in is wrong.

No. Chrome answers the command successfully and puts the thrown value in `exceptionDetails.exception`. @yinnho reported the same for their obscura-lineage engine in the table on #541.

## Relationship to other issues

- #541 is the resolving half of this call and **is fixed on `main`** (measurements in [this comment](https://github.com/h4ckf0r0day/obscura/issues/541#issuecomment-5462261678)); this issue is the rejecting half that remains.
- #677 and #719 are the `Runtime.exceptionThrown` **event**. Different surface: this is the `exceptionDetails` field on the evaluate **reply**.

## Notes

I would like to take this one. Two adjacent findings on the synchronous path that I am deliberately keeping out of scope, and will file separately if you want them:

- `undefined_variable_xyz` with `awaitPromise: false, returnByValue: true` returns `{"type":"object","subtype":"null","value":null}`. A `ReferenceError` reported as a successful `null`, from the `catch(e) { __result = undefined; }` in the non-await wrapper.
- `throw new Error('x')` is a `SyntaxError` on that same path, because the wrapper parenthesises the expression and `(throw ...)` is not an expression.

ntdatt812:
## What changed

Fixes #746.

`Runtime.evaluate` and `Runtime.callFunctionOn` could not report that a page threw. Each of the four shapes of the command lost the error a different way:

| input | | before | after |
|---|---|---|---|
| `Promise.reject(new Error('x'))` | `awaitPromise: true` | protocol error | `exceptionDetails` |
| `callFunctionOn` returning a rejection | `awaitPromise: true` | **success, `{}`** | `exceptionDetails` |
| `undefined_variable_xyz` | `awaitPromise: false` | success, `type: "undefined"` | `exceptionDetails` |
| `undefined_variable_xyz` | `returnByValue: true` | **success, `value: null`** | `exceptionDetails` |
| `throw new Error('boom')` | `awaitPromise: false` | protocol error (SyntaxError) | `exceptionDetails` |
| `var x = 1; x * 2` | `returnByValue: true` | success, `value: null` | `2` |

The `callFunctionOn` row is the one that matters most: a rejection was **indistinguishable from a resolution**. `Promise.reject({code: 42})` came back byte-identical to `Promise.resolve({code: 42})`, and an `Error` came back as `{}` because it has no own enumerable properties to serialize. That is the path `page.evaluate(fn)` uses, so `await page.evaluate(() => Promise.reject(new Error('boom')))` resolved to `{}` instead of throwing.

### One outcome protocol

CDP reports a thrown value on a **successful** reply, in `exceptionDetails` — not as a protocol error. So the difference between "this is the result" and "this was thrown" has to survive the trip out of the runtime, which is what the new `thrown` flag on `RemoteObjectInfo` carries.

Every shape now writes the same two globals the await path already wrote (`__obscura_await_meta`, `__obscura_await_rejected`), and the check for them moved out of the `if await_promise` block to cover both. The reply builder no longer has to know which path produced the value. Concretely:

- the `callFunctionOn` wrapper sets the flag on rejection **and clears it on success** — it is a global, so without the reset one rejection marks every later call (`a_rejection_does_not_leak_into_the_next_call`);
- the synchronous branch records the thrown value instead of `catch(e) { __result = undefined; }`;
- `returnByValue: true, awaitPromise: false` no longer short-circuits through `evaluate`, whose wrapper answers an exception with `null`. That was a second, independent implementation of the same swallow, on the shape Puppeteer uses most.

A thrown value is reported **by reference even when the caller asked for a value**: `JSON.stringify(new Error("boom"))` is `{}`, so serializing it would throw away the one thing the client needs.

`meta_extract_js` also grows an `Error` branch. Without it the field would be present and useless: an `Error` reported `className: "Error", description: "Error"` and no `subtype`, where Chrome gives `subtype: "error"` and a description carrying the message and stack.

### The expression is now a string, not pasted source

`Runtime.evaluate` takes a *script*, not an expression. Pasting it into `(...)` made `throw new Error('x')` a parse-time SyntaxError — which no `catch` can see, so the whole command failed — and put statement bundles in a function body with no `return`, so `var x = 1; x * 2` evaluated to the wrapper's own `undefined`.

Passing the script as a JSON string literal into `(0, eval)(…)` fixes both, and deletes two workarounds that existed only to keep the pasted form parseable:

- a trailing `;` had to be trimmed, or `(expr;)` failed to parse;
- a trailing `//# sourceURL=` comment (Puppeteer and Playwright both append one) would swallow the closing paren unless it sat on its own line.

Neither is possible against a string literal. `a_trailing_semicolon_and_source_url_still_parse` pins that the class stays gone. `(0, eval)` rather than `eval` so the script runs at global scope, where `var` lands on `globalThis` as it does in Chrome and cannot see the wrapper's `__result`.

## Validation

```
render     656 tests run: 656 passed, 3 skipped
no-render  521 tests run: 521 passed, 3 skipped
```

across `obscura-js`, `obscura-cdp` and `obscura-browser`. No existing test changed behaviour — the risk I was most concerned about, since every `Runtime.evaluate` caller goes through this path.

New tests:

- `evaluate_reports_a_rejection_through_exception_details`
- `call_function_on_reports_a_rejection_through_exception_details`
- `call_function_on_marks_a_rejection_instead_of_returning_it`
- `call_function_on_separates_a_rejected_object_from_a_resolved_one` — `Promise.reject({code:42})` must not equal `Promise.resolve({code:42})`
- `a_rejection_does_not_leak_into_the_next_call` — the global-reset case
- `a_resolved_evaluation_carries_no_exception_details`
- `a_synchronous_throw_is_reported_instead_of_swallowed`
- `a_synchronous_throw_is_reported_when_a_value_was_asked_for`
- `a_throw_statement_is_run_as_a_script_not_wrapped_in_parentheses`
- `a_statement_bundle_yields_its_completion_value`
- `a_trailing_semicolon_and_source_url_still_parse`

`test_evaluate_for_cdp_reports_promise_rejection` was rewritten: it asserted `unwrap_err()`, which pinned the bug.

## Rendering

Not applicable.

## Performance

No expected impact. The synchronous path costs one extra small script execution to read the outcome flag, on a path that already executes a script per call. Indirect eval replaces string pasting with an equivalent parse.

## Checklist

## Deliberately not in scope

- **`wrap_expression`** keeps its `catch → null`. Its callers are `evaluate` and `evaluate_with_timeout`, internal helpers a lot of engine code depends on for that contract; #746 is about the CDP surface.
- **The RemoteObject shape for primitives.** While measuring this I found `value` is a JSON string for every primitive on the handle path (`1 + 1` reports `value: "2"`), `undefined` becomes `{type: "object", subtype: "null"}` under `returnByValue`, and integers describe as `"2.0"`. All predate this PR and are unchanged by it; filed separately since they touch every primitive rather than the error path.

Thanks to @yinnho, whose comments on #746 identified the synchronous-path asymmetry and the indirect-eval approach from an engine that already reports this shape.

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -3650,6 +3650,7 @@ impl Page {
                 Err(e) => {
                     tracing::debug!("evaluate_for_cdp error: {}", e);
                     obscura_js::runtime::RemoteObjectInfo {
+                        thrown: false,
                         js_type: "undefined".into(),
                         subtype: None,
                         class_name: String::new(),
@@ -3662,6 +3663,7 @@ impl Page {
         } else {
             let val = self.evaluate(expression);
             obscura_js::runtime::RemoteObjectInfo {
+                thrown: false,
                 js_type: match &val {
                     serde_json::Value::String(_) => "string".into(),
                     serde_json::Value::Number(_) => "number".into(),
@@ -3695,6 +3697,7 @@ impl Page {
         } else {
             let value = self.evaluate(expression);
             Ok(obscura_js::runtime::RemoteObjectInfo {
+                thrown: false,
                 js_type: match &value {
                     serde_json::Value::String(_) => "string".into(),
                     serde_json::Value::Number(_) => "number".into(),
@@ -3733,6 +3736,7 @@ impl Page {
                 Err(e) => {
                     tracing::debug!("callFunctionOn error: {}", e);
                     obscura_js::runtime::RemoteObjectInfo {
+                        thrown: false,
                         js_type: "undefined".into(),
                         subtype: None,
                         class_name: String::new(),
@@ -3744,6 +3748,7 @@ impl Page {
             }
         } else {
             obscura_js::runtime::RemoteObjectInfo {
+                thrown: false,
                 js_type: "undefined".into(),
                 subtype: None,
                 class_name: String::new(),
--- a/crates/obscura-cdp/src/domains/runtime.rs
+++ b/crates/obscura-cdp/src/domains/runtime.rs
@@ -149,7 +149,7 @@ pub async fn handle(
             };
             emit_post_eval_nav(ctx, session_id).await?;
 
-            Ok(json!({ "result": remote_object_from_info(&info) }))
+            Ok(evaluation_reply(&info))
         }
         "callFunctionOn" => {
             let function_declaration = params
@@ -213,7 +213,7 @@ pub async fn handle(
             };
             emit_post_eval_nav(ctx, session_id).await?;
 
-            Ok(json!({ "result": remote_object_from_info(&info) }))
+            Ok(evaluation_reply(&info))
         }
         "getProperties" => {
             // Puppeteer's $$() flow:
@@ -455,6 +455,33 @@ fn validate_context_id(
     Ok(())
 }
 
+/// Shape one `Runtime.evaluate` or `Runtime.callFunctionOn` reply.
+///
+/// A thrown value, or the value a promise rejected with, is not a protocol
+/// failure: the command succeeds and reports it through `exceptionDetails`,
+/// which is what a client rebuilds the page error from. The value is repeated
+/// in `result` the way Chrome does, so a client that reads only `result` sees
+/// the error object rather than a success that never happened.
+///
+/// `exceptionId` is a fixed 1 because nothing here correlates exceptions
+/// across commands; clients key off the presence of the field, not its id.
+fn evaluation_reply(info: &RemoteObjectInfo) -> Value {
+    let remote = remote_object_from_info(info);
+    if !info.thrown {
+        return json!({ "result": remote });
+    }
+    json!({
+        "result": remote.clone(),
+        "exceptionDetails": {
+            "exceptionId": 1,
+            "text": "Uncaught",
+            "lineNumber": 0,
+            "columnNumber": 0,
+            "exception": remote,
+        }
+    })
+}
+
 fn remote_object_from_info(info: &RemoteObjectInfo) -> Value {
     let mut obj = json!({ "type": info.js_type });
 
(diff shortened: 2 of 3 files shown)


ntdatt812:
Obstacle course, for the CONTRIBUTING pre-PR item a unit test cannot cover. Companion repo at the ref your CI pins (`6ebac82`), same host, base and candidate back to back:

| --- | --- | --- |
| this branch | 32/33 | 3020.1ms |

No delta. The one failure is `observer-intersection`, which fails identically on untouched `main` here and is already tracked as #671, so it is not this branch. It is deterministic rather than flaky on this host: re-running that stage alone with `--runs 3 --warmup 1` reproduces it exactly.

yinnho:
Read the whole diff — this matches the shape we reported from our engine in #746, and the "one outcome protocol" (both halves write the same two globals so the reply builder stays path-agnostic) is cleaner than keeping the rejection check inside the `await_promise` branch. Two notes and one question:

**The switch to `(0, eval)(src)` deserves its own callout.** We arrived at the same change from a different direction: byte-WAF challenge scripts (the juejin.cn class) are statement-level JS — `try{}catch{}` blocks, mixed statements — and pasting them into an expression wrapper was unfixable; indirect eval was the only shape that worked. Your diff kills three birds at once: statements become legal input, the trailing-semicolon trim goes away, and a trailing `//# sourceURL=` comment can no longer swallow the closing paren. That third one is nasty to debug, for the record — a parse-time SyntaxError inside generated code surfaces as a generic failure far away from the cause.

**Question — `callFunctionOn` with `awaitPromise: false`.** In `evaluate`, every shape now writes `__obscura_await_rejected` before the now-unconditional read: the sync wrapper writes `false` on success and `true` in `catch`, the await wrapper does the same. For `callFunctionOn` I can only find the flag write in the *await* wrapper (`__result = await __fn.call(...)`). If a non-await call reaches the same unconditional `readRejected` check without a wrapper run in between, a `true` left over by a previous command turns a clean result into a phantom `exceptionDetails` — the stale-global variant of the exact bug the reset-on-success discipline is there to prevent. Two cheap fixes if that path exists: clear the flag at the top of the command before the wrapper runs, or route the non-await call through a wrapper that writes it. If the non-await flow never reaches the read, worth a one-line comment saying so — the next reader will have the same question.

**Minor, optional:** `lineNumber`/`columnNumber` are hardcoded 0. Chrome fills them from the throw site and populates `exceptionDetails.stackTrace.callFrames`. Nothing client-side consumes them today (Playwright/Puppeteer read `exception.description`), but if you want them later they are nearly free: the `instanceof Error` meta branch already has `v.stack`, and the first `at <file>:<line>:<col>` frame parses out with one regex.

One detail worth keeping as-is: the object-store insert in `thrown_info`. Thrown values are reported by reference even under `returnByValue` (right call — `JSON.stringify(new Error(...))` is `{}`), so they participate in `Runtime.releaseObject` bookkeeping like any other objectId rather than leaking. And flipping `test_evaluate_for_cdp_reports_promise_rejection` is the right move — the old expectation *was* the defect, and the comment you left on it says exactly why.

SGavrl (approved):

Issue #662, opened 2026-08-14
prepare_screenshot_resources fetches CSS resources from the stylesheet text instead of the render tree


aech:
## The Problem

`prepare_screenshot_resources` (`crates/obscura-browser/src/page.rs:2772`) collects the resources to load with `css_resource_urls(&css, &base_url)` (`page.rs:2824`). The source is the **text** of the stylesheet, not the render tree. A rule that matches no element on the page is fetched anyway.

For a complete screenshot this is understandable. But it has an effect that goes beyond rendering: where such an address changes server state, the fetch destroys it.

This is not a spec violation. CSS Images 3 and CSS Values 4 do not prescribe the timing of a load. The argument is the consistent behavior of two engines, and the effect.

## Measured

Login page of an SAP AppRouter, reachable without logging in, so reproducible without access: `https://sappubliccatalog.authentication.eu10.hana.ondemand.com/login`

| | protected background images | font-awesome fonts | total fonts |
|---|---|---|---|
| Chromium 151 | 0 | 0 | 4 |
| Firefox 150 | 0 | 0 | 4 |
| obscura | 4 | 6 | 18 |

The four images are in `application.css` and match no rendered element:

```
url('/images/gray-cloud.png')   url('/images/logo-pcf.png')
url('/images/logo-pws.png')     url('/images/p-logo.png')
```

All four respond with `302` to `/login`. Cross-checked with curl.

## Why this is more than unnecessary traffic

Every call to `/login` mints a new OAuth `state` into the session and delivers it in the page. The four image requests go via `302` to `/login` and thereby overwrite the `state` the page has just received. The callback fails with `state_not_verified`, the round starts over, and obscura aborts with `Too many redirects`.

Fetching a resource nobody displays therefore makes the login impossible.

## Reproduction without SAP

Four endpoints are enough. The page carries a stylesheet with a rule that no element uses, and whose `background-image` points to a protected address.

```
/login             mints a new state on EVERY GET, serves HTML with
                   <link rel="stylesheet"> and the link to /authorize?state=...
/style.css         .not-used-by-any-element { background-image: url('/protected/bg.png'); }
/protected/bg.png  302 to /login
/callback          compares the state with the most recently minted one
```

**Expected:** one `/login`, zero image fetches, the callback succeeds.
**Observed with obscura:** 20 logins, 10 image fetches, `state_not_verified`.

Chromium and Firefox report `LOGIN OK`.

## Two things that are relevant

[...]

PR #748, opened 2026-08-29
fix(browser): warm only the font sources the renderer will consider

### LINKED ISSUE #667: The screenshot warmup fetches every entry in a font-face list, not one
## The problem

`prepare_screenshot_resources` (`crates/obscura-browser/src/page.rs:3168`) collects the resources to load with `css_resource_urls` (`page.rs:3220`). This function returns every `url()` found in the stylesheet, including those inside a `@font-face` rule. A rule that offers the same font in six formats produces six downloads.

But the `src` list of a `@font-face` is not a set of resources. It is a priority order, and an engine loads one of its entries, not all of them.

All references against `8ceac62`.

## Measured

Both engines over the same path, raw CDP and `Network.requestWillBeSent`. No Playwright on one side and CDP on the other, otherwise the comparison measures the tools instead of the engines.

### A page that can be rebuilt without access

A stylesheet with the common IE8 form, that is, two `src` declarations, the first with the bare `.eot`, the second with the actual list:

```css
@font-face {
  font-family: 'Probe';
  src: url('/font/probe.eot');
  src: url('/font/probe.eot?#iefix') format('embedded-opentype'),
       url('/font/probe.woff2') format('woff2'),
       url('/font/probe.woff') format('woff'),
       url('/font/probe.ttf') format('truetype'),
       url('/font/probe.svg#probe') format('svg');
}
```

The served file is deliberately not a valid font format. This makes every activation fail, and a conformant engine keeps walking the list. This separates the two questions: what an engine never fetches in the first place because of the hint, and where it stops when something succeeds.

| | Font requests |
|---|---|
| Chromium 151.0.7922.34 | 3: `.woff2`, `.woff`, `.ttf` |
| obscura | 6: `.eot`, `.eot`, `.svg`, `.ttf`, `.woff`, `.woff2` |

The difference is exactly `.eot` and `.svg`. Chromium never fetches them because their `format()` hint names a format it does not support. That obscura does not support them either is written in its own code: `font_source_may_be_supported` (`crates/obscura-render/src/paint.rs:7009`) rejects exactly these two, and `fetch_and_decode_font` (`paint.rs:7030`) accepts only `wOF2`, `wOFF`, sfnt and collections.

The `.eot` appears twice because it is in both `src` declarations, once bare and once with `?#iefix`. Those are two different addresses, so no deduplication merges them. That is the file, not a second defect.

### A real page

The login page of an SAP Approuter, reachable without login: `https://sappubliccatalog.authentication.eu10.hana.ondemand.com/login`

| | Fonts | Families | Formats per family |
|---|---|---|---|
| Chromium 151 | 4 | 4 | 1, all `.woff2` |
| obscura | 14 | 7 | 2, `.woff` **and** `.woff2` |

Seven families, each fetched twice. Measured on 2026-08-14.

The difference in the number of families does not belong here. It is the subject of #662: three of the seven families match no rendered element. Both findings act on the same function independently of each other.

## What the standard says

CSS Fonts 4 says it in two places, and the two differ in strength.

§4.3, the `src` descriptor, describes the model:

> Its value is a prioritized, comma-separated list of external references or locally-installed font face names. When a font is needed the user agent iterates over the set of references listed, using the first one it can successfully parse and activate.

§4.3.3 turns this into an obligation, but only for part of it:

> External references consist of a URL, followed by an optional hint describing the format of the font resource referenced by that URL. Conformant user agents must skip downloading a font resource if the format hint indicates an unsupported or unknown font format, or if any of the font technologies are unsupported by the user agent. If no format hint is supplied, the user agent should download the font resource.

https://drafts.csswg.org/css-fonts-4/#src-desc

Two different statements about obscura follow from this.

**Fetching `.eot` and `.svg` violates a MUST.** Their hint names a format for which no decoder exists.

**Fetching `.woff` alongside `.woff2` violates no MUST.** obscura supports both formats, so the wording does not apply. Here the behavior only contradicts the model from §4.3 and the matching behavior of Chromium. It doubles the font traffic without breaking a single sentence of the spec.

## Why I am not attaching a pull request here

I built the obvious solution and then discarded it. It has a side effect I did not see while drafting.

Today, after the warmup, **every** format is in the cache. If the first choice fails during decoding, the renderer finds the next one there. If the warmup fetches only one entry, this fallback is empty, and the capture path cannot load it after the fact. It is explicitly built that way:

> Capture callers disable it so a screenshot observes only bytes already prepared by the page transport.
> (`paint.rs:193`, on `set_sync_loading_enabled`)

A `404` on the `.woff2` then means: no font in the image at all. Before, it was there.

A complete fix would have to give the warmup itself a fallback semantics, that is, try in order and stop only on success. That is an architecture decision on your code, and not one I should make from the outside.

There was a second reason. My first draft took the `@font-face` blocks out of the general scan and evaluated them separately. `css_resource_urls` handles comments and strings correctly. My new path did not. On the canonical output of Font Squirrel, that is, exactly the form shown above with its comments, it returned zero addresses instead of six. A single trailing `/* Legacy iOS */` was enough. Whoever takes this on should do it **inside** `css_resource_urls`, where these guarantees already live.

## What I can offer

For the MUST half I have a scoped patch in mind: inside `css_resource_urls`, skip an entry whose `format()` hint names a format for which no decoder exists. This removes only `.eot` and `.svg`, leaves every decodable format in the cache, and touches neither the priority order nor the cascade. This preserves the fallback.

Let me know if you want this as a PR.

I can also contribute the fixture if it is useful to you. It consists of a small server for the page above and a CDP client that counts the requests.

## What I have not checked

`tech()` from the same part of the spec. An entry with a font technology that obscura does not support would also have to be skipped. I have not measured which technologies the renderer supports, and I did not want to propose a rule that removes fonts that work today.

Also open: the cascade within the block. A later `src` declaration replaces an earlier one, but a declaration without a single usable entry is a parse error and is discarded, so the earlier one survives. Whoever implements the priority order needs this rule as well.

ntdatt812:
Fixes #667.

## What changed

`prepare_screenshot_resources` collected its work with `css_resource_urls`, which returns every `url(` in the stylesheet text. A `@font-face` `src` list is a priority order, not a set of resources, so the IE8 form in the report produced six downloads for one font.

The useful part of this is that **the engine already resolves that list correctly. Only the warmup did not.** `obscura-render` builds its font rules as:

```rust
let sources: Vec<_> = font_face_urls(face)
    .filter(|src| font_source_may_be_supported(src))
```

and those two encode exactly the two rules the warmup was missing:

* `font_face_declaration`, which `font_face_urls` reads `src` through, ends in `.last()`. A rule carrying several `src` descriptors uses the final one. That is the cascade, and it is what the `src: url(.eot); src: url(...)` idiom relies on. `font_face_uses_the_last_duplicate_src_descriptor` already pins it.
* `font_source_may_be_supported` drops `.eot` and `.svg` after stripping the query and fragment, so `probe.eot?#iefix` and `probe.svg#probe` go too.

So on the reported fixture the renderer considers three candidates and the warmup fetched six, including the two the renderer had already ruled out.

`css_font_face_rule` applies both rules and reports the block as consumed. That is the same shape as the `@import` skip a few lines above it in the same scanner, which exists for the same reason: the construct belongs to a different fetch path, so scanning it generically issues the wrong requests.

The `url(...)` recording moved into `push_css_url`, shared by the generic scan and the font path, so the two cannot drift apart on quoting, fragments, `data:` or an unresolved `var()`.

An unterminated block returns `None` and falls back to the generic scanner, matching `css_import_rule_len`: warming too much is better than dropping the rest of the stylesheet.

<details>
<summary><b>Why three and not one, which is what Chrome fetches</b></summary>

Chrome fetches one because it stops at the first candidate it can use. Here `fetch_and_decode_font` walks the list and tries the next when a decode fails, which is what makes `font_face_uses_the_first_decodable_source` pass. Warming only the first would leave that fallback fetching cold on exactly the pages that need it, so this PR narrows the warmup to what the renderer will consider, not to what it will end up using. Going to one is a change to the decode strategy rather than to the warmup, and belongs in its own change if you want it.

This also does not touch #662, the other defect in the same function: collecting from the stylesheet text rather than from the render tree. That one is larger and independent, and this does not block it.
</details>

## Validation

Four tests next to the existing `css_resource_discovery_ignores_strings_comments_data_and_fragments`, which still passes unchanged.

The first is the reported fixture. Red before the fix, and the red is the report's own number:

```
left:  ["…/probe.eot", "…/probe.eot?", "…/probe.woff2", "…/probe.woff", "…/probe.ttf", "…/probe.svg"]
right: ["…/probe.woff2", "…/probe.woff", "…/probe.ttf"]
```

The other three cover `local()` and `data:` sources, a `}` inside a string in the block, and an unterminated block. Two of them pass on `main` as well, because the generic scanner already ignores `local(` and `data:`; they are here to pin the block-skip boundary, which is new code and would silently eat the rule after the block if it were wrong.

```
5 tests run: 5 passed
```

Full four-configuration sequence from CONTRIBUTING, in a `rust:latest` container so the result matches CI rather than my Windows host:

| Step | Result |
| --- | --- |
| `cargo nextest run --release --features render --no-fail-fast` | **1510 passed**, 4 skipped |
| `cargo nextest run --release --workspace --exclude obscura --exclude obscura-render --no-default-features --no-fail-fast` | **761 passed**, 3 skipped |

I have not run `cargo fmt` over the tree, per CONTRIBUTING. `cargo fmt --check -p obscura-browser` reports 33 pre-existing diffs, at `page.rs` lines 1178 through 6936 and in `context.rs` and `pdf.rs`; this PR touches 530-781 and 4392-4474, so none of them are mine.

## Rendering

No output change, and no fixture is needed: this removes requests for bytes the renderer never decoded. The three sources it does warm are the three `paint.rs` would have selected from the same rule, so the font that ends up used is the same one. The render-feature suite is unchanged at 1510 passing, including the `@font-face` selection tests in `obscura-render` that own this behaviour.

## Performance

This is a reduction. On the reported fixture the screenshot warmup issues three requests where it issued six, and the two `.eot` fetches it drops are the ones most likely to be large and useless. Nothing is added to a hot path: the new scan runs once per stylesheet, over a block the previous code was already walking character by character, and it walks it once rather than leaving every `url(` to the generic loop.

The saving is bounded by how many sources a page's `@font-face` rules declare, so a page using a single `woff2` sees no change at all.

## Checklist

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -527,6 +527,16 @@ fn css_resource_urls(css: &str, base: &url::Url) -> Vec<String> {
             index += length;
             continue;
         }
+        // A `@font-face` `src` list is a priority order, not a set of
+        // resources. Scanning it generically warms every entry of every
+        // descriptor; the renderer considers far fewer. `css_font_face_rule`
+        // collects the ones it will consider and reports the whole block as
+        // consumed, the same shape as the `@import` skip above.
+        if let Some((length, sources)) = css_font_face_rule(rest, base) {
+            urls.extend(sources);
+            index += length;
+            continue;
+        }
         let Some(first) = rest.chars().next() else {
             break;
         };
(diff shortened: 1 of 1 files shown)


ntdatt812:
Obstacle course, for the CONTRIBUTING pre-PR item a unit test cannot cover. Companion repo at the ref your CI pins (`6ebac82`), same host, base and candidate back to back:

| --- | --- | --- |
| this branch | 32/33 | 3020.5ms |

No delta. The one failure is `observer-intersection`, which fails identically on untouched `main` here and is already tracked as #671, so it is not this branch. It is deterministic rather than flaky on this host: re-running that stage alone with `--runs 3 --warmup 1` reproduces it exactly.

SGavrl (approved):

PR #776, opened 2026-08-30
fix(js): enter the isolate per operation instead of for its lifetime

### LINKED ISSUE #756: Dropping a Page aborts the process when a second isolate is live (heap->isolate() == Isolate::TryGetCurrent())
### Before you start

- [x] I searched existing issues. This is the same fatal as #19, #258 and #430, all closed — see below.
- [x] I tested against `main`.

### obscura version or commit

`9326880` (`main`, clean checkout — no local patches)

### OS and architecture

macOS arm64

Embedded Rust library.

### Affected area

Isolate lifecycle / `Page` teardown.

### Build variant

Render-enabled.

### Reproduction

Two live pages in one process, then drop either one.

```rust
use obscura::Browser;

#[tokio::test(flavor = "current_thread")]
async fn two_pages_on_one_browser_drop_cleanly() {
    std::env::set_var("OBSCURA_ALLOW_PRIVATE_NETWORK", "1");
    let base = spawn_server();                    // trivial local HTML fixture

    let browser = Browser::new().unwrap();
    let mut one = browser.new_page().await.unwrap();
    one.goto(&base).await.unwrap();
    let mut two = browser.new_page().await.unwrap();
    two.goto(&base).await.unwrap();

    drop(one);                                     // <-- aborts here

    assert_eq!(two.evaluate("document.title"), serde_json::json!("fixture"));
}
```

No network and no CDP involved. A complete runnable test file is attached below.

### Expected behavior

Dropping one page leaves the others usable, as it does with a single page.

### Actual behavior

```
# Fatal error in , line 0
...
  5  deno_core::runtime::jsrealm::JsRealmInner::destroy
  6  <deno_core::runtime::jsruntime::InnerIsolateState as core::ops::drop::Drop>::drop
  7  core::ptr::drop_in_place<core::option::Option<obscura_js::runtime::ObscuraJsRuntime>>
  8  core::ptr::drop_in_place<obscura_browser::page::Page>
```

SIGTRAP, exit 133.

The boundary is exactly **two live isolates**, not two browsers:

| case | result |
|---|---|
| one browser, one page | ok |
| one browser, **two pages** | **abort** |
| two browsers, one page each | **abort** |

Each `Page` owns its own `ObscuraJsRuntime` (`page.rs:226`), so N pages means N
isolates on the thread. The realm is destroyed in `Drop` without entering its
own isolate first, so V8's check fires whenever a different isolate happens to
be current.

Not applicable — this is a host-process abort, not page behaviour.

### Relationship to #19 / #258 / #430

Same assertion, and all three are closed:

* #19 — concurrency ≥ 5, single thread + N isolates
* #258 — `tab_close`
* #430 — concurrent CDP navigations at concurrency 2

Those were all about *concurrent operations*. This one needs no concurrency at
all: two pages exist, one is dropped, and the process dies. It may be that the
earlier fixes covered the op/navigation paths and left `Drop` alone, or that one
of them regressed — I have not bisected.

### Practical impact

Any embedder holding more than one page. It effectively caps a process at a
single live page, which is a severe limit for a library whose natural use is a
pool of pages.

### Full test file

```rust
use std::io::{Read, Write};

use obscura::Browser;

fn spawn_server() -> String {
    let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
    let addr = listener.local_addr().unwrap();
    std::thread::spawn(move || {
        for incoming in listener.incoming() {
            let Ok(mut stream) = incoming else { continue };
            std::thread::spawn(move || {
                let mut buf = [0u8; 4096];
                let _ = stream.read(&mut buf);
                let body = "<!doctype html><html><head><title>fixture</title></head><body>hi</body></html>";
                let response = format!(
                    "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\nContent-Length: {}\r\nConnection: close\r\n\r\n{}",
                    body.len(), body,
                );
                let _ = stream.write_all(response.as_bytes());
                let _ = stream.shutdown(std::net::Shutdown::Both);
            });
        }
    });
    format!("http://{}", addr)
}

#[tokio::test(flavor = "current_thread")]
async fn one_browser_one_page_drops_cleanly() {
    std::env::set_var("OBSCURA_ALLOW_PRIVATE_NETWORK", "1");
    let base = spawn_server();
    let browser = Browser::new().unwrap();
    let mut page = browser.new_page().await.unwrap();
    page.goto(&base).await.unwrap();
    assert_eq!(page.evaluate("document.title"), serde_json::json!("fixture"));
}

#[tokio::test(flavor = "current_thread")]
async fn two_pages_on_one_browser_drop_cleanly() {
    std::env::set_var("OBSCURA_ALLOW_PRIVATE_NETWORK", "1");
    let base = spawn_server();
    let browser = Browser::new().unwrap();
    let mut one = browser.new_page().await.unwrap();
    one.goto(&base).await.unwrap();
    let mut two = browser.new_page().await.unwrap();
    two.goto(&base).await.unwrap();
    drop(one);
    assert_eq!(two.evaluate("document.title"), serde_json::json!("fixture"));
}

#[tokio::test(flavor = "current_thread")]
async fn a_second_browser_can_be_dropped_without_aborting() {
    std::env::set_var("OBSCURA_ALLOW_PRIVATE_NETWORK", "1");
    let base = spawn_server();
    let first = Browser::new().unwrap();
    let mut page_one = first.new_page().await.unwrap();
    page_one.goto(&base).await.unwrap();
    let second = Browser::new().unwrap();
    let mut page_two = second.new_page().await.unwrap();
    page_two.goto(&base).await.unwrap();
    drop(page_one);
    drop(first);
    assert_eq!(page_two.evaluate("document.title"), serde_json::json!("fixture"));
}
```

Run each with `--exact`, since the first abort takes the process down with it.

alexskinner:
Fixes #756.

Two live pages in one process aborted V8 on the first drop, whichever page it was:

```
```

### Why an embedder could not avoid it

rusty_v8 enters an isolate when it is constructed and exits it when it is dropped, and documents the consequence in `OwnedIsolate::drop`:

> v8::OwnedIsolate instances must be dropped in the reverse order of creation. They are entered upon creation and exited upon being dropped.

Each `Page` owns its own `ObscuraJsRuntime`, so N pages are N isolates stacked in construction order — and independent `Page` objects handed to a caller cannot honour that contract.

Entering the doomed isolate before teardown is not enough on its own: it is then entered twice, and `Isolate::Dispose` aborts in turn with *"Disposing the isolate that is entered by a thread"*.

### The larger half, found while chasing that

V8 also requires the owning isolate to be the thread's current one whenever a **context** is entered or left, and rusty_v8's scopes never arrange it — they only ever pass the isolate explicitly. Keeping an isolate entered for life made that true *by accident* for a single page, because that page's isolate is trivially the current one.

With two pages only the newest is current, so **running any script on the older page aborted with the same check, with no teardown involved at all**. A second page silently disabled the first:

```rust
let mut first = browser.new_page().await?;   first.goto(&base).await?;
let mut second = browser.new_page().await?;  second.goto(&base).await?;
second.evaluate("document.title");           // fine
first.evaluate("document.title");            // aborts the process
```

I confirmed that on a clean `main` at 9326880 before touching anything. It is the same root cause as the reported crash, so it is fixed here rather than filed separately — the issue's stated expectation ("dropping one page leaves the others usable") is not reachable without it.

### The fix

Exit the isolate once construction is done and enter it only around work that touches V8:

- `ObscuraJsRuntime::runtime()` returns a guard that enters on creation and exits on drop.
- The field is renamed to `js_runtime` so the compiler finds every path that reaches V8 and forces it through the guard. `realm_states` is the one caller that keeps the raw field — `op_state` is a Rust-side table with no V8 access.
- `Drop` enters once for the teardown that `OwnedIsolate::drop` will exit.

Entries then nest properly however many pages exist, and in whatever order they are used and dropped.

### Testing

New coverage for an older page still running script while a newer one is alive, for interleaving two pages in both directions, and for dropping the first, the last, and one out of the *middle* of three while the survivors keep working (plus creating a fresh page afterwards, so the entry stack is left in a state a new isolate can be pushed onto).

`cargo test -p obscura --features render` is green apart from the three `child_frame_scripts` failures already present on `main` at 9326880; `cargo test -p obscura-render --features paint` is fully green.

Two pre-existing things this does **not** fix, noted so they are not mistaken for fallout:
- `cargo test -p obscura-js --lib` still aborts on `# Check failed: !IsFrozen()`, unchanged from `main`.
- `cargo test -p obscura-render --lib` without `--features paint` does not compile on `main` (`no field word_ifc_items` at `dom.rs:19710` — a `#[cfg(feature = "paint")]` field used from an ungated test).

### DIFF
--- a/crates/obscura-js/src/runtime.rs
+++ b/crates/obscura-js/src/runtime.rs
@@ -128,7 +128,7 @@ pub struct RemoteObjectInfo {
 }
 
 pub struct ObscuraJsRuntime {
-    runtime: JsRuntime,
+    js_runtime: JsRuntime,
     state: Rc<RefCell<ObscuraState>>,
     object_store: HashMap<String, String>,
     object_counter: u64,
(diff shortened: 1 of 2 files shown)


alexskinner:
Closing this — #824 landed a fix for #756 from your side, so this is superseded. Rebasing it onto current main would only re-litigate ground that is already covered.

For anyone landing here from the issue: the merged fix is 6f4e6b5 (`fix(js): support independent isolate teardown`).

Issue #481, opened 2026-07-25
navigator.userAgent contradicts the User-Agent request header on every platform


vdatalog:
### Before you start

- [x] I searched existing issues and this is not a duplicate.
- [x] I tested against the latest release or `main`.
- [x] This is not a security vulnerability (those go through SECURITY.md, not a public issue).

### obscura version or commit

0.1.10 (official release binary) and e314b9b (built from source, 2026-07-24)

### OS and architecture

Windows 11 x86_64 and Ubuntu 24.04 x86_64 (WSL2) — reproduced identically on both.

### How are you using obscura?

CLI (fetch / serve / scrape / mcp)

### Stealth mode

None

### Reproduction

# what the server receives
obscura fetch https://postman-echo.com/headers --dump original --stealth

# what the page sees
obscura fetch https://example.com --stealth \
  --eval 'navigator.userAgent + " || platform=" + navigator.platform'

### Expected behavior

The `User-Agent` request header, the `Sec-CH-UA-Platform` client hint and
`navigator.userAgent` / `navigator.platform` should tell the server one consistent story,
the way a real browser does by construction.

### Actual behavior

They disagree. `navigator.*` always reports **Windows**, regardless of the host platform,
while the request header always reports the **real** platform:

| | HTTP request header | JS on the page |
|---|---|---|
| **Linux host** | `Mozilla/5.0 (X11; Linux x86_64) ... Chrome/145.0.0.0` | `Mozilla/5.0 (Windows NT 10.0; Win64; x64) ... Chrome/145.0.0.0` |
| `Sec-CH-UA-Platform` | `"Linux"` | `navigator.platform` → `Win32` |
| **Windows host** | `Mozilla/5.0 (X11; Linux x86_64) ... Chrome/145.0.0.0` | `Mozilla/5.0 (Windows NT 10.0; Win64; x64) ... Chrome/145.0.0.0` |
| `Sec-CH-UA-Platform` | `"Linux"` | `navigator.platform` → `Win32` |

So on Linux the header is right and JS is wrong; on Windows *both* are wrong, in opposite
directions. Either way the two never match, and a one-line server-side check catches it.

**Additionally, without `--stealth` the Chrome version itself disagrees:**

```
header:              ... Chrome/145.0.0.0 Safari/537.36
navigator.userAgent: ... Chrome/143.0.0.0 Safari/537.36
```

With `--stealth` both report 145, so only the OS part remains mismatched in that mode.

[...]

PR #778, opened 2026-08-31
fix(cdp): honor acceptLanguage on both spellings of setUserAgentOverride

### LINKED ISSUE #777: Emulation.setUserAgentOverride is a silent no-op, and Network.setUserAgentOverride drops acceptLanguage
### Before you start

### obscura version or commit

`9326880` (main, 2026-08-30)

### OS and architecture

Reproduced in the Linux CI container (rustc 1.98, x86_64). The defect is in the dispatch layer and is not platform-specific.

CDP (Puppeteer / Playwright / raw WebSocket)

### Stealth mode

None (the code path is the same in both modes)

### Reproduction

Two unit tests against `main`, no network needed:

```rust
// crates/obscura-cdp/src/domains/network.rs
handle("setUserAgentOverride",
       &json!({ "userAgent": "UA/1.0", "acceptLanguage": "de-DE,de;q=0.9" }),
       &mut ctx, &session_id).await.unwrap();
assert_eq!(*page.http_client.accept_language.read().await, "de-DE,de;q=0.9");

// crates/obscura-cdp/src/domains/emulation.rs
handle("setUserAgentOverride",
       &json!({ "userAgent": "UA/2.0", "acceptLanguage": "fr-FR,fr;q=0.9" }),
       &mut ctx, &session_id).await.unwrap();
assert_eq!(*page.http_client.user_agent.read().await, "UA/2.0");
```

Both fail on `9326880`:

```
FAIL domains::network::tests::set_user_agent_override_applies_accept_language
     assertion `left == right` failed: acceptLanguage must reach the client
       left: "en-US,en;q=0.9"      <- the default, untouched
      right: "de-DE,de;q=0.9"

FAIL domains::emulation::tests::emulation_set_user_agent_override_is_applied
     2 tests run: 0 passed, 2 failed
```

The `Network` case shows the locale never leaving the dispatcher; the `Emulation` case shows the whole command doing nothing, user agent included. Both calls returned `{}` — success.

### Expected behavior

`setUserAgentOverride` carries `userAgent`, `acceptLanguage` and `platform`. A client that sets a locale should see it on the wire and in the page, the way `userAgent` already works — `Network.setUserAgentOverride` writes `http_client.user_agent`, and the next navigation seeds `navigator.userAgent` from that same field, so the two cannot drift.

### Actual behavior

**1. `Emulation.setUserAgentOverride` is a silent no-op.** The `Emulation` handler ends in a catch-all:

```rust
"setTouchEmulationEnabled" => Ok(json!({})),
_ => Ok(json!({})),
```

`setUserAgentOverride` is not one of its four arms, so it falls through, replies `{}` and applies nothing — not even the user agent. This is the spelling the CDP spec lists as current; `Network.setUserAgentOverride` is the deprecated one.

**2. `Network.setUserAgentOverride` discards `acceptLanguage` and `platform`:**

```rust
"setUserAgentOverride" => {
    let ua = params.get("userAgent").and_then(|v| v.as_str()).unwrap_or("");
    if let Some(page) = ctx.get_session_page(session_id) {
        page.http_client.set_user_agent(ua).await;
    }
    Ok(json!({}))
}
```

**3. The plumbing for the locale half already exists and is wired to nothing.** `HttpClient::set_accept_language` is `pub`, and `accept_language_override_reaches_the_wire` proves it reaches the socket — but nothing in the tree calls it outside that test.

**4. `navigator.language` has no override hook**, unlike its neighbours:

```js
defGetter('userAgent',  function() { return globalThis.__obscura_ua || "…"; });
defGetter('platform',   function() { return globalThis.__obscura_platform || "Win32"; });
defGetter('language',   function() { return "en-US"; });
defGetter('languages',  function() { return ["en-US", "en"]; });
```

So a client that asks for `de-DE` is told the call succeeded and gets `en-US` on every surface — a wrong answer rather than a missing one. That is the same shape as #734, which was just fixed: there `Intl` disagreed with `navigator.language`; here the operator's stated locale disagrees with all four.

### Impact and workaround

Playwright's `newContext({ locale })` and `newContext({ userAgent })` both go through `Emulation.setUserAgentOverride`, so on obscura they are accepted and ignored. There is no workaround through CDP; `Network.setExtraHTTPHeaders` can force the header, but `navigator.language` stays `en-US` and then contradicts it.

### Notes

- I have a fix and will open a PR against this issue.
- Scoped deliberately to the locale half. `platform` is the same silent drop but lands in #481's territory, and `Intl`'s default locale cannot follow a mid-session override at all: V8 resolves the ICU default once per isolate and caches it at first `Intl` use, so a pin after isolate creation is a no-op. I verified that separately while checking #734 — moving `set_default_locale` below `JsRuntime::new` fails `intl_locale_matches_navigator_language_regardless_of_host`.

ntdatt812:
## What changed

Fixes #777.

`setUserAgentOverride` carries three fields. obscura read one of them, on the deprecated spelling of the command only:

- **`Emulation.setUserAgentOverride` applied nothing at all.** It is not one of the `Emulation` handler's four arms, so it fell to the trailing `_ => Ok(json!({}))` and was acknowledged without being applied — user agent included. This is the spelling the CDP spec lists as current.
- **`Network.setUserAgentOverride` discarded `acceptLanguage`.** It read `userAgent` and replied `{}`.

Both returned success, so a client that set a locale was told it worked and got `en-US` on every surface. That is the same shape as #734, which you fixed last week: a browser disagreeing with itself about where it is — except here it disagrees with the operator who told it.

### What made this small

The plumbing was already built and wired to nothing. `HttpClient::set_accept_language` is `pub` and `accept_language_override_reaches_the_wire` proves it reaches the socket, but no production caller existed. And the JS side already had the pattern to copy — at runtime creation `page.rs` seeds `navigator.userAgent` *from the HTTP client's user agent*, so the header and the page cannot drift:

```rust
if let Ok(ua) = self.http_client.user_agent.try_read() {
    rt.set_user_agent(&ua);
}
```

The locale now follows the same route, one line below it. That is what keeps this change off the V8 lock: the override writes only `http_client`, and the isolate reads it back at its next construction, so `setUserAgentOverride` remains a `get_session_page` / no-`JsRuntime` method.

| file | change |
|---|---|
| `network.rs` | `apply_user_agent_override` extracted; honours `acceptLanguage` |
| `emulation.rs` | the missing `setUserAgentOverride` arm, delegating to it |
| `dispatch.rs` | `Emulation.setUserAgentOverride` added to `is_v8_free_method` — same audit as the `Network` entry already there: `get_session_page` and client field writes, no `JsRuntime` |
| `runtime.rs` | `set_language`, `languages_from_accept_language`, and the two new globals in `IDENTITY_GLOBALS` |
| `bootstrap.js` | `navigator.language` / `languages` read the override; both globals added to the pre-hide list |
| `page.rs` | seeds the locale from the client at runtime creation |

Three details worth a reviewer's eye:

**Unsent fields are no longer overwritten.** `userAgent` used to be `unwrap_or("")`. That was harmless while the command carried nothing else, but now a locale-only call would blank the user agent as a side effect, so each field is applied only if the caller sent it.

**`IDENTITY_GLOBALS` grew from 7 to 9.** That array is what carries identity into child realms so "a frame must present the same identity as its parent". Adding `navigator.language` without adding it there would have made an iframe report `en-US` while the top document reported `de-DE` — a fresh instance of exactly the inconsistency this fixes.

**The generated script interpolates JSON, not hand-quoted strings.** The neighbouring setters do `replace('\'', "\\'")`; `acceptLanguage` arrives from the wire in a way a configured user agent does not, so `set_language` formats through `serde_json::Value` instead.

## Validation

`cargo nextest run --release --features render -p obscura-cdp -p obscura-js -E 'test(user_agent_override) or test(language) or test(frame_inherits)'`

Both repro tests from #777 fail on `9326880` and pass here. New tests:

- `set_user_agent_override_applies_accept_language` — Network, both fields
- `emulation_set_user_agent_override_is_applied` — Emulation, both fields
- `emulation_user_agent_override_leaves_unsent_fields_alone` — a locale-only call must not blank the user agent
- `set_language_drives_navigator_language_and_languages`
- `navigator_languages_hands_out_a_fresh_array_each_read` — the list is observable, so a page must not be able to mutate what later reads report
- `an_empty_accept_language_leaves_the_default_alone` — `""` and `"*"` are no-ops
- `a_language_tag_cannot_escape_the_generated_script` — feeds `de-DE','x');globalThis.__pwned=1;//` and asserts nothing ran
- `accept_language_parsing_drops_quality_values_and_wildcards`

`frame_inherits_the_parent_browser_identity` — the existing test for the invariant `IDENTITY_GLOBALS` serves — now seeds a non-default locale on the parent and checks `navigator.language` and `navigator.languages` alongside the user agent and platform it already checked. It asserts the value as well as the agreement, so it cannot pass by both realms falling back to `en-US`.

Full `obscura-cdp`, `obscura-js` and `obscura-browser` suites run in both the render and no-render configurations; results in a comment below.

## Rendering

Not applicable. No layout, paint, screenshot, screencast or PDF output changes.

## Performance

No expected impact. The added work is one `Accept-Language` parse and one small script execution per `JsRuntime` construction, on the same path that already runs `set_user_agent` and `set_platform`. The override handlers themselves stay off the V8 lock, so no request-path latency changes.

## Checklist

## Deliberately not in scope

- **`platform`**, the third field, is dropped the same way. It belongs to #481's territory (`navigator.platform` vs the header), which is under active discussion with two conflicting measurements, so I left it alone rather than pre-empting that thread.
- **`Intl`.** It cannot follow a mid-session override: V8 resolves the ICU default once per isolate and caches it at first `Intl` use, so a pin after `JsRuntime::new` is a silent no-op. I checked this against your #734 fix — moving `set_default_locale` below the isolate construction fails `intl_locale_matches_navigator_language_regardless_of_host`. Making `Intl` follow a CDP locale would mean building the isolate with the locale already known, which is a different change.

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -1447,6 +1447,15 @@ impl Page {
                 &self.context.ua_platform_version,
             );
         }
+        // navigator.language(s) come from the same Accept-Language value the
+        // client sends, exactly as navigator.userAgent is seeded from the
+        // client's user agent above. Reading it here rather than at the
+        // override means a CDP locale change reaches the page on the next
+        // navigation without this command having to touch the isolate.
+        if let Ok(accept_language) = self.http_client.accept_language.try_read() {
+            rt.set_language(&accept_language);
+        }
+
         if let Some((lat, lon)) = env_geolocation() {
             rt.set_geolocation(lat, lon);
         }
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -1452,6 +1452,20 @@ impl Page {
         // client's user agent above. Reading it here rather than at the
         // override means a CDP locale change reaches the page on the next
         // navigation without this command having to touch the isolate.
+        //
+        // Seeding here moves the two navigator globals and nothing else. ICU's
+        // default locale stays where `ObscuraJsRuntime::new` pinned it (#734):
+        // it is process-global, so re-pinning it per navigation would let one
+        // page's override decide what `Intl` resolves to in a sibling page in
+        // the same process that has not read `Intl` yet, and V8 caches that
+        // resolution per isolate at first use.
+        //
+        // The cost is that Chrome applies an override to navigator.language
+        // immediately, while a page here keeps the previous value until it
+        // navigates. That is chosen, not missed: the clients that set a locale
+        // (Playwright's `newContext({ locale })`) set it before navigating, and
+        // a mid-session re-pin could not reach an already-cached `Intl`
+        // resolution anyway.
         if let Ok(accept_language) = self.http_client.accept_language.try_read() {
             rt.set_language(&accept_language);
         }
--- a/crates/obscura-cdp/src/dispatch.rs
+++ b/crates/obscura-cdp/src/dispatch.rs
@@ -402,6 +402,7 @@ fn is_v8_free_method(method: &str) -> bool {
             | "Network.setBlockedURLs"
             | "Network.setExtraHTTPHeaders"
             | "Network.setUserAgentOverride"
+            | "Emulation.setUserAgentOverride"
             | "Network.getCookies"
             | "Network.getAllCookies"
             | "Network.setCookie"
--- a/crates/obscura-cdp/src/domains/emulation.rs
+++ b/crates/obscura-cdp/src/domains/emulation.rs
@@ -124,6 +124,14 @@ pub async fn handle(
         // Touch emulation does not affect layout yet, but acknowledging it is
         // compatible with clients that pair it with a metrics override.
         "setTouchEmulationEnabled" => Ok(json!({})),
+        // The current spelling of setUserAgentOverride; Network's is the
+        // deprecated one. Without this arm it reached the catch-all below and
+        // was acknowledged without being applied, so a client that set a user
+        // agent or a locale here was told it worked and got neither.
+        "setUserAgentOverride" => {
+            super::network::apply_user_agent_override(params, ctx, session_id).await;
+            Ok(json!({}))
+        }
         _ => Ok(json!({})),
     }
 }
@@ -131,6 +139,72 @@ pub async fn handle(
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[tokio::test]
+    async fn emulation_set_user_agent_override_is_applied() {
+        let mut ctx = CdpContext::new();
+        let page_id = ctx.create_page();
+        let session_id = Some("emu-ua-session".to_string());
+        ctx.sessions.insert(session_id.clone().unwrap(), page_id);
+
+        handle(
+            "setUserAgentOverride",
+            &json!({ "userAgent": "UA/2.0", "acceptLanguage": "fr-FR,fr;q=0.9" }),
+            &mut ctx,
+            &session_id,
+        )
+        .await
+        .expect("Emulation.setUserAgentOverride must succeed");
+
+        let page = ctx.get_session_page(&session_id).expect("page");
+        assert_eq!(
+            *page.http_client.user_agent.read().await,
+            "UA/2.0",
+            "Emulation.setUserAgentOverride must apply the user agent"
+        );
+        assert_eq!(
+            *page.http_client.accept_language.read().await,
+            "fr-FR,fr;q=0.9",
+            "Emulation.setUserAgentOverride must apply the locale"
+        );
+    }
+
+    #[tokio::test]
+    async fn emulation_user_agent_override_leaves_unsent_fields_alone() {
+        // A locale-only call must not blank the user agent, which is what the
+        // previous unwrap_or("") would have done once this command started
+        // carrying more than one field.
+        let mut ctx = CdpContext::new();
+        let page_id = ctx.create_page();
+        let session_id = Some("emu-partial".to_string());
+        ctx.sessions.insert(session_id.clone().unwrap(), page_id);
+
+        let before = ctx
+            .get_session_page(&session_id)
+            .expect("page")
+            .http_client
+            .user_agent
+            .read()
+            .await
+            .clone();
+
+        handle(
+            "setUserAgentOverride",
+            &json!({ "acceptLanguage": "ja-JP" }),
+            &mut ctx,
+            &session_id,
+        )
+        .await
+        .expect("a locale-only override must succeed");
+
+        let page = ctx.get_session_page(&session_id).expect("page");
+        assert_eq!(
+            *page.http_client.user_agent.read().await,
+            before,
+            "an unsent userAgent must be left as it was"
+        );
+        assert_eq!(*page.http_client.accept_language.read().await, "ja-JP");
+    }
 
     #[tokio::test]
     async fn device_metrics_override_updates_page_and_window_viewport() {
--- a/crates/obscura-cdp/src/domains/network.rs
+++ b/crates/obscura-cdp/src/domains/network.rs
@@ -24,6 +24,34 @@ fn cookie_jar_for<'a>(ctx: &'a CdpContext, session_id: &Option<String>) -> &'a A
         .unwrap_or(&ctx.default_context.cookie_jar)
 }
 
+/// Apply a `setUserAgentOverride` call to a session's HTTP client.
+///
+/// Chrome exposes this command on both `Network` (deprecated) and
+/// `Emulation` (current), and clients are split across the two spellings,
+/// so both dispatch here rather than each growing its own copy.
+///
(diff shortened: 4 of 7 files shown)


ntdatt812:
Full suite results referenced in the PR body, run in a Linux container on rustc 1.98:

```
render     (--features render)  655 tests run: 655 passed, 3 skipped
no-render  (default features)   520 tests run: 520 passed, 3 skipped
```

across `obscura-cdp`, `obscura-js` and `obscura-browser`.

`rustfmt --check` is clean across every line range this PR touches in all six Rust files. The tree carries unrelated pre-existing formatting diffs elsewhere, so I compared reported diff lines against the changed hunks rather than reading a whole-file pass or fail.

`cargo clippy --all-targets` reports one error, `never_loop` at `crates/obscura-js/src/import_map.rs:235` — a file this PR does not touch, and it reproduces on `main`. Nothing in the changed files.

yinnho:
Clean shape — one shared `apply_user_agent_override` behind both spellings, Emulation delegating to Network, and the only-apply-sent-fields discipline (the locale-only call blanking the UA is exactly the bug the old `unwrap_or("")` was waiting to become). The tests pin the right contracts.

One hazard worth checking, from the same per-isolate territory as #753: if `rt.set_language` also re-pins the ICU default locale (not just the JS global), then per-navigation seeding moves a **process-global** default — `set_default_locale` is not per-isolate. Combined with the sticky per-isolate cache (V8 resolves and caches the default at first `Intl` use), the interleaving is: page B is created, seeded `zh-CN`, and hasn't touched `Intl` yet; page A overrides `acceptLanguage` to `fr-FR` and navigates, moving the process default; then B's first `Intl` call resolves against `fr-FR` and caches it permanently. A pool of pages in one process — the embedder model from #756 — makes that window easy to hit. The invariant that avoids it: pin ICU once, at isolate creation, from *that page's own* client value; let navigation-time seeding move only the `navigator.language(s)` global and never re-pin ICU. (Mid-session re-pinning can't work anyway — your own #777 note on the sticky cache is the reason.)

Separately: Chrome updates `navigator.language` immediately on the override, while this seeds on next navigation. That's the right trade for the main client (Playwright's `newContext({ locale })` always overrides before navigating), but a client that changes locale mid-session sees a stale `navigator.language` until it navigates — worth a code comment so the next reader knows it's chosen, not missed.

ntdatt812:
Thanks — that is the right hazard to raise, and the invariant you name is the one this PR already holds. It does not re-pin ICU; `set_language` moves two JS globals and nothing else.

Checked rather than asserted, on `ebf42c5`:

- `set_default_locale` appears exactly once in the tree, at `crates/obscura-js/src/runtime.rs:330` — the `"en-US"` pin from #734, inside `ISOLATE_CREATE_LOCK` at isolate creation.
- The full 463-line diff of this PR contains no occurrence of `set_default_locale` or `icu::`. Its `runtime.rs` hunks are at lines 93, 549, 936 and 3193, so line 330 is not in any of them.
- `set_language` executes one runtime script that assigns `globalThis.__obscura_language` and `globalThis.__obscura_languages`; `bootstrap.js` reads those for `navigator.language(s)`. Nothing in that path reaches ICU.

So the interleaving you describe — page B seeded `zh-CN`, page A overriding to `fr-FR` and moving the process default before B's first `Intl` use — cannot happen through this change: `Intl` in both pages resolves against the same `en-US` that was pinned before either isolate existed. Per-page ICU would need the pin to move to isolate creation from that page's own client value, which is a separate change to the #734 pin and, as you say, the sticky per-isolate cache is why it has to happen there rather than mid-session.

Your second point was a real gap, and I have pushed a comment for it. The divergence is now stated where the seeding happens, along with why ICU deliberately stays put:

```rust
// Seeding here moves the two navigator globals and nothing else. ICU's
// default locale stays where `ObscuraJsRuntime::new` pinned it (#734):
// it is process-global, so re-pinning it per navigation would let one
// page's override decide what `Intl` resolves to in a sibling page in
// the same process that has not read `Intl` yet, and V8 caches that
// resolution per isolate at first use.
//
// The cost is that Chrome applies an override to navigator.language
// immediately, while a page here keeps the previous value until it
// navigates. That is chosen, not missed: the clients that set a locale
// (Playwright's `newContext({ locale })`) set it before navigating, and
// a mid-session re-pin could not reach an already-cached `Intl`
// resolution anyway.
```

The push is comment-only, so no behaviour moves with it. `cargo fmt -p obscura-browser -- --check` reports the same 23 pre-existing `page.rs` diffs as the unmodified tree and none in the touched range, which also confirms the file still parses. I could not complete a native `cargo check` on this Windows host — `v8-137.3.0/build.rs` panics at `Failed to create symlink` before it reaches the crate — so I am leaving the compile proof to CI on this push rather than claiming a local build I did not get.

PR #781, opened 2026-08-31
Fix <body onload> content attribute reflection

### LINKED ISSUE #768: `<body onload="...">` content attribute never fires — only addEventListener('load', ...) works
### Before you start

### obscura version or commit

0.1.11 (official binary)

### OS and architecture

Linux x86_64 (Debian 12, kernel 6.12.57)

### Affected area

JavaScript or DOM behavior

### Build variant

No-render with stealth (official no-render-stealth archive or --features stealth)

### Stealth mode

Not applicable

### Reproduction

```
$ obscura fetch "data:text/html,<html><body onload=%22window.__x=1%3B%22><p>hi</p></body></html>" --wait 3 --eval "(function(){ return 'x=' + window.__x; })()"
Fetching data:text/html,...
Page loaded: data:text/html,... - ""
x=undefined
```
Contrast with the equivalent `addEventListener` form, which works correctly:
```
$ obscura fetch "data:text/html,<html><body><script>window.addEventListener('load',function(){window.__x=1})</script></body></html>" --wait 3 --eval "(function(){ return 'x=' + window.__x; })()"
Fetching data:text/html,...
Page loaded: data:text/html,... - ""
x=1
```

### Expected behavior

Per the HTML standard, the `onload` content attribute on `<body>` (and `<frameset>`) is one of the "window-reflecting body element event handlers" — it's compiled and exposed as `window.onload`, and fires when the `load` event reaches the Window. `window.__x` should be `1` in the first example, the same as it is in the second.

### Actual behavior

The `<body onload="...">` handler never runs at all: `window.__x` stays `undefined`, and both `document.body.onload` and `window.onload` read back as `null` after the page finishes loading (`document.readyState === "complete"`), confirming the attribute is never compiled into a handler or forwarded to `window`.

### Impact

This is a real-world compatibility gap, not an edge case: `<body onload="...">` is a very old but still common pattern, notably used by several commercial anti-bot / WAF JS challenges (e.g. F5/TrafficShield-style "TS" cookie challenges) to kick off their proof-of-work script on page load. Because the handler never runs, the challenge's redirect/cookie flow never completes and the real page content is never reached — e.g. `obscura fetch https://cgv.vn` (with `--stealth`, to get past the separate TLS issue in #<issue1 number once filed>) hangs on the raw challenge HTML forever, while `lightpanda fetch https://cgv.vn` completes the same challenge and returns the real page. Once `window.onload` is wired up correctly, the rest of that flow (the challenge script's `document.forms[0].submit()` call) already works fine in obscura — verified separately with a local repro where `form.submit()` from a plain `<script>` tag (not gated behind `onload`) does correctly navigate.

### Logs

```shell
N/A — the eval output above shows the failure directly.
```

lisa0314:
## Summary

Fixes #768.

The HTML `<body onload="...">` content attribute is a Window-reflecting event handler, but obscura previously left it disconnected from `window.onload`. This patch:

- reflects the active document body/frameset `onload` handler through `window.onload`;
- compiles inline `onload` content attributes through the existing inline-handler resolver;
- keeps detached body elements isolated;
- dispatches one shared `load` event object and invokes `window.onload` with `this === window`;
- applies the same behavior to frame realms;
- preserves load-listener delivery when an `onload` handler throws.

## Validation

- `cargo test -p obscura-browser body_onload_content_attribute_reflects_to_window --lib --features render`
- `cargo test -p obscura-js frame_body_onload_content_attribute_reflects_to_window --lib`
- Reproduction now returns `x=1` instead of `x=undefined`.

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -2528,8 +2528,13 @@ impl Page {
             let _ = js.execute_script(
                 "<load-event>",
                 "globalThis.__documentReadyState__ = 'complete';\n\
-                 if (typeof window.onload === 'function') { try { window.onload(); } catch(e) {} }\n\
-                 try { window.dispatchEvent(new Event('load', {bubbles:false,cancelable:false})); } catch(e) {}",
+                 try {\n\
+                   const loadEvent = new Event('load', {bubbles:false,cancelable:false});\n\
+                   if (typeof window.onload === 'function') {\n\
+                     try { window.onload.call(window, loadEvent); } catch(e) {}\n\
+                   }\n\
+                   try { window.dispatchEvent(loadEvent); } catch(e) {}\n\
+                 } catch(e) {}",
             );
         }
         if let Some(token) = exec_wd {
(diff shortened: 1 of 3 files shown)


yinnho:
This matches the mechanics we shipped downstream for the same hole (byte-WAF challenge pages boot from `<body onload="readygo()">`, so this path is load-bearing for us) — the shared-slot reflection and the detached-body isolation are exactly the two contracts that are easy to get wrong, and the two-directional test covers both. Three notes from our implementation, none blocking:

**1. The window-reflecting set is wider than `onload`.** Per the HTML spec, the window-targeted event handler content attributes on `body`/`frameset` are `onload`, `onunload`, `onbeforeunload`, `onmessage`, `onpagehide`, `onpageshow`, `onhashchange`, `onpopstate`, `onstorage`, and `onerror`. Forwarding only `onload` fixes the reported break, but `<body onunload="...">` still registers a dead element handler — the load path never descends into the body element, so none of the others ever fire either. We went through the classification list explicitly; forwarding a non-window attribute by mistake (say `onclick`) breaks delegated events in the other direction, so the set is worth getting right once rather than adding to per-report.

**2. `Element.prototype.onload` is defined `enumerable: false`.** WebIDL event handler attributes are enumerable — Chrome reports `Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'onclick').enumerable === true`, and your `window.onload` definition correctly uses `enumerable: true`. The element-side `false` is an inconsistency that own-property enumeration differs catch (the same differ class as the #734 follow-up: structural comparison of prototypes, not value probes).

**3. The setter collapses non-callable objects to `null`.** `window.onload = {}` in Chrome keeps the object — an `EventListener` without `handleEvent` is a no-op listener, but the getter reads back the object, not `null`. Strings do coerce to `null` in Chrome, so that half matches. Edge case, but it's the kind of thing a "set weird values, read back" probe hits.

One ordering note worth a comment in the code rather than a change: the load site invokes the attribute handler *before* `dispatchEvent`, so `addEventListener('load')` listeners always run after it. Chrome interleaves by registration order — the content attribute registers when the body tag is parsed, so a listener registered from a `<head>` script (before the body exists) fires first. For the common shape (attribute in markup, listeners later) your order matches Chrome exactly.

SGavrl (approved):

PR #782, opened 2026-08-31
test(js): pin the Chrome shape of a blob URL and its synchronous store

### LINKED ISSUE #751: URL.createObjectURL leaks blob:obscura/ URLs and silently accepts invalid input (Chrome mints blob:<origin>/<v4-uuid> and throws TypeError)
## What happens

On current `main`, `URL.createObjectURL` in `js/bootstrap.js` (the definitions sit around lines 13162/13181):

```js
const id = 'blob:obscura/' + Math.random().toString(36).substring(2);
...
return 'blob:obscura/fallback';   // falsy / non-Blob input
```

Three deviations from Chrome, all observable from inside a page:

1. **Engine self-identification.** The URL names the engine. Any page that reads back a blob URL it created — a Worker's `scriptURL`, an anchor `href`, `performance` resource entries, a CSP violation report, `new URL(blobUrl).origin` — sees the literal string `blob:obscura/`. One `startsWith('blob:obscura')` is a complete anti-bot check.
2. **URL shape.** Chrome mints `blob:https://<document-origin>/<v4-uuid>` (and `blob:null/<uuid>` on opaque origins). The id is a real UUID with version 4 and variant nibbles set. A random base36 token fails every UUID parser/validator, and the missing origin breaks code that resolves the blob URL against the page (relative fetches, postMessage targetOrigin checks).
3. **Invalid input is accepted.** Chrome throws instead of minting a URL:
   - `URL.createObjectURL()` → `TypeError: Failed to execute 'createObjectURL' on 'URL': 1 argument required, but only 0 present.`
   - `URL.createObjectURL('nope')` → `TypeError: Failed to execute 'createObjectURL' on 'URL': parameter 1 is not of type 'Blob'.`

   The current fallback hands back a well-formed-looking string, so a buggy caller finds out later, at the fetch, with a confusing error.

## Repro

Evaluate in any obscura page context:

```js
URL.createObjectURL()                 // → 'blob:obscura/fallback'  (Chrome: TypeError)
URL.createObjectURL('nope')           // → 'blob:obscura/fallback'  (Chrome: TypeError)
URL.createObjectURL(new Blob(['x']))  // → 'blob:obscura/k2j3h9…'   (Chrome: 'blob:https://<origin>/xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx')
```

## Suggested fix

- Read the origin at mint time: `new URL(location.href).origin` (falls back to `'null'` on opaque origins), prefix `blob:` + origin + `/`.
- Generate a v4 UUID: hex, dashes at 8/13/18/23, `'4'` at index 14, one of `89ab` at index 19.
- Throw the two Chrome TypeErrors above on missing/non-Blob input.

We hit the same thing in aginxbrowser (same lineage) and fixed it in yinnho/aginxbrowser@4860141 — origin-bearing prefix, v4 UUID, exact Chrome TypeError strings, plus a pinned test (`blob_urls_are_chrome_shaped_and_non_blob_input_throws`). One extra note from that work: if the blob is stored for synchronous Worker construction, `revokeObjectURL` should delete from *every* store that Worker construction consults, not just the async text store — otherwise a revoked URL still constructs a Worker in the same tick.

ntdatt812:
## What

This PR no longer changes behaviour. `main` already mints Chrome-shaped blob URLs — @963s landed that in #804 (`c9e4c9f`) while this was open, and the production half of what I had here is redundant. Rebased onto it and narrowed to the two tests that fix landed without.

## Why

`create_object_url_rejects_non_blob_input` (SEC-503 / #820) covers the rejection half well, but the accepted half only asserts the scheme:

```js
URL.createObjectURL(new Blob(['hello'])).startsWith('blob:')
```

The form #751 was filed about — `blob:obscura/<base36>` — satisfies that too, so the assertion cannot tell the fixed shape from the bug. The shape is the whole point of #751: a page that reads a blob URL back (a Worker's `scriptURL`, an anchor href, `new URL(u).origin`, a CSP report) saw the engine's name, and the base36 token failed every UUID parser.

Two paths in the merged fix also have no coverage: the source is stored *synchronously* so a Worker constructed from the URL in the same tick can read it — the async `blob.text().then()` store is what broke AWS WAF's proof-of-work worker — and `revokeObjectURL` has to remove the entry again.

## Testing

Run in a Linux container on rustc 1.98, against `main`'s implementation:

```
cargo nextest run -p obscura-js -E 'test(a_blob_url) + test(create_object_url_rejects_non_blob_input)'
    PASS obscura-js runtime::tests::a_blob_url_is_chrome_shaped
    PASS obscura-js runtime::tests::a_blob_url_is_stored_synchronously_and_revocation_removes_it
    PASS obscura-js runtime::tests::create_object_url_rejects_non_blob_input
3 tests run: 3 passed, 385 skipped
```

To show the new tests earn their place rather than restating a green build, I reintroduced the exact bug — `const id = 'blob:obscura/' + Math.random().toString(36).slice(2);` at the mint site — and re-ran the same three:

```
    FAIL obscura-js runtime::tests::a_blob_url_is_chrome_shaped
       left:  "origin": "blob:obscura/mgrloud2qak", "uuid": false, "names_the_engine": true
       right: "origin": "blob:http://example.com/", "uuid": true,  "names_the_engine": false
    PASS obscura-js runtime::tests::a_blob_url_is_stored_synchronously_and_revocation_removes_it
    PASS obscura-js runtime::tests::create_object_url_rejects_non_blob_input
3 tests run: 2 passed, 1 failed, 385 skipped
```

That is the gap: the existing test stays green while the URL names the engine again.

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -40,6 +40,7 @@
     '_urlResolveOp', '_decodeBodyWithCharset', '_utf8DecodeBytes',
     '_selectionFor', '_isConstructorCE', '_isValidCustomElementName', '_shadowRootForHost',
     '_blobPartToBytes', '_bytesToBinaryString', '_formEncode', '_hexv',
+    '_blobUrlOrigin', '_isBlobLike', '_blobUuid',
     '_commonFonts', '_isXMLDocument', '_isValidPITarget', '_isHTMLEl',
     '_nodeList', '_rngNodeLength', '_rngNodeIndex', '_rngSame', '_rngRoot',
     '_rngAncestors', '_rngOrder', '_rngCmp', '_rngCheckOffset',
@@ -7387,8 +7388,9 @@ if (typeof URL === 'undefined' || !URL.prototype || !URL.__obscura) {
     _updateSearch(qs) { this._set('search', qs ? ('?' + qs) : ''); }
     toString() { return this._c.href; }
     toJSON() { return this._c.href; }
-    static createObjectURL() { return 'blob:null/fake-' + Math.random().toString(36).slice(2); }
-    static revokeObjectURL() {}
+    // createObjectURL/revokeObjectURL are installed further down, next to the
+    // blob store they read and write. Defining stubs here only risked one of
+    // them surviving and handing a page a 'fake-' id to fingerprint. (#751)
     // WHATWG URL.parse: like the constructor but returns null instead of throwing.
     static parse(url, base) { const c = _urlParseOp(url, base); if (!c) return null; const u = Object.create(_URL.prototype); u._c = c; u._sp = null; return u; }
     static canParse(url, base) { return _urlParseOp(url, base) !== null; }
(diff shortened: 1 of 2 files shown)


963s:
right, the accepted half didn't pin the shape past the prefix — these two fill that cleanly.

PR #791, opened 2026-09-01
fix(network): preserve CDP binary response bodies

### LINKED ISSUE #394: v0.1.10 regression: XHR deadline still fires on Aviasales + Booking.com crashes with Cannot redefine property: src
## Environment
- Obscura v0.1.10 (Linux x86_64, installed from GitHub release tarball)
- Docker on Ubuntu 24.04
- `obscura serve --port 9222` (no --stealth)
- Playwright-core via CDP

## Issue 1: Aviasales XHR — #361 not fixed

#361 was closed as fixed in v0.1.10 (`Heavy SPAs get a larger script budget so they finish booting #371`), but the bug persists.

**Symptom:** `page.on('response')` captures 0 JSON/XHR responses on aviasales.by flight search. The SPA loads but never fires its search API calls.

**Logs:**
```
[intercept] MOW→DXB 2026-07-19 → https://www.aviasales.by/search/MOW1907DXB1?currency=usd
[cycle] empty MOW→DXB 2026-07-19 (0 chunks)
```

Same page works perfectly with CloakBrowser (Playwright + stealth Chromium) — captures multiple `v3.2/search/results` JSON responses within seconds.

### Automated regression test

```js
// test-aviasales-xhr.mjs
import { chromium } from 'playwright-core';

const TIMEOUT = 60_000;
const TARGET = 'https://www.aviasales.by/search/MOW1907DXB1?currency=usd';
const JSON_PATTERN = /tickets|search\/results|v3\.2/;

async function main() {
  const browser = await chromium.connectOverCDP('http://127.0.0.1:9222');
  const context = await browser.newContext();
  const page = await context.newPage();

  const jsonResponses = [];
  page.on('response', (resp) => {
    const url = resp.url();
    if (JSON_PATTERN.test(url)) {
      resp.json().then((data) => jsonResponses.push({ url, data })).catch(() => {});
    }
  });

  await page.goto(TARGET, { waitUntil: 'domcontentloaded', timeout: TIMEOUT });
  await page.waitForTimeout(45_000);
  await context.close();
  await browser.close();

  console.log('Captured ' + jsonResponses.length + ' JSON responses');
  if (jsonResponses.length === 0) {
    console.error('FAIL: captured 0 JSON responses');
    process.exit(1);
  }
  console.log('PASS');
  process.exit(0);
}
main().catch((e) => { console.error(e.message); process.exit(1); });
```

**Current output (v0.1.10):**
```
Captured 0 JSON responses
FAIL: captured 0 JSON responses
```

## Issue 2: Booking.com crash — Cannot redefine property: src

**Symptom:** Obscura process crashes when navigating to any Booking.com hotel page. Process exits with code null.

**Logs:**
```
[obscura:stderr] ERROR obscura::console: Timer error: TypeError: Cannot redefine property: src
    at Object.defineProperty (<anonymous>)
    at new Image (<obscura:bootstrap>:7219:12)
    at <script>:3:11
    at <obscura:bootstrap>:368:11
[obscura] process exited with code null
```

### Automated regression test

```js
// test-booking-crash.mjs
import { chromium } from 'playwright-core';

async function main() {
  const browser = await chromium.connectOverCDP('http://127.0.0.1:9222');
  const context = await browser.newContext();
  const page = await context.newPage();

  let crashed = false;
  page.on('crash', () => { crashed = true; });

  try {
    await page.goto('https://www.booking.com/hotel/ru/four-seasons-resort-bali-at-sayan.html', {
      waitUntil: 'domcontentloaded', timeout: 30_000
    });
  } catch (e) {
    console.error('FAIL: browser disconnected:', e.message.split('\n')[0]);
    process.exit(1);
  }

  if (crashed) { console.error('FAIL: page crashed'); process.exit(1); }

  const title = await page.title();
  await context.close();
  console.log('Title:', title);
  if (!title || title === 'Just a moment...') { process.exit(1); }
  console.log('PASS');
  process.exit(0);
}
main().catch((e) => { console.error('FAIL:', e.message.split('\n')[0]); process.exit(1); });
```

**Current output (v0.1.10):**
```
FAIL: browser disconnected: Protocol error (Target.getTargetInfo): Target closed.
```

## Root cause (Issue 2)

Booking.com's anti-bot JS probes `Image.prototype.src`. The Image shim tries to `Object.defineProperty` on `src`, but the real `HTMLImageElement` already defines it as non-configurable.

**Suggested fix:**
```js
const desc = Object.getOwnPropertyDescriptor(Image.prototype, 'src');
if (!desc || desc.configurable) {
  Object.defineProperty(Image.prototype, 'src', { ... });
}
```

## Full regression test suite

A complete automated test suite is available at:
https://github.com/richtravel/r1chtravel/blob/main/scripts/obscura-regression-test.mjs

Run:
```bash
obscura serve --port 9222
npm i playwright-core
node scripts/obscura-regression-test.mjs
```

## Impact
Both bugs block Obscura from being used in production for flight/hotel scraping. CloakBrowser handles both sites fine.

Jerc92:
## What changed

- preserve binary JavaScript fetch response bodies for CDP Network events
- apply Chromium-compatible charset decoding to main-document response bodies
- return opaque or non-decodable bodies through `Network.getResponseBody` with `base64Encoded: true`
- retain request-ID correlation through response and body lookup
- leave XHR response decoding to the separate #755 change

Chromium returns recognized non-UTF-8 text, including GBK, as decoded Unicode with `base64Encoded: false`. This branch follows that boundary instead of treating every non-UTF-8 byte sequence as binary.

This is required by the Obscura 0.2.1 Google Flights, CDP, and Booking report and is a follow-up to #394. It does not close #394 because the live Aviasales selection response and readable-body gate has not passed.

## Validation

- browser release nextest: 81/81 passed
- affected CDP network/body release nextest: 5/5 passed
- full workspace release nextest: 1552 passed, 3 failed, 4 skipped
  - the robots timeout reproduced unchanged at the original PR head
  - the MCP snapshot test passed on rerun and at the original PR head
  - the intersection observer test passed in isolation and is covered by prerequisite #786
- all four documented release builds passed
- obstacle course at this head: 32/33, only `observer-intersection` failed
- obstacle course with prerequisite #786 applied in a disposable integration worktree: 33/33
- independent correctness and KISS/duality reviews: passed

## Performance

Interleaved release runs used identical local documents and discarded warmups. The repository noise band is +/-10%.

- UTF-8 fast path: CPU unchanged; median latency 92.661 to 83.123 ms; peak RSS 29.594 to 29.930 MiB (+1.1%)
- opaque invalid-UTF-8 path: CPU unchanged; median latency 77.968 to 80.168 ms (+2.8%); peak RSS 27.875 to 27.547 MiB (-1.2%)
- GBK document: the candidate completed with 27.578 MiB median peak RSS; the original PR head timed out on the same deterministic fixture, so no like-for-like latency ratio is reported

The response-store entry and raw-byte caps are checked before decoding or base64 allocation. Store capacity and ownership are unchanged.

## Rendering

Not applicable. This does not change layout, paint, screenshots, screencast, or PDF output.

## Checklist

- [x] Tests cover the failure and Chromium boundary behavior.
- [x] Render, stealth, and no-default build configurations pass.
- [x] CPU, latency, and memory were compared against the original PR head.
- [x] No public Rust API changed.

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -3114,19 +3114,21 @@ impl Page {
             PageError::NetworkError(e.to_string())
         })?;
 
-        // Store binary main resources (images, PDFs, octet-stream) base64 so
-        // Network.getResponseBody returns intact bytes. A UTF-8-lossy text store
-        // corrupts them (issue #340). Text-like types stay as text.
-        let main_is_binary = !is_text_like_content_type(response.content_type());
-        self.record_network_event_with_body(
+        // Apply Chromium's charset-aware DevTools body policy. This preserves
+        // both non-UTF-8 text and opaque binary bytes without lossy conversion.
+        let request_id = self.record_network_event_inner(
             url.as_str(),
             "GET",
             "Document",
             response.status,
             &response.headers,
-            &response.body,
-            main_is_binary,
+            response.body.len(),
         );
+        let body = &response.body;
+        let content_type = response.content_type();
+        self.store_response_body_with(request_id, body.len(), || {
+            stored_devtools_document_response_body(body, content_type)
+        });
 
         if !response.redirected_from.is_empty() {
             self.url = Some(response.url.clone());
@@ -4084,23 +4086,32 @@ impl Page {
     }
 
     fn store_response_body(&mut self, request_id: String, body: &[u8], base64_encoded: bool) {
+        self.store_response_body_with(request_id, body.len(), || StoredResponseBody {
+            body: if base64_encoded {
+                BASE64.encode(body)
+            } else {
+                String::from_utf8_lossy(body).to_string()
+            },
+            base64_encoded,
+        });
+    }
+
+    fn store_response_body_with<F>(
+        &mut self,
+        request_id: String,
+        source_len: usize,
+        make_stored: F,
+    )
+    where
+        F: FnOnce() -> StoredResponseBody,
+    {
         let max_entries = response_body_entry_limit();
         let max_bytes = response_body_byte_limit();
-        if max_entries == 0 || max_bytes == 0 || body.len() > max_bytes {
+        if max_entries == 0 || max_bytes == 0 || source_len > max_bytes {
             return;
         }
-        let body = if base64_encoded {
-            BASE64.encode(body)
-        } else {
-            String::from_utf8_lossy(body).to_string()
-        };
-        self.response_bodies.insert(
-            request_id.clone(),
-            StoredResponseBody {
-                body,
-                base64_encoded,
-            },
-        );
+        let stored = make_stored();
+        self.response_bodies.insert(request_id.clone(), stored);
         self.response_body_order.push_back(request_id);
         while self.response_body_order.len() > max_entries {
             if let Some(oldest) = self.response_body_order.pop_front() {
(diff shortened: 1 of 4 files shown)


yinnho:
Hit the same hole from a different entry point — worth a test seed if not already covered. A GBK page ships `content-type: text/html; charset=gbk`: that passes any text-like content-type check, so the body got stored as a string, and `Network.getResponseBody` handed back U+FFFD for every GBK character. The gate that caught it for us is "store base64 unless the bytes are valid UTF-8", checked in both places that store bodies independently (the document store and the script-initiated fetch-{N} store). With that, a GBK document and a page-side fetch() of a GBK endpoint round-trip byte-exact through getResponseBody in our tests — added two for exactly this, full suite green (yinnho/aginxbrowser@99a28f9).

This matches what we'd already shipped on our side for binary bodies (transport + XHR round-trip), so no counter-data from me — just the non-UTF-8-text case as an extra test input beyond binary content-types.

Jerc92:
Thanks, this exposed the missing document-store case. I added GBK coverage for both the main document and page-side fetch in 9ae2ac8. One Chromium parity detail differs from the proposed gate: raw CDP against Chromium 145 returns declared GBK as decoded Unicode with base64Encoded=false for both paths. Obscura now follows that behavior and uses base64 only for opaque or non-decodable bodies. The response-store caps still run before decoding or base64 allocation.

Jerc92:
Covered in the current head `9ae2ac8`. The regression matrix includes `text/html; Charset='GBK'` for the document store and `text/plain; charset=gbk` for the page-fetch store. Both assert decoded `中文` with `base64Encoded: false`; invalid textual bytes still fall back to byte-exact base64. The two stores retain separate tests because they have independent ownership paths.

yinnho:
Verified your Chromium numbers with a local headless Chrome 152 + raw CDP harness before touching anything: declared-GBK comes back as decoded Unicode with base64Encoded=false for both the document and a page-side fetch of a GBK endpoint, `application/json` with a stray 0xFF comes back base64, octet-stream stays base64. So the gate we proposed ("base64 unless valid UTF-8") was wrong for every declared non-UTF-8 charset — your charset-aware policy is the Chrome-parity one. We realigned our stores to it (explicit charset, else the family default, base64 only when that decode fails without replacement): yinnho/aginxbrowser@56ee5bf.

Two things that fell out of the experiment, in case they're useful:

- Bodies with no Content-Type at all, and text/plain without a charset, still come back as *text* from Chrome 152 (0x81 0x8D → U+0081 U+008D, a windows-1252-style total decode), not base64. The current head treats absent content-type as opaque, so that looked like a small divergence — guessing it just depends on what /no-type asserts.
- The fatal-decode path is easy to get wrong: with encoding_rs, allocating `bytes.len()+1` output makes `decode_to_string_without_replacement` report OutputFull for multi-byte output (one `€` from a single windows-1252 byte), which silently degrades to base64. `max_utf8_buffer_length_without_replacement` is the safe capacity. We had that bug.

Jerc92:
Useful, thank you. The fatal decoder already uses max_utf8_buffer_length_without_replacement, but the missing Content-Type case was still treated as opaque. Commit 04df2c8 now decodes absent Content-Type as Windows-1252 text in both the document and fetch response-body stores. Added 0x81/0x8D document/unit and raw CDP coverage; focused release nextest is 3/3. Explicit opaque MIME handling is unchanged.

yinnho:
Pulled 04df2c8 and checked it against our build — aligned. We'd already been treating absent Content-Type as the plain-text family with the windows-1252 default (same `0x81 0x8D -> U+0081 U+008D` case, base64 only when the declared-label decode isn't replacement-free), and our document and fetch stores share one policy function too. So the two implementations now agree with each other and with Chrome 152 across the matrix. Thanks for closing this out.

Issue #810, opened 2026-09-02
fix(net): is_forbidden_ip misses CGNAT (100.64/10) and other special ranges — SSRF to cloud metadata


mnaza:
## Summary

`is_forbidden_ip` — the shared SSRF deny-set used by BOTH the literal-host check (`validate_url`) and the DNS-resolution guard (`SsrfGuardResolver`), on BOTH the reqwest and wreq/stealth transports — relies on std's `Ipv4Addr::is_private()`, which only covers RFC 1918 (10/8, 172.16/12, 192.168/16). It does not block several IANA special-purpose ranges that host internal/metadata services.

## Location

`crates/obscura-net/src/client.rs:573` (`is_forbidden_ip`).

## Missing ranges

- **100.64.0.0/10 — CGNAT / RFC 6598.** Cloud metadata endpoints live here: **Alibaba Cloud `100.100.100.200`**, Oracle Cloud internal. (AWS/GCP `169.254.169.254` is already caught via `is_link_local`.)
- **198.18.0.0/15 — benchmarking / RFC 2544** (often network-management ranges).
- **192.88.99.0/24 — 6to4 relay anycast / RFC 7526.**

## Exploit

A page (or a DNS name it controls) resolving to `100.100.100.200` passes both the string check and the DNS-rebinding guard on both transports, and `fetch()`/navigation reaches the cloud metadata service → credential/token theft (SSRF).

## Fix direction

Add explicit prefix checks to `is_forbidden_ip`'s IPv4 arm for 100.64.0.0/10, 198.18.0.0/15, and 192.88.99.0/24. The IPv6 arm already recurses into the embedded IPv4 for mapped/compatible forms, so those are covered automatically.

## Severity

HIGH — SSRF to cloud metadata (CGNAT range) on both transports; the guard is otherwise correctly wired.

Issue #788, opened 2026-09-01
new FormData(form) doesn't read the form's field values (silently empty); .keys()/.values() also not iterable


hprabowo076:
### Environment
- obscura 0.2.1 (Linux x86_64 release build)
- CDP reports Chrome/145.0.0.0, V8 14.5.0.0

### Bug 1: `new FormData(form)` auto-populate is a no-op

Per spec, `new FormData(form)` should read the form's current field values.
Instead it constructs an empty FormData regardless of the form's contents.

Repro (self-contained, no network needed):

    obscura fetch 'data:text/html,<form id=f><input name=a value=x><input name=b value=y></form>' \
      --quiet --eval '(function(){
        var fd = new FormData(document.getElementById("f"));
        return JSON.stringify({ a: fd.get("a"), b: fd.get("b") });
      })()'

Expected: `{"a":"x","b":"y"}`
Actual:   `{"a":null,"b":null}`

Manually-built FormData works correctly:

    obscura fetch 'data:text/html,<div></div>' --quiet --eval '(function(){
      var fd = new FormData(); fd.append("a","x"); return fd.get("a");
    })()'
    → "x"   (correct)

### Bug 2: `.keys()` / `.values()` are not iterable (`.entries()`/`forEach`/`.has()`/`.getAll()` are fine)

    obscura fetch 'data:text/html,<div></div>' --quiet --eval '(function(){
      var fd = new FormData(); fd.append("a","x");
      var out = {};
      try { out.keys = [...fd.keys()]; } catch(e) { out.keysErr = String(e); }
      try { out.values = [...fd.values()]; } catch(e) { out.valuesErr = String(e); }
      out.entries = [...fd.entries()];   // works
      out.has = fd.has("a");             // works
      out.getAll = fd.getAll("a");       // works
      return JSON.stringify(out);
    })()'

Actual: `keysErr`/`valuesErr` = `TypeError: fd.keys is not a function or its return value is not iterable`; `entries`/`has`/`getAll` all correct.

### Impact

Bug 1 breaks the extremely common `fetch(url, { method: 'POST', body: new FormData(form) })`
pattern that most real-world HTML forms use for AJAX submission. A page using this pattern
submits a body with **zero fields** with no JS-visible exception — from inside the page it looks
like a successful submit, but the server receives an empty payload. This makes CDP-driven
form automation against such pages fail non-obviously (discovered while scripting a real
production order form: the submit fired, returned a normal-looking response, but the server
[...]

PR #816, opened 2026-09-02
fix(net): complete the SSRF deny-set incl. IPv6 embedded-IPv4 (6to4/NAT64)

963s:
A full walk of the IANA IPv4+IPv6 special-purpose registry against `is_forbidden_ip` (classifying each address on the official binary as blocked-by-guard vs dialed) turned up gaps on both arms. The worst is on the IPv6 side and was missed by #810/#812.

## IPv6 embedded-IPv4 bypass (both transports) — the important one

The v6 arm only unwrapped `::ffff:a.b.c.d` and `::a.b.c.d`. But **6to4 (`2002::/16`) and NAT64 (`64:ff9b::/96`) also carry a v4 address**, and their embedded form bypassed the guard entirely — a plain literal host, no DNS, no proxy:

| address | embeds | before |
|---|---|---|
| `[2002:a9fe:a9fe::]` | 169.254.169.254 (AWS metadata) | **reached** |
| `[2002:7f00:1::]` | 127.0.0.1 (loopback) | **reached** |
| `[2002:a00:1::]` | 10.0.0.1 (RFC1918) | **reached** |
| `[2002:6464:64c8::]` | 100.100.100.200 (Alibaba) | **reached** |
| `[64:ff9b::7f00:1]` | 127.0.0.1 (NAT64) | **reached** |

This defeats the AWS/GCP `169.254.169.254` block (`is_link_local`) that was assumed airtight, and loopback too. Fixed by extracting the embedded v4 from **every** format in one place (`embedded_ipv4`) and re-checking it through `is_forbidden_ip`.

## IPv4 ranges completed

`0.0.0.0/8`, `100.64.0.0/10` (CGNAT — Alibaba `100.100.100.200`), `192.0.0.0/24` (Oracle `192.0.0.192`), `192.31.196.0/24`, `192.52.193.0/24`, `192.88.99.0/24` (6to4 relay), `192.175.48.0/24`, `198.18.0.0/15`, `224.0.0.0/4` multicast, `240.0.0.0/4`. Native IPv6: `2001:db8::/32`, `100::/64`, `ff00::/8`.

## Relationship to #812

#812 adds CGNAT + 198.18/15 + 192.88.99/24. This PR is the **union** of that plus the embedded-v4 class and the rest of the registry, so it supersedes #812. Happy to instead rebase down to just the remainder if #812 lands first — whichever the maintainers prefer.

## Verification

- new tests `iana_special_purpose_ipv4_ranges_are_forbidden`, `ipv6_embedded_ipv4_forms_are_forbidden`, `native_ipv6_special_ranges_are_forbidden`; RED before (`64:ff9b::7f00:1 should be forbidden`), GREEN after
- adjacent-address guards so public hosts stay allowed, including a **6to4/NAT64-wrapped `8.8.8.8`** (no over-block)
- `obscura-net` suite: 89 pass, the one failure (`configured_roots_trust_a_private_ca_via_ssl_cert_file`) is a pre-existing local-TLS-server test that does not come up in this sandbox and is unrelated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/obscura-net/src/client.rs
+++ b/crates/obscura-net/src/client.rs
@@ -573,41 +573,73 @@ pub fn env_allows_private_network() -> bool {
 pub fn is_forbidden_ip(ip: IpAddr) -> bool {
     match ip {
         IpAddr::V4(v4) => {
-            let o = v4.octets();
-            v4.is_loopback()
-                || v4.is_private()
-                || v4.is_link_local()
-                || v4.is_broadcast()
-                || v4.is_documentation()
-                || v4.is_unspecified()
-                // std's is_private() covers only RFC1918, so add the IANA
-                // special-purpose ranges that also host internal services and
-                // are common SSRF targets:
-                //   100.64.0.0/10  CGNAT / RFC6598 — cloud metadata (e.g.
-                //                  Alibaba 100.100.100.200) lives here.
-                //   198.18.0.0/15  benchmarking / RFC2544.
-                //   192.88.99.0/24 6to4 relay anycast / RFC7526.
-                || (o[0] == 100 && (64..=127).contains(&o[1]))
-                || (o[0] == 198 && (o[1] == 18 || o[1] == 19))
-                || (o[0] == 192 && o[1] == 88 && o[2] == 99)
-        }
-        IpAddr::V6(v6) => {
-            if v6.is_loopback()
-                || v6.is_unspecified()
-                || v6.is_unique_local()
-                || v6.is_unicast_link_local()
+            // Ranges std has a stable predicate for.
+            if v4.is_loopback()          // 127.0.0.0/8
+                || v4.is_private()       // 10/8, 172.16/12, 192.168/16
+                || v4.is_link_local()    // 169.254/16 (incl. 169.254.169.254 metadata)
+                || v4.is_broadcast()     // 255.255.255.255
+                || v4.is_documentation() // 192.0.2/24, 198.51.100/24, 203.0.113/24
+                || v4.is_unspecified()   // 0.0.0.0 exactly
             {
                 return true;
             }
-            // Unwrap IPv4-mapped (::ffff:a.b.c.d) and IPv4-compatible (::a.b.c.d)
-            // forms and re-check the embedded v4 so e.g. [::ffff:127.0.0.1] or
-            // [::ffff:169.254.169.254] cannot slip past the v6 arm.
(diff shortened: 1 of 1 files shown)


yinnho:
Hit the exact same hole in our obscura-js-derived engine this week and shipped the same shape of fix — same `embedded_ipv4` helper, same test vectors (`[2002:a9fe:a9fe::]` reaching the AWS metadata endpoint, `[64:ff9b::6464:64c8]` reaching Alibaba's).

Your patch also caught a gap in our first cut: I'd constrained the NAT64 arm to the well-known `64:ff9b::/96` only, and reading your prefix-pair check made me go add the RFC 8215 local-use `64:ff9b:1::/48` form, which carries its v4 in the same low 32 bits.

One thing worth pinning in the tests: wrappers around a *public* v4 must stay reachable (`2002:0808:0808::` is 8.8.8.8 and should not be denied). Your construction gets that right by re-checking the embedded address, but it's the kind of behavior that regresses into a blanket prefix ban later, so an explicit "public wrapper stays allowed" assertion earns its keep.

963s:
thanks, good to see someone hit this independently. makes it a lot easier to argue it's real and not just a corner case.

on the nat64 arm: the pair check is intentional. it only keys on the `64:ff9b` pair and pulls the v4 from the low 32 bits, so the rfc 8215 local-use `64:ff9b:1::/48` form already falls out of the same code with no second arm, as long as the v4 sits in the low 32 bits (which is your case). the one thing it won't catch is a strict rfc 6052 /48 that puts the v4 in the middle bits, and i left that out on purpose.

agree on the public wrapper test though. `2002:0808:0808::` is 8.8.8.8 and has to stay reachable. the re-check handles it fine right now, but that's exactly the kind of thing someone breaks later by collapsing it into a flat prefix ban, so i'll add an explicit test that a public v4 wrapper stays allowed (6to4 + nat64 + mapped).

we've been circling the same surface for a while anyway, your #788 and #751 were good catches.

963s:
closing this in favor of #825, which landed the embedded-address handling (NAT64 64:ff9b::/96, 6to4 2002::/16, and the mapped/compatible re-check) plus the local-use 64:ff9b:1::/48 block, and #812 for the CGNAT / IANA ranges. between them main already covers what this PR was after.

the public-wrapper regression @yinnho and i talked about is in there too (2002:808:808:: and 64:ff9b::808:808 assert as allowed, and validate_url has the [64:ff9b::808:808] case), so there's nothing separate to add.

thanks for the reviews.

PR #835, opened 2026-09-04
fix(cdp): implement Input.insertText and label association getters

### LINKED ISSUE #577: CDP: Input.insertText not implemented (Unknown Input method: insertText)
**Describe the bug**
The `Input` domain does not implement `Input.insertText` — the CDP dispatcher responds `Unknown Input method: insertText`. `Input.dispatchKeyEvent` works (key presses fine), so only the text-insertion method is missing.

**To reproduce** (raw CDP over WS):
1. `obscura serve --port 9222`
2. Create/attach a page, navigate to a page with a text input (e.g. https://en.wikipedia.org)
3. Focus the input, then send:
   ```json
   {"method": "Input.insertText", "params": {"text": "hello"}}
   ```

Response:
```json
{"error": {"code": -32601, "message": "Unknown Input method: insertText"}}
```

**Expected**: text is inserted into the focused field (Chrome implements `Input.insertText`; it's the standard path for clipboard-style/paste input in CDP clients).

**Impact**: CDP clients that type via `Input.insertText` (e.g. Hermes Agent's browser_type tool) cannot enter text at all. Implementing it (or falling back to `Input.dispatchKeyEvent` per character server-side) would restore typing.

Env: obscura 0.2.0 (obscura-x86_64-linux.tar.gz), x86_64 Linux.

SGavrl:
Input.insertText had no handler, so Playwright's fill() failed with 'Unknown Input method: insertText'. The method now drives the existing insert_text_js snippet. Playwright's getByLabel also needs the DOM label association: Element.labels and HTMLLabelElement.control now follow the HTML labelable-elements rules, for both for-linked and wrapping labels.

Fixes #577.

## What changed

Two gaps blocked Playwright form filling over CDP, reported by a Momentic compatibility POC against v0.2.1:

• `Input.insertText` had no dispatch arm, so `locator.fill()` (and any raw insertText client) got a protocol error. It now drives the existing `insert_text_js` snippet on the session page, the same path `dispatchKeyEvent` uses for typed text, so quotes, backslashes, and newlines survive.
• `getByLabel` never resolved because the label association was missing from the DOM layer: `Element.labels` returned nothing useful and `HTMLLabelElement` was a bare `Element` alias with no `control`. Both getters now follow the HTML labelable-elements list (button, input except hidden, meter, output, progress, select, textarea), covering `for`-linked and wrapping labels.

## Validation

• Momentic-style probe: Playwright Core `connectOverCDP` against `obscura serve`, local fixture with a labeled input. `getByLabel("Name").fill("hello")` now passes and fires the page's `input` listener; `page.screenshot()` returns a valid PNG; navigator identity is populated. Same result in render and render+stealth builds.
• New regression tests:
• `obscura-cdp::input_key_event_escaping::insert_text_types_into_the_focused_field` (insertText with quotes, backslash, newline)
• `obscura-js runtime::tests::label_control_and_labels_link_labelable_elements` (for-linked, wrapping, hidden-input, and non-labelable cases)
• Full suite: `cargo nextest run --release --features render --no-fail-fast` -> 1581 passed, 0 failed, 4 skipped.
• Obstacle course: `OBSCURA_BIN=./target/release/obscura python3 obstacle-course/run.py --runs 1 --warmup 0` -> 33/33.

## Rendering

Not applicable.

## Performance

The new CDP arm runs once per insertText call. The `labels` getter scans `document.querySelectorAll('label')` only when the element is labelable and has an id; the scan is JS-side filtering, no selector parsing of the id. No hot path, memory, or rendering changes.

## Checklist

• [x] The change is focused and does not remove existing behavior without justification.
• [x] Tests cover the failure or feature.
• [x] Existing tests pass, including render and no-render configurations when affected.
• [x] I checked for CPU, latency, and memory regressions.
• [x] Public API or user-facing behavior changes are documented.

### DIFF
--- a/crates/obscura-cdp/src/domains/input.rs
+++ b/crates/obscura-cdp/src/domains/input.rs
@@ -316,6 +316,15 @@ pub async fn handle(
 
             Ok(json!({}))
         }
+        // Chrome's Input.insertText: Playwright's fill() focuses the field in
+        // page and then types the whole value through this one call (#577).
+        "insertText" => {
+            let text = params.get("text").and_then(|v| v.as_str()).unwrap_or("");
+            if let Some(page) = ctx.get_session_page_mut(session_id) {
+                page.evaluate(&insert_text_js(text));
+            }
+            Ok(json!({}))
+        }
         "dispatchKeyEvent" => {
             let event_type = params.get("type").and_then(|v| v.as_str()).unwrap_or("");
             let key = params.get("key").and_then(|v| v.as_str()).unwrap_or("");
--- a/crates/obscura-cdp/tests/input_key_event_escaping.rs
+++ b/crates/obscura-cdp/tests/input_key_event_escaping.rs
@@ -164,3 +164,52 @@ async fn dispatch_key_event_char_carries_a_newline_into_a_textarea() {
         "a newline sent as a char must reach the field, not be dropped by a malformed snippet"
     );
 }
+
+// #577: Playwright's fill() focuses the field in page and then types the whole
+// value with one Input.insertText call. The method must exist and drive the
+// same snippet the key-event text path uses, so quotes, backslashes, and
+// newlines survive intact.
+#[tokio::test(flavor = "current_thread")]
+async fn insert_text_types_into_the_focused_field() {
+    std::env::set_var("OBSCURA_ALLOW_PRIVATE_NETWORK", "1");
+    let url = serve_page().await;
+    let mut ctx = CdpContext::new();
+    let page_id = ctx.create_page();
+    let session_id = "session-1";
+    ctx.sessions.insert(session_id.to_string(), page_id.clone());
+
+    cdp(&mut ctx, 1, "Page.navigate", json!({"url": url, "waitUntil": "load"}), session_id).await;
+    cdp(
+        &mut ctx,
+        2,
+        "Runtime.evaluate",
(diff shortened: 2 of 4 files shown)


yinnho:
We ported these two getters into our obscura-js-derived engine this week too — label→control click forwarding was already there, but `el.labels` was undefined and `label.control` missing, so getByLabel-style queries found nothing.

While pinning tests against the spec I collected the association edge cases where a natural implementation drifts from Chrome; this one hits four of them, in case a follow-up ever wants them:

- `for=""` (present but empty): `control` should be null, no fallback to descendants — the `if (forId)` truthiness check falls through to the descendant scan.
- an ancestor label whose own `for` points elsewhere (`<label for=other><input id=a>`): `a.labels` is empty in Chrome (that label's control is the for-target, not `a`), but the ancestor walk adds it unconditionally.
- non-first labelable descendant (`<label><input id=a><input id=b>`): `b.labels` is empty (the label's control is `a`); the walk adds it for `b` too.
- duplicate ids (`label[for=x]` with an earlier element also `id=x`): only the element `getElementById` resolves to owns the label; the for-scan string-matches and lists it for both.

What ended up simplest for us was routing both directions through a single labeled-control resolver (for→getElementById→labelable check, else first labelable descendant) and building `labels` by filtering the document's labels through it — the edge cases then fall out instead of needing their own arms.

PR #840, opened 2026-09-04
fix(cdp): repair the build after the 747/790/756 interaction, accept …

SGavrl:
Merging #747, #790 and the #756 isolate-entry refactor in any order left main not compiling: a removed cleaned_expr binding still fed evaluation_recipes, and two module-eval paths still read the runtime as a field. Reconcile both call sites.

Browser.grantPermissions/resetPermissions previously returned an unknown-method error, which aborts Playwright's browser-context setup. Accept them as no-ops; no engine API is gated on permission grants today.

PerformanceObserver.supportedEntryTypes was undefined, so clients that feature-detect with supportedEntryTypes.includes() threw. Report the entry types the engine actually emits records for.

### DIFF
--- a/crates/obscura-cdp/src/domains/browser.rs
+++ b/crates/obscura-cdp/src/domains/browser.rs
@@ -31,6 +31,12 @@ pub async fn handle(method: &str, _params: &Value) -> Result<Value, String> {
         // lets the client's setup sequence complete instead of tearing down
         // the page on an unknown-method error.
         "setWindowBounds" => Ok(json!({})),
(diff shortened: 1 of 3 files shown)


yinnho:
Picked up both of these downstream today, thanks. One note on supportedEntryTypes: we expose the honest subset — mark, measure, navigation, paint — exactly what our Performance timeline actually records, instead of Chrome's full list. Web-vitals wrappers gate their LCP/CLS/longtask observers on this list, and advertising types that never fire sends them into a wait that never resolves. The regression test walks it from the other side: every entryType present in getEntries() must be in supportedEntryTypes.

Issue #680, opened 2026-08-18
Runtime.evaluate returns 'No page' on freshly created tab via CDP


arrrrny:
## Description

`Runtime.evaluate` returns `{"error":{"code":-32601,"message":"No page"}}` on a freshly created tab, even after `Page.navigate` and `Runtime.enable` both succeed. This prevents any JS execution (DOM reads, injections, event dispatch) via the CDP websocket connection.

## Reproduction

1. Start obscura: `./obscura serve -p 9333`
2. Open a page via CDP:
   ```python
   tabs = json.load(urllib.request.urlopen("http://127.0.0.1:9333/json", timeout=5))
   page = next(t for t in tabs if t.get("type")=="page")
   ws = await websockets.connect(page["webSocketDebuggerUrl"])
   ```
3. Enable domains + navigate:
   ```python
   await ws.send(json.dumps({"id":1,"method":"Runtime.enable","params":{}}))
   await ws.send(json.dumps({"id":2,"method":"Page.navigate","params":{"url":"https://example.com"}}))
   await asyncio.sleep(3)
   ```
4. Evaluate — fails:
   ```python
   await ws.send(json.dumps({"id":3,"method":"Runtime.evaluate",
       "params":{"expression":"1+1","returnByValue":True}}))
   # Returns: {'id': 3, 'error': {'code': -32601, 'message': 'No page'}}
   ```

## Expected behavior

`Runtime.evaluate` returns the result of the expression (e.g. `{"result":{"type":"number","value":2}}`) — same as Chrome DevTools Protocol.

## Actual behavior

`Runtime.evaluate` returns `{"error":{"code":-32601,"message":"No page"}}` regardless of the expression.

## Environment

- obscura v0.2.0 (`obscura-aarch64-macos-stealth.tar.gz`)
- macOS 15.7 (Intel x86_64)
- Also reproduced on: `obscura-x86_64-macos-stealth` (same error)

## Notes

- `Page.navigate` dispatches successfully (response id comes back with no error)
- `Page.enable`, `Runtime.enable` both return `{}` (success)
- The websocket connection itself is stable (events like `Page.loadEventFired` fire)
- This appears to be a page-attachment issue: the page object exists (visible in `/json` endpoint, has a URL), but `Runtime.evaluate` can't find it

## CDP endpoints observed

- `/json` → returns the tab list with valid `webSocketDebuggerUrl`
[...]

PR #848, opened 2026-09-05
fix(js): hit-test by paint order, not document order

### LINKED ISSUE #738: elementFromPoint ignores z-index among positioned siblings
`document.elementFromPoint()` appears to resolve overlapping positioned siblings by DOM order rather than paint order.

Minimal repro:

```html
<style>
.dialog { position: fixed; top: 100px; left: 100px; width: 400px; height: 200px; z-index: 1000; }
.close-button { position: absolute; top: 10px; right: 10px; width: 32px; height: 32px; z-index: 1002; }
.loading-overlay { position: absolute; inset: 0; z-index: 1001; }
</style>
<div class="dialog">
  <button class="close-button">×</button>
  <div class="loading-overlay"></div>
</div>
<script>
const button = document.querySelector('.close-button');
const rect = button.getBoundingClientRect();
const hit = document.elementFromPoint(rect.left + rect.width / 2, rect.top + rect.height / 2);
console.log(hit.className);
</script>
```

The button is earlier in the dialog DOM and has `z-index: 1002`; the overlay follows it and has `z-index: 1001`. In Chrome the result is `close-button`, but in Obscura it is `loading-overlay`.

I found this while using CDP `Input.dispatchMouseEvent`: a coordinate click aimed at the visibly topmost close button was delivered to the lower-z overlay. Hit-testing should select the topmost painted element within the stacking context.

NishilRathod:
Closes #738.

`elementFromPoint` ranked hit candidates by node id — document order — so
`z-index` never entered the decision. In the reported repro a close button with
`z-index: 1002` loses to a `z-index: 1001` overlay that follows it in the DOM.
That matters beyond the DOM API: `Input.dispatchMouseEvent` resolves its target
through `document.elementFromPoint` (`crates/obscura-cdp/src/domains/input.rs`
:131, :154, :280), so CDP coordinate clicks landed on the visually-buried
element.

## Approach

The painter already sorts boxes into the CSS 2.1 Appendix E bands to build its
display list, so the order it reaches them in *is* paint order. Recomputing
z-order separately for hit testing would drift from the painter the first time
either side changed, so this exports what the painter already knows:

- `stacking_band_order()` — that ordering, lifted **verbatim** out of
  `paint_laid_dom_scrolled` so both callers share one source of truth. The paint
  path is unchanged; it now calls the extraction.
- `paint_sequence_into()` — mirrors the painter's own recursion into stacking
  units and floats to produce a flat front-to-back sequence. It takes
  `&DomLayout` and never a `Pixmap`, so a hit test triggers no raster work.
- `op_paint_slots` — serves slots from the same memoized `PreparedRender` that
  already answers `op_layout_geometry`, so there is no extra layout either. It
  answers only the nodes asked about, as one comma-separated string.
- `bootstrap.js` — ranks candidates by paint slot, keeping document order as the
  tie-break and as the fallback for the no-render build, where there is no
  layout to order by.

`elementsFromPoint` fell out of the same ranking. It used to return the single
`elementFromPoint` result in a one-element array; it now reports the real stack.

## Result

Before: `{"hit":"loading-overlay","stack":["loading-overlay"]}`
After:  `{"hit":"close-button","stack":["close-button","loading-overlay","dialog"]}`

`dialog` is in the stack because it is the containing positioned box, also under
the point. `<html>`/`<body>` remain excluded, which is pre-existing behaviour of
this implementation and a deliberate difference from Chrome — happy to change it
if you would rather match.

## Testing

- `cargo nextest run --release --features render --no-fail-fast` — 1634 passed.
- No-render build compiles, and `cargo nextest run --release
  --no-default-features -p obscura-js` — 386 passed. This is the path that
  exercises the document-order fallback, since `op_paint_slots` is
  feature-gated.
- Two regression tests. Both style from a `<style>` block rather than inline, on
  purpose: computed style only surfaces inline values for `z-index`/`position`,
  so a comparator reading z through `getComputedStyle` would pass an
  inline-styled test and still rank everything `z:0` on a real page.

Two things I could not clear, both pre-existing — please tell me if you want
either handled here:

1. **Obstacle course is 32/33, not 33/33.** The failure is
   `observer-intersection` (`expected 'io:50', got ''`), which is #671. I ran
   the course on clean `main` at `a1e09de` and got the identical 32/33 with the
   same stage failing, so this change does not move it.
2. **I could not drive the fix end-to-end over CDP.** On a fresh connection
   `Target.getTargets` returns an empty list and `Page.navigate` answers
   `No page for session`, which is #543 / #570 / #680. The JS-level behaviour is
   verified above, and the click path's dependency on `elementFromPoint` is
   direct, but I have not observed a real `Input.dispatchMouseEvent` land on the
   button.

No rendered pixels change, so I have not added `render-repros/` fixtures. Say
the word if you want them anyway.

## Credit

@yinnho diagnosed this on the issue — the root cause, the `getComputedStyle`
trap, and the observation that the layout walk already solves stacking. The
approach follows their fix in yinnho/aginxbrowser@7feec24 (Apache-2.0).

### DIFF
--- a/crates/obscura-js/js/bootstrap.js
+++ b/crates/obscura-js/js/bootstrap.js
@@ -15056,6 +15056,54 @@ if (typeof Element !== 'undefined' && !Element.prototype.toggleAttribute) {
   };
 }
 
+// Rank hit-test candidates the way the page paints them.
+//
+// The renderer already sorts boxes into the CSS 2.1 Appendix E stacking bands
+// to build its display list, so the order it reaches them in IS paint order.
+// `op_paint_slots` reports that position rather than recomputing z-order here:
+// a comparator written in JS would have to read `z-index` through
+// `getComputedStyle`, which only surfaces inline styles, so anything styled
+// from a <style> block would silently rank as z:0.
+//
+// Returns the candidates front-to-back (topmost first). Without the op — the
+// no-render build registers no layout — this falls back to document order,
+// which is what hit testing did before paint order was available.
+function __rankHitCandidates(candidates) {
+  if (candidates.length < 2) return candidates.slice();
+
+  var slots = null;
+  try {
+    var op = Deno.core.ops.op_paint_slots;
+    if (op) {
+      var reply = op(candidates.map(function(el) { return el._nid | 0; }).join(','));
+      if (reply) {
+        slots = reply.split(',').map(function(v) { return parseInt(v, 10); });
+        if (slots.length !== candidates.length) slots = null;
+      }
+    }
+  } catch (_e) {
+    slots = null;
+  }
+
+  var ranked = candidates.map(function(el, i) {
+    var slot = slots ? slots[i] : -1;
+    return {
+      el: el,
+      // A boxless element (-1) cannot be ordered by paint; keep it below
+      // anything that paints, and settle it by document order as before.
+      slot: (typeof slot === 'number' && !isNaN(slot)) ? slot : -1,
+      nid: el._nid | 0,
+    };
+  });
+
+  ranked.sort(function(a, b) {
+    if (a.slot !== b.slot) return b.slot - a.slot;
+    return b.nid - a.nid;
+  });
+
+  return ranked.map(function(entry) { return entry.el; });
+}
+
 // Document.elementFromPoint / elementsFromPoint — no layout engine, so this is a stub:
 // in-viewport coords return <body> (or <html> as fallback), out-of-viewport returns null.
 // Wrong-but-non-throwing beats "undefined", which traps ad/analytics bootstraps in retry loops
@@ -15068,22 +15116,18 @@ if (typeof Document !== 'undefined' && !Document.prototype.elementFromPoint) {
   // containing (x,y) would never reach a deep <input> inside <label><p>.
   // Returns the deepest matching element (highest nid wins as a proxy for
   // tree depth) so descendants beat ancestors.
-  Document.prototype.elementFromPoint = function(x, y) {
-    if (typeof x !== 'number' || typeof y !== 'number' || !isFinite(x) || !isFinite(y)) {
-      return null;
-    }
-    var w = (typeof window !== 'undefined' && window.innerWidth) || 1280;
-    var h = (typeof window !== 'undefined' && window.innerHeight) || 720;
-    if (x < 0 || y < 0 || x > w || y > h) return null;
-    var all = this.querySelectorAll('*');
-    var best = null;
-    var bestNid = -1;
+  // Every element whose box contains the point, unordered. Shared by
+  // elementFromPoint and elementsFromPoint so the two cannot disagree about
+  // what was hit, only about how much of the stack they report.
+  function __hitCandidatesAt(doc, x, y) {
+    var all = doc.querySelectorAll('*');
+    var candidates = [];
     for (var i = 0; i < all.length; i++) {
       var el = all[i];
       if (!el || !el.getBoundingClientRect) continue;
       // documentElement / body span the viewport; skip them so we pick a
       // real descendant instead of falling back to <html>/<body>.
-      if (el === this.documentElement || el === this.body) continue;
+      if (el === doc.documentElement || el === doc.body) continue;
       var r = el.getBoundingClientRect();
       if (r.width === 0 || r.height === 0) continue;
       if (x >= r.left && x <= r.right && y >= r.top && y <= r.bottom) {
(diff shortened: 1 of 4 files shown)


yinnho:
This matches how we landed it. Extracting the painter's existing order instead of storing slots on layout was the part I'd consider essential — one source of truth for paint and hit-test — and answering only the nodes the hit test actually asks about through `op_paint_slots` is nicer than our version (we export the whole rank table to JS in one call). The `<style>`-block-only tests are the right call too; that `getComputedStyle` trap bit us exactly as the PR describes.

On the `<html>`/`<body>` question: Chrome does include them. For any in-viewport point, `document.elementsFromPoint` in Chrome ends `[..., dialog, body, html]` — both paint the background and always contain the point, so matching Chrome would mean appending them at the end when the point is inside the viewport. Full disclosure though: our tree excludes them too, with the same rationale written in the PR (they span the viewport and would shadow real descendants in `elementFromPoint`), so we don't match Chrome here either — I can't offer our behavior as the reference on this one.

The CDP end-to-end blocker is the #543/#570/#680 family, and your diagnosis matches what we saw when comparing architectures: on our tree a fresh connection works because the bridge registers the session at `Target.createTarget` response time (browser-level websocket, sessions keyed before the page ever attaches), so `Page.navigate` never meets an unmapped sessionId. If you want to unblock a real `Input.dispatchMouseEvent` verification ahead of those issues, that registration point is the place to look.

32/33 with `observer-intersection` failing identically on clean main is a clean baseline, nothing to add. And agreed on skipping `render-repros/` — the fix is defined as no-pixel-change, fixtures would assert nothing.

NishilRathod:
Thanks — all three suggestions are in, pushed as `6566d86`.

**`<html>`/`<body>`.** You were right, and I measured it rather than take it on
trust: headless Chrome on the #738 repro gives

```
BUTTON.close-button | DIV.loading-overlay | DIV.dialog | BODY | HTML
```

and `LEN:2 -> BODY,HTML` on a bare `<html><body></body></html>`. So
`elementsFromPoint` now appends them. They are appended, not ranked — putting
two viewport-spanning boxes into the candidate set would let them shadow every
real descendant in `elementFromPoint`, which is why they were excluded to begin
with. `elementFromPoint` still returns `ranked[0]` and is unchanged.

**One thing reviewers should see rather than find.** This changed an existing
test. `test_elements_from_point_returns_array` asserted a length of 1 on an
empty page; that encoded the old single-element behaviour, and Chrome answers 2
on the same markup. The expectation is corrected to `["BODY","HTML"]` and the
measured Chrome output is recorded in the test's doc comment. If you would
rather keep the old behaviour, say so and I will revert this commit — the paint
-order fix in `8c77121` stands on its own without it.

**Your two tests.** A `display:none` sibling with `z-index: 9999` must never
take the hit (it generates no box, so the zero-size rect check filters it before
ranking — the test pins that rather than adding behaviour). And a style mutation
between two hit tests must change the winner, which is the cheap proof that the
`PreparedRender` memo shared with `op_layout_geometry` invalidates instead of
serving a stale order. Both pass. I also extended the stack assertion to Chrome's
exact sequence rather than just the first entry, as you suggested.

**CDP.** Left alone deliberately — the `Target.createTarget` registration point
you describe is a fix for #543/#570/#680, not for this bug, and folding it in
here would make the PR about two unrelated things. Useful pointer though; it
matches what I hit.

**Unrelated flaky test, worth flagging.** While confirming the suite, I found
`obscura-browser page::tests::cancelling_frame_script_fetch_keeps_all_siblings_resumable`
fails non-deterministically. On clean `main` at `a1e09de`, 8 isolated runs gave
**3 passed / 5 failed**; on this branch 6 runs gave 5 passed / 1 failed. Same
binary each time, so it is timing-dependent and pre-existing, not from this
change. Happy to open a separate issue if it is not already known.

Full suite is otherwise green (1635/1636, the one failure being that flake), the
no-render build and its 386 tests pass — that is the path exercising the
document-order fallback — and the obstacle course is still 32/33 with
`observer-intersection` (#671) failing identically to clean `main`.

yinnho:
Re-read `6566d86` against what I asked for — everything lands, and two of the
calls are better than what I suggested.

Append-not-rank for `<body>`/`<html>` is the right resolution of the tension I
flagged: the viewport-spanning exclusion existed for a reason, and appending
keeps it while restoring the Chrome sequence. Measuring headless Chrome before
correcting the existing test is exactly the discipline this needed — that
expectation encoded our old behaviour, not the web's, so keep the correction.

The memo-invalidation test is the one I'd call load-bearing. Ranking from a
memo the hit test shares with layout geometry is the right design precisely
because that test exists — without it, a stale `PreparedRender` after a style
mutation would silently regress and nothing in the suite would notice. The
`display:none` pin I'd keep too, for the reason you wrote: a boxless element
entering the ranking would win by accident, and that failure mode is invisible
in the other tests.

On the flake — `cancelling_frame_script_fetch_keeps_all_siblings_resumable` is
not one I know, and your protocol is the right way to establish it (isolated
runs, same binary, clean `main` baseline; we got burned once bisecting a flake
off a single run). Please open the separate issue. For what it's worth, the
last timing-dependent flake we chased in this suite was two env knobs racing
one static — the fix pattern was taking a lock at knob-read time and
releasing via drop guard. If the resumable-sibling test reads shared fetch
state per test, it may be the same family.

Nothing further from me — this is good to merge from my side.

Issue #569, opened 2026-08-07
`request.continue()` overrides are silently dropped for all intercepted requests


Zed128:
### Before you start

- [x] I searched existing issues and this is not a duplicate.
- [x] I tested against the latest release or `main`.
- [x] This is not a security vulnerability (those go through SECURITY.md, not a public issue).

## Summary
When `page.setRequestInterception(true)` is enabled, calling `req.continue({ headers, url, method, postData })` from a `page.on('request', ...)` listener has **no effect** for any intercepted request — including the top-level navigation request itself. All four override fields are silently dropped, and the request goes out unmodified.

### Obscura version or commit

70062e0

### OS and architecture

Windows 11 x64

### How are you using obscura?

CDP server with Puppeteer or Playwright

### Stealth mode

Reproduces with both stealth and non-stealth

### Reproduction
```javascript
const puppeteer = require("puppeteer-core");

const CDP_WS = "ws://127.0.0.1:9222/devtools/browser";
const ECHO = "https://postman-echo.com";

async function runTest(browser, label, path, makeOverride, check) {
  const page = await browser.newPage();
  await page.setRequestInterception(true);
  page.on("request", (req) => {
    if (req.url().includes(path)) {
      req.continue(makeOverride(req));
    } else {
      req.continue();
    }
  });

  let passed = false;
  let detail = "";
  try {
    await page.goto(`${ECHO}${path}`, { waitUntil: "domcontentloaded" });
    const body = await page.content();
    passed = check(body);
    if (!passed) detail = body.slice(0, 200);
  } catch (e) {
    detail = e.message;
  } finally {
    await page.close();
  }

  console.log(`${passed ? "✅PASS" : "❌FAIL"}  [${label}]`);
  return passed;
}

async function main() {
  const browser = await puppeteer.connect({ browserWSEndpoint: CDP_WS });
  const results = [];

  // 1. headers — add x-obscura-test, /headers echoes request headers as JSON.
  results.push(await runTest(
    browser,
    "headers",
    "/headers",
    (req) => ({ headers: { ...req.headers(), "x-obscura-test": "1" } }),
    (b) => b.includes('"x-obscura-test"'),
  ));

  // 2. url — rewrite /fakeurl (would 404) to /get, response should be the
  //    /get echo with url: https://postman-echo.com/get.
  results.push(await runTest(
    browser,
    "url",
    "/fakeurl",
    () => ({ url: `${ECHO}/get` }),
[...]

Issue #754, opened 2026-08-29
XMLHttpRequest with responseType='arraybuffer' corrupts binary responses (fetch() is correct)


alexskinner:
### Before you start

- [x] I searched existing issues. **Related to #716 but a different path** — see the note below.
- [x] I tested against `main`.
- [x] This is not a security vulnerability.

### obscura version or commit

`9326880` (`main`)

### OS and architecture

macOS arm64

### How are you using obscura?

Embedded Rust library and CDP server.

### Affected area

`XMLHttpRequest` response bodies.

### Build variant

Render-enabled.

### Reproduction

Serve a file containing every byte value and read it back both ways.

```sh
python3 -c "open('bytes.bin','wb').write(bytes(range(256)))"
python3 -m http.server 9977 --bind 127.0.0.1
```

```js
const viaFetch = new Uint8Array(await (await fetch('bytes.bin')).arrayBuffer());

const viaXhr = await new Promise(res => {
  const x = new XMLHttpRequest();
  x.open('GET', 'bytes.bin', true);
  x.responseType = 'arraybuffer';
  x.onload = () => res(new Uint8Array(x.response));
  x.send();
});

console.log(viaFetch.length, viaXhr.length);   // expect 256 256
```

(Loopback needs `OBSCURA_ALLOW_PRIVATE_NETWORK=1`, or serve the fixture from anywhere else.)

### Expected behavior

Both are 256 bytes and byte-for-byte identical to the file. Chrome 147:

```
fetch len 256 | xhr len 256 | diverge at 256
FETCH 78 79 7a 7b 7c 7d 7e 7f 80 81 82 83 84 85 86 87 88 89 8a 8b …
XHR   78 79 7a 7b 7c 7d 7e 7f 80 81 82 83 84 85 86 87 88 89 8a 8b …
```

### Actual behavior

`fetch()` is correct. `XMLHttpRequest` returns 255 bytes, and every byte from `0x80` on is wrong:

```
fetch len 256 | xhr len 255 | diverge at 128
FETCH 78 79 7a 7b 7c 7d 7e 7f 80 81 82 83 84 85 86 87 88 89 8a 8b 8c 8d 8e 8f …
XHR   78 79 7a 7b 7c 7d 7e 7f 01 c2 83 c4 85 c6 87 c8 89 ca 8b cc 8d ce 8f d0 …
```

Bytes `0x00–0x7F` survive; everything above is mangled. The pattern identifies the
[...]

PR #874, opened 2026-09-06
fix(cdp): decode the base64 postData a continueRequest carries

ntdatt812:
## What

`Fetch.continueRequest`'s `postData` is passed through the wire encoding it arrives in. Decode it, as the `fulfillRequest` arm four lines below already does for its `body`.

## Why

`postData` is declared `binary` in the Fetch domain, so a client sends it base64-encoded. `server.rs` kept the string as-is:

```rust
body: req.params.get("postData").and_then(|v| v.as_str()).map(|s| s.to_string()),
```

and `ops.rs:2458` turns that into the outgoing body with `override_body = body.map(String::into_bytes)`. So the request body became the base64 text itself — a server reading a rewritten POST got neither the JSON nor the form the client meant to send, and no error anywhere said so.

The asymmetry is the argument: `Fetch.fulfillRequest`'s `body` is the same wire type and is already decoded through `decode_base64` in the very next arm.

Related to #569 — see my comment there. This is only one of the two ways an override goes missing, and deliberately the small one; what it does **not** address is below.

## Testing

Run in a Linux container on rustc 1.98:

```
cargo nextest run -p obscura-cdp -E 'test(continue_request) + test(parse_cdp_headers) + test(fetch_resolution_is_handled_once)'
5 tests run: 5 passed, 154 skipped
```

Reverting the one-line change to `.map(|s| s.to_string())` fails the new test with the bug stated literally:

```
assertion `left == right` failed: the request body must be the bytes the client encoded, not the base64 text
  left:  Some("eyJhIjoxfQ==")
 right:  Some("{\"a\":1}")
2 tests run: 1 passed, 1 failed
```

The second new test — an absent `postData` must stay `None` rather than becoming an empty body — passes either way by design; it is the guard against over-correcting this into always sending a body.

## What this does not fix

Two things, so nobody assumes they came with it:

- **The navigation path.** `Page.navigate` awaits the navigation to completion and only then synthesises `Fetch.requestPaused` from the drained network events, so for a navigation the pause is reported after the request is over. Fixing that means pausing before the request is issued, and a `url` rewrite then has to re-enter the same SSRF/private-network gate the original request passed.
- **`domains/fetch.rs`.** Its `Fetch.*` handlers resolve through `ctx.fetch_intercept.paused`, which nothing ever inserts into, so they return `Ok(json!({}))` without applying anything. That arm also passes `headers: None` unconditionally. The live path is `handle_fetch_resolution` in `server.rs`, which this PR touches.

Both are described in more detail on #569.

### DIFF
--- a/crates/obscura-cdp/src/server.rs
+++ b/crates/obscura-cdp/src/server.rs
@@ -1234,7 +1234,11 @@ fn handle_fetch_resolution(
                     url: req.params.get("url").and_then(|v| v.as_str()).map(|s| s.to_string()),
                     method: req.params.get("method").and_then(|v| v.as_str()).map(|s| s.to_string()),
                     headers: parse_cdp_headers(&req.params),
-                    body: req.params.get("postData").and_then(|v| v.as_str()).map(|s| s.to_string()),
+                    // postData is CDP `binary`, so it arrives base64-encoded --
+                    // the same wire type as fulfillRequest's `body` just below,
+                    // which is decoded. Passing it through raw sent the base64
+                    // text itself as the request body.
+                    body: req.params.get("postData").and_then(|v| v.as_str()).map(decode_base64),
                 },
                 "Fetch.fulfillRequest" => {
                     let status = req.params.get("responseCode").and_then(|v| v.as_u64()).unwrap_or(200) as u16;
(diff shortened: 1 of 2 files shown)


yinnho:
Right fix — we shipped the same decode in our tree when wiring #569. Two contract notes from that implementation, since this PR routes through the shared `decode_base64` helper:

One, the helper returns `String` via `from_utf8_lossy`, so a base64'd binary POST (a protobuf upload, a gzipped request body, a multipart file part) comes out with U+FFFD replacement characters before `into_bytes()` re-serializes them — the request-side twin of #754, which this repo fixed for responses by keeping bytes as bytes. We carry `postData` and `fulfillRequest`'s `body` as `Vec<u8>` from decode to wire for the same reason; the asymmetry argument in the PR description applies to the end type too, not just the decoding.

Two, the helper decodes leniently: `filter_map` drops every non-alphabet character and zero-pads partial chunks, so raw text sent where base64 was expected becomes quietly mangled bytes with no error anywhere — the exact "no error anywhere said so" shape the description argues against. Chrome answers these with a protocol error; we reply `continueRequest postData is not valid base64: <e>` (mirrored for `fulfillRequest`), pinned by a test feeding `"!!!not-base64"` to both arms. The detail that makes strictness safe: the failed resolution doesn't consume the pause, so a valid answer on the same `requestId` afterwards still drives the page fetch — a client mistake becomes visible and recoverable instead of wedged.

ntdatt812:
Both notes were right, and the first one lands squarely on my own test name — it asserts "the request body must be the bytes the client encoded" while the decode still went through a `String`. Pushed for both.

**Bytes, not `String`.** `InterceptResolution::Continue.body` is `Option<Vec<u8>>` now. It turns out to make the consumer simpler rather than wider: `override_body` at `ops.rs:2408` is already `Option<Vec<u8>>`, so `override_body = body.map(String::into_bytes)` becomes `override_body = body`. Pinned by a test that carries `AP/+` — the bytes `00 FF FE`, which are not valid UTF-8 — and asserts they arrive byte for byte. Under the old path they came out as three U+FFFD.

**Strict, and it does not wedge.** `decode_base64` decoded leniently and returned `String`; nothing calls it now, so it is deleted rather than left beside the new one. In its place `decode_optional_binary(params, field, what)` returns `Ok(None)` for an absent field, `Ok(Some(bytes))`, or an error naming the field, and both arms answer a bad value with `-32602` and `continueRequest postData is not valid base64: …` (mirrored for `fulfillRequest`).

The recovery property you named needed a reorder, and it is worth spelling out because the old shape was one line away from a wedge: `handle_fetch_resolution` did `intercepted_paused.remove(request_id)` *before* parsing, so returning an error there would have taken the pause with it and left the request unanswerable. The resolution is now built first and the pause is only taken once it is `Ok`. The test drives exactly that: `"!!!not-base64"` is rejected, nothing is sent to the page, the entry is still in the map, and a corrected `continueRequest` on the same `requestId` afterwards resolves normally.

Padding is optional in the decoder — encoders disagree about it and the group length is unambiguous without it — but a character outside the alphabet, padding with anything after it, and a final group of one symbol are all refused.

**One place I did not follow you**, so it is not mistaken for an oversight: `fulfillRequest`'s body is still a `String` at the resolution boundary. It is not the same one-line change as `postData`, because that body crosses a JSON reply at `ops.rs:2430` (`"body": b`) where a `Vec<u8>` serialises as an array of numbers and the page-side consumer reads a string. Making it bytes end to end means changing that reply's shape too, which is a wider change than this PR and belongs with #754's response-side work rather than bolted onto a `postData` fix. The strictness half is mirrored today; the byte half is the one left, and I would rather it were its own change than a silent half-move here.

Verified in a Linux container on rustc 1.98: `cargo nextest run -p obscura-cdp` over these arms is 8 passed, 0 failed, and `cargo fmt --check` reports no diff in either changed file.

Each new claim was checked by breaking it. Putting the bytes back through `String::from_utf8_lossy` fails only the binary test, with the corruption spelled out:

```
assertion `left == right` failed: a binary body must reach the wire byte for byte
  left:  Some([0, 239, 191, 189, 239, 191, 189])
 right:  Some([0, 255, 254])
```

`239, 191, 189` is U+FFFD twice, which is your first note as a number.

Taking the pause on the error path instead of leaving it fails only the two rejection tests, on `the request must stay paused`. Nothing else in the selection moves in either case.

ntdatt812:
Rebased onto `main` (`1a3169d`). Was CONFLICTING, now clean.

`#912` landed in the meantime and changed what this had to be. It added `body_base64` to `Fetch.fulfillRequest`, carrying the CDP body through unchanged so `op_fetch_url` can hand JS the exact bytes — which is precisely the lossy end this pull request's own comment said was out of scope. So your version of that arm is kept, not mine; `body_base64` and the lossy `body` view both survive.

What is kept from here is the strict decode, now layered on top rather than replacing it:

- `Fetch.continueRequest` reads `postData` through `decode_optional_binary`, so raw text in a base64 field is refused with `-32602` instead of reaching the wire as mangled bytes. `Ok(None)` still means "leave this alone" rather than "make it empty".
- `Fetch.fulfillRequest` keeps your shape and decodes `raw_body` to check it: a body that is not base64 would otherwise reach JS as the raw string through `body_base64` anyway, so it is refused at the boundary.
- `decode_base64` stays, as a lossy wrapper over `decode_base64_bytes`, for `domains/fetch.rs:164` which still wants the text view.

One thing I could not run: `cargo build` needs to create a symlink while unpacking the `v8` crate, which Windows refuses without the privilege, so this is not compiled here. The file parses (`rustfmt`), and the one thing the rebase genuinely broke — the resolution `match` became a `Result` in this branch while `main`'s `fulfillRequest` arm still returned the bare value — is fixed. CI compiles it.

PR #889, opened 2026-09-08
fix(cdp): emit navigation events for native form submissions

### LINKED ISSUE #886: Native <form> submission produces no CDP navigation/network events at all
### Before you start

### obscura version or commit

h4ckf0r0day/obscura:0.2.1 (Docker image)

### OS and architecture

Linux x86_64

### Affected area

CDP, Puppeteer, or Playwright compatibility

### Build variant

I do not know

### Stealth mode

None

### Reproduction

Environment
  - Obscura: h4ckf0r0day/obscura:0.2.1 (Docker image)
  - Client: Playwright Java 1.62.0, connected via
    connectOverCDP(`ws://<ip>:9222/devtools/browser`)
  - Obscura launched with serve --port 9222 --host 0.0.0.0 --allow-private-network

Summary
  Clicking a real `<form>`'s submit button (a genuine browser-native navigation, not a
  CDP-commanded Page.navigate) never produces any observable CDP event, even though the
  underlying HTTP request demonstrably happens and completes successfully.

Steps to reproduce
  1. Connect Playwright to Obscura over CDP.
  2. Navigate to a page with a standard <form method="POST" action="/some-endpoint"> and a
     submit button.
  3. Fill the form's fields, then click the submit button (`Locator.click()`, or a native DOM
     `.click()` via `page.evaluate()`).
  4. Independently confirm, via the target server's own access log, that the POST request
     arrives and is fully processed (in our case, in ~150ms, with a normal 200 response).

### Expected behavior

  Obscura should emit the standard CDP events for the resulting navigation and its network
  request — e.g. Network.requestWillBeSent/Network.responseReceived for the POST, and
  Page.frameNavigated/Page.lifecycleEvent for the new document — the same way a spec-compliant
  CDP target (real Chromium) does for any form-driven navigation, regardless of whether it was
  CDP-commanded.

### Actual behavior

  No such events are ever emitted:
  - Locator.click() on the submit button hangs indefinitely (30s+), reporting "waiting for
    navigation to finish" for the page's own current URL — not a new one.
  - A subsequent page.waitForURL(...) call also times out completely, never observing any URL
    change.
  - A subsequent page.evaluate(...) — even something trivial like reading document.title —
    throws Execution context was destroyed, most likely because of a navigation, implying the
    real page did navigate away underneath the CDP session, but no corresponding event was ever
    sent to the client.
  - Inspecting Playwright's own recorded network trace shows zero record of the POST request at
    all — not pending, not failed, simply absent — despite the server-side proof that it
    occurred.

### Rendering details

_No response_

### Logs

```shell

```

lisa0314:
## Summary

- Emit the complete CDP network and page lifecycle sequence for native form navigation triggered through `Input.dispatchMouseEvent`, including the actual POST request method and submitted body.
- Classify document replacement from pending navigation state before processing it. History-only changes emit `Page.navigatedWithinDocument` and retain the loader and default/isolated execution contexts; document replacement keeps the full lifecycle.
- Retain runtime network-event forwarding for clicked and evaluated same-document navigation.

The pending-document classification incorporates Antonio Rossi's `69cd7c8ef60a48905b6efe72ea591811b9e5cefc`, preserved as an author-attributed commit. Additional coverage checks both history paths and data-URL replacement. The two history failures also reproduce on main without the original form-navigation patch; the data-URL case already passed before this follow-up.

## Testing

Validated against main `89964b3`:

- Focused release nextest: 9/9 with render and 9/9 without render, covering form submission, page-frame contracts, and same-document navigation.
- Actual browser-WebSocket probes pass clicked/evaluated history and data-URL replacement. Playwright `connectOverCDP` checks pass both history paths, POST form navigation, and subsequent locator use after data-URL replacement.
- Release CLI/worker builds with render and without default features pass. Stealth configurations were not tested in this follow-up.
- Full release/render nextest: 1,710 passed, 2 failed, 4 skipped; lower-concurrency rerun: 1,711 passed, 1 failed, 4 skipped. Browser asynchronous-font/resource timing failures were also reproduced on a clean, rebuilt main. These runs are not a full-suite pass.
- Original obstacle course: 32/33. Clean main also fails `observer-intersection`.

Fixes #886

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -3300,7 +3300,7 @@ impl Page {
         let main_is_binary = !is_text_like_content_type(response.content_type());
         self.record_network_event_with_body(
             url.as_str(),
-            "GET",
+            method,
             "Document",
             response.status,
             &response.headers,
--- a/crates/obscura-cdp/src/domains/input.rs
+++ b/crates/obscura-cdp/src/domains/input.rs
@@ -241,35 +241,72 @@ pub async fn handle(
                     // with no document fetch. The client still has to be told the
                     // frame moved, or the click looks like it did nothing.
                     if moved {
+                        page.sync_js_network_events();
                         let url = page.url_string();
                         let frame_id = page.frame_id.clone();
-                        Some((page.id.clone(), frame_id, url))
+                        let network_events = page.network_events.drain(..).collect::<Vec<_>>();
+                        let reached_network_idle = page.lifecycle.is_network_idle();
+                        Some((
+                            page.id.clone(),
+                            frame_id,
+                            url,
+                            network_events,
+                            reached_network_idle,
+                        ))
                     } else {
                         None
                     }
                 } else {
                     None
                 };
-                if let Some((page_id, frame_id, url)) = moved_frame {
-                    let loader_id = ctx
-                        .current_loader_ids
-                        .get(&page_id)
-                        .cloned()
-                        .unwrap_or_else(|| format!("loader-blank-{page_id}"));
-                    ctx.pending_events.push(crate::types::CdpEvent {
-                        method: "Page.frameNavigated".into(),
-                        params: json!({
-                            "frame": crate::domains::page::frame_value(
-                                &frame_id,
-                                None,
-                                &loader_id,
-                                &url,
-                                "text/html",
-                            ),
-                            "type": "Navigation",
-                        }),
-                        session_id: Some(session_id.clone().unwrap_or_default()),
-                    });
+                if let Some((page_id, frame_id, url, network_events, reached_network_idle)) =
+                    moved_frame
+                {
+                    let is_document_navigation = network_events
+                        .iter()
+                        .any(|event| event.resource_type == "Document" && event.url == url);
+                    if !is_document_navigation {
+                        let loader_id = ctx
+                            .current_loader_ids
+                            .get(&page_id)
+                            .cloned()
+                            .unwrap_or_else(|| format!("loader-blank-{page_id}"));
+                        crate::domains::page::emit_runtime_network_events(
+                            ctx,
+                            session_id,
+                            &frame_id,
+                            &url,
+                            &page_id,
+                            &network_events,
+                        );
+                        ctx.pending_events.push(crate::types::CdpEvent {
+                            method: "Page.frameNavigated".into(),
+                            params: json!({
+                                "frame": crate::domains::page::frame_value(
+                                    &frame_id,
+                                    None,
+                                    &loader_id,
+                                    &url,
+                                    "text/html",
+                                ),
+                                "type": "Navigation",
+                            }),
+                            session_id: Some(session_id.clone().unwrap_or_default()),
+                        });
+                    } else {
+                        let loader_id = format!("loader-{}", uuid::Uuid::new_v4());
+                        crate::domains::page::emit_navigation_events(
+                            ctx,
+                            session_id,
+                            &frame_id,
+                            &loader_id,
+                            &url,
+                            &page_id,
+                            &network_events,
+                            obscura_browser::lifecycle::WaitUntil::Load,
+                            reached_network_idle,
+                        );
+                    }
                 }
             } else if event_type == "mouseWheel" {
                 let delta_x = params.get("deltaX").and_then(|v| v.as_f64()).unwrap_or(0.0);
--- a/crates/obscura-cdp/tests/form_submit_method_bypasses_listener.rs
+++ b/crates/obscura-cdp/tests/form_submit_method_bypasses_listener.rs
@@ -10,6 +10,7 @@ use obscura_cdp::types::CdpRequest;
 use serde_json::{json, Value};
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::net::TcpListener;
+use tokio::sync::oneshot;
 
 // Serves a form whose `submit` listener always calls preventDefault().
 async fn serve_form() -> String {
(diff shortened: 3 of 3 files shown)


lkraider:
I independently reproduced the click-submit lifecycle timeout on the v0.2.2 macOS render+stealth release using a localhost form and Playwright expect_navigation(domcontentloaded). Fill, raw Input.insertText, and Enter-submit pass in the same fixture; click-submit times out. Chromium passes all of them. I am testing this focused patch on our v0.2.2 integration branch instead of carrying our older, broader navigation patch or opening a duplicate PR. I will report the integration result after the build completes.

lkraider:
Integration result: this patch fixes the localhost Playwright click-submit timeout on our v0.2.2 macOS render+stealth build. Fill, raw insertText, Enter-submit and a second navigation still pass. All 11 fixture actions match Chromium. The integration also carries a separate by-value undefined fix (#938), which is required for no-match locator.count(); no other input/navigation patches are present. Native build and focused Rust tests passed: https://github.com/lkraider/obscura/actions/runs/34359904293 . Broader Linux suites are still running.

AntonioRossi:
I independently reproduced the missing click-navigation lifecycle and POST-as-GET metadata while integrating Playwright MCP 0.0.80 with Obscura v0.2.2 and then main `eec047a`. I found this PR while preparing an upstream contribution. The POST correction overlaps exactly, so I am linking our additional reproduction and implementation for consideration here.

One case worth adding is document replacement through a `data:` link, which produces no HTTP document request. With Playwright connected over CDP and a page available:

```js
const destination = 'data:text/html,' + encodeURIComponent('<h1>Destination</h1>');
await page.goto('data:text/html,' + encodeURIComponent('<a id="next" href="https://example.com/">Next</a>'));
await page.evaluate(url => document.querySelector('#next').setAttribute('href', url), destination);
const cdp = await page.context().newCDPSession(page);
const events = [];
cdp.on('Runtime.executionContextsCleared', () => events.push('cleared'));
cdp.on('Runtime.executionContextCreated', () => events.push('created'));
await cdp.send('Runtime.enable');
events.length = 0;
await page.locator('#next').click();
console.log(events);
console.log(await page.evaluate(() => document.body.textContent));
await cdp.detach();
```

On unmodified main in our diagnostic, the URL changes but the click emits no execution-context replacement events; the subsequent evaluation fails with `utilityScript.evaluate is not a function`. In our integration repair, context replacement is emitted and the final result is `Destination`.

Our implementation classifies replacement-document navigation from pending navigation state before processing it, rather than the presence of an HTTP `Document` event. It also emits `Page.navigatedWithinDocument` for history-only changes and preserves execution contexts in that case:
https://github.com/AntonioRossi/obscura/commit/9e7c4281863ba9e3aaf8166bc9ec307a1a257f8e

The combined integration branch has passing Policy, Core, and Performance CI, 1,704 local render tests, five MCP compatibility checks, and eight MCP smoke checks: https://github.com/AntonioRossi/obscura/pull/1 . Those results apply to our integration branch; I have inspected but have not run the head of this PR. The `data:` case may be useful coverage for the document-event classification here.

AntonioRossi:
I have now executed this PR's head `7294535299b20413725c2797cde6f0ed10a3edf9` on current main `8819c284cb5b451ef0f369f660a326aed2571678`, preserving its authorship.

The comparison plus added regressions passes 6/8 cases. Form POST body/method/lifecycle and the current data-URL click case pass. Two history cases still fail: a routed click emits `Page.frameNavigated` instead of `Page.navigatedWithinDocument`, and evaluated `history.pushState` clears execution contexts.

A focused residual repair is available in [69cd7c8](https://github.com/AntonioRossi/obscura/commit/69cd7c8). It captures pending document-navigation state before processing it, then preserves the loader and contexts for history-only changes. The comparison branch is [fix/navigation-upstream-comparison](https://github.com/AntonioRossi/obscura/tree/fix/navigation-upstream-comparison), final tested head `0aeaac4a1feffb1f5f0de903f1e08ff09158b998`. Added tests are in `page_frame_contract.rs`; the later test-only commits select the main default context and compare the JS number without integer/float JSON assumptions.

At that final head, all 8 focused cases pass in render, render+stealth and both no-render variants. Full release/render nextest: 1,709 passed, 4 skipped. All four CLI/worker builds pass. Actual Playwright checks with stealth active pass form submission followed by locator/evaluation, evaluated and clicked history state retention, and a usable replacement data-URL document. Original obstacle course remains 32/33 (same observer fixture failure); separately corrected fixture `3184db3` passes 33/33. Five-round performance comparison against frozen upstream with matched render/stealth builds passes unchanged thresholds: DOM 64.0→63.9 ms / 49.8→50.8 MiB, storage 15.6→15.6 ms / 35.2→36.8 MiB.

This corrects the scope of my earlier data-URL concern: the current comparison passes that case; the reproduced remaining failures are the history cases. I am carrying this PR plus the residual repair in our migration branch while review remains open, and offering the small follow-up here rather than opening a duplicate form-navigation PR.

lisa0314:
Thanks @lkraider for the integration verification, and @AntonioRossi for the follow-up tests and focused repair.

I reproduced both remaining history cases with this PR's patch applied to main `89964b3`: a routed click emits `Page.frameNavigated`, and evaluated `history.pushState` clears execution contexts. Both also fail without this PR's changes. The data-URL replacement case already passes, matching your correction.

I have prepared a local follow-up incorporating the pending-document navigation classification from your `69cd7c8`, with attribution, while retaining runtime network-event forwarding on the evaluated same-document path. History-only changes now emit `Page.navigatedWithinDocument` and preserve the loader and execution contexts. Document replacements retain the full navigation lifecycle.

Validation:
- All 9 focused tests pass with and without render, including new regressions for clicked history, evaluated history, and data-URL replacement.
- Browser-WebSocket checks pass. Playwright checks pass for both history paths, POST form navigation, and subsequent locator use after data-URL replacement.
- The release/render CLI build passes.

The full release/render suite is not green: the initial run had 1,710 passed, 2 failed, and 4 skipped; a lower-concurrency rerun had 1,711 passed, 1 failed, and 4 skipped. The failing browser timing/resource tests were also reproduced on clean main. The original obstacle course remains 32/33, with `observer-intersection` also failing on main.

I propose adding this focused follow-up to the existing PR rather than opening a duplicate navigation PR.

PR #891, opened 2026-09-08
fix(render): never let page-owned layout open synchronous HTTP requests

### LINKED ISSUE #890: render: layout/paint open direct, unproxied, synchronous requests for unknown images and block CDP input for 10 s+ per image
### Before you start

- [x] I searched existing issues and this is not a duplicate (#662 is about *which* CSS resources the transport prefetches; this is about layout/paint opening their own requests outside the transport).
- [x] I tested against the latest release or `main` (reproduced with the v0.2.2 stealth release binary; the code paths involved are unchanged on `main` at 727cc46).

### obscura version or commit

v0.2.2 (`a1e09de6`), official `obscura-x86_64-linux-stealth.tar.gz`

### OS and architecture

Debian 13 x86_64 (also inside a python:3.13-slim container)

CDP server with Puppeteer or Playwright (own CDP client)

### Affected area

Rendering, layout, or paint

### Build variant

Render-enabled with stealth (official stealth archive or --features render,stealth)

### Stealth mode

On (--stealth)

### Reproduction

A page-owned runtime still falls back to the renderer's synchronous compatibility loader (`HttpResourceLoader` -> `ureq`, `crates/obscura-render/src/paint.rs`) for every image, CSS `url()` or font the page transport has not seeded yet. That loader

- ignores `OBSCURA_PROXY`, the cookie jar, request/response callbacks, interception and `Network.setBlockedURLs` (it also ignores `HTTP(S)_PROXY`);
- blocks the V8 thread for the whole request: 10 s timeout x up to 3 attempts with backoff, serially per unknown URL;
- is reached by any synchronous geometry read, so `Input.dispatchMouseEvent` (`mouseWheel` calls `elementFromPoint`), `getBoundingClientRect`, `getComputedStyle` flushes, ResizeObserver and `Runtime.evaluate` captures all stall.

Only the capture entry points are protected (`with_sync_render_loading_disabled`). `Page::prepare_screenshot_resources` seeds the cache through the transport at navigation/settle, but with a deadline that drops unfinished requests, so any asset slower than ~1 s stays unknown and hits the loader on the next layout.

Self-contained repro (Linux, root for `unshare --net`, Python 3 stdlib, no external network). It serves a document with one `` whose origin answers after 12 s, puts an authenticated forward proxy in front, starts `obscura --stealth serve` with `OBSCURA_PROXY`, navigates and sends `mouseMoved` + `mouseWheel`:

```bash
sudo OBSCURA_BIN=/path/to/obscura python3 repro_sync_render_loader.py
```

<details>
<summary>repro_sync_render_loader.py</summary>

```python
#!/usr/bin/env python3
"""Repro: layout/paint open a direct, unproxied, synchronous request for an
unknown image and block Input.dispatchMouseEvent for the whole request.

Linux, root (uses `unshare --net`), Python 3 stdlib only, no external network.
Usage: sudo OBSCURA_BIN=/path/to/obscura python3 repro_sync_render_loader.py

Topology inside a fresh network namespace (loopback only):
  - 93.184.216.34 is added to `lo` so obscura's SSRF guard treats it as public
  - :18443  TLS document origin  (https://example.test/  -> <img src=http://93.184.216.34:18080/slow.svg>)
  - :18081  forward proxy (CONNECT for the document, plain GET for the image); requires Proxy-Authorization
  - :18080  image origin, answers after 12 s
obscura runs with OBSCURA_PROXY pointing at :18081. Every request the image
origin receives directly on :18080 bypassed the proxy.
"""
import json, os, socket, ssl, subprocess, sys, tempfile, threading, time
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
from pathlib import Path

PUBLIC = "93.184.216.34"
SVG = b'<svg xmlns="http://www.w3.org/2000/svg" width="20" height="10"><rect width="20" height="10" fill="#f00"/></svg>'
HTML = f'<!doctype html><html><body><img src="http://{PUBLIC}:18080/slow.svg"></body></html>'.encode()
seen = []  # (port, path, seconds, proxy_auth_present)
t0 = time.monotonic()


class Handler(BaseHTTPRequestHandler):
    def log_message(self, *_):
        pass

    def do_CONNECT(self):  # proxy: tunnel to the TLS document origin
        upstream = socket.create_connection((PUBLIC, 18443))
        self.send_response(200, "Connection established")
        self.end_headers()
        client = self.connection

        def pump(src, dst):
            try:
                while data := src.recv(65536):
                    dst.sendall(data)
            except OSError:
                pass
            finally:
                try:
                    dst.shutdown(socket.SHUT_WR)
                except OSError:
                    pass

        threading.Thread(target=pump, args=(client, upstream), daemon=True).start()
        pump(upstream, client)

    def do_GET(self):
        path = self.path.split("?")[0]
        if path.startswith("http"):
            path = "/" + path.split("/", 3)[3]
        seen.append((self.server.server_port, path, round(time.monotonic() - t0, 2),
                     self.headers.get("Proxy-Authorization") is not None))
        if path.endswith(".svg"):
            time.sleep(12)
            payload, ctype = SVG, "image/svg+xml"
        else:
            payload, ctype = HTML, "text/html"
        self.send_response(200)
        self.send_header("Content-Type", ctype)
        self.send_header("Content-Length", str(len(payload)))
        self.end_headers()
        try:
            self.wfile.write(payload)
        except (BrokenPipeError, ssl.SSLError):
            pass


def cdp(ws, msg, timeout):
    """Minimal CDP call over a raw websocket (no dependencies)."""
    import base64, struct
    data = json.dumps(msg).encode()
    mask = os.urandom(4)
    head = bytearray([0x81])
    if len(data) < 126:
        head.append(0x80 | len(data))
    else:
        head += bytes([0x80 | 126]) + struct.pack(">H", len(data))
    ws.sendall(bytes(head) + mask + bytes(b ^ mask[i % 4] for i, b in enumerate(data)))
    ws.settimeout(timeout)
    buf = b""
    while True:
        try:
            chunk = ws.recv(65536)
        except socket.timeout:
            return None
        buf += chunk
        while len(buf) >= 2:
            ln = buf[1] & 0x7F
            off = 2
            if ln == 126:
                ln = struct.unpack(">H", buf[2:4])[0]; off = 4
            elif ln == 127:
                ln = struct.unpack(">Q", buf[2:10])[0]; off = 10
            if len(buf) < off + ln:
                break
            frame, buf = buf[off:off + ln], buf[off + ln:]
            reply = json.loads(frame)
            if reply.get("id") == msg["id"]:
                return reply


def main():
    if "--inside" not in sys.argv:
        return subprocess.call(["unshare", "--net", sys.executable, __file__, "--inside"])
    subprocess.run(["ip", "link", "set", "lo", "up"], check=True)
    subprocess.run(["ip", "addr", "add", PUBLIC + "/32", "dev", "lo"], check=True)
    with tempfile.TemporaryDirectory() as d:
        cert, key = Path(d) / "cert.pem", Path(d) / "key.pem"
        subprocess.run(["openssl", "req", "-x509", "-newkey", "rsa:2048", "-nodes", "-days", "1",
                        "-subj", "/CN=example.test", "-addext", "subjectAltName=DNS:example.test",
                        "-keyout", str(key), "-out", str(cert)], check=True,
                       stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
        servers = [ThreadingHTTPServer(("0.0.0.0", p), Handler) for p in (18443, 18081, 18080)]
        tls = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
        tls.load_cert_chain(cert, key)
        servers[0].socket = tls.wrap_socket(servers[0].socket, server_side=True)
        for s in servers:
            threading.Thread(target=s.serve_forever, daemon=True).start()
        port = 19222
        proc = subprocess.Popen(
            [os.environ.get("OBSCURA_BIN", "obscura"), "--stealth", "serve", "--host", "127.0.0.1", "--port", str(port)],
            env={"PATH": os.environ["PATH"], "RUST_LOG": "off", "SSL_CERT_FILE": str(cert),
                 "OBSCURA_PROXY": f"http://user:pass@{PUBLIC}:18081",
                 # resolve example.test to the fixture through the proxy's CONNECT
                 "OBSCURA_TIMEZONE": "UTC"},
            stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
        try:
            for _ in range(100):
                try:
                    ws = socket.create_connection(("127.0.0.1", port), timeout=1); break
                except OSError:
                    time.sleep(0.05)
            ws.sendall(f"GET /devtools/browser HTTP/1.1\r\nHost: 127.0.0.1:{port}\r\nUpgrade: websocket\r\nConnection: Upgrade\r\nSec-WebSocket-Key: AAAAAAAAAAAAAAAAAAAAAA==\r\nSec-WebSocket-Version: 13\r\n\r\n".encode())
            ws.settimeout(3)
            while b"\r\n\r\n" not in ws.recv(4096):
                pass
            target = cdp(ws, {"id": 1, "method": "Target.createTarget", "params": {"url": "about:blank"}}, 5)["result"]["targetId"]
            session = cdp(ws, {"id": 2, "method": "Target.attachToTarget", "params": {"targetId": target, "flatten": True}}, 5)["result"]["sessionId"]
            cdp(ws, {"id": 3, "method": "Page.enable", "sessionId": session}, 5)
            nav = cdp(ws, {"id": 4, "method": "Page.navigate", "params": {"url": "https://example.test/"}, "sessionId": session}, 30)
            print("navigate:", "ok" if nav and "result" in nav else nav, flush=True)
            cdp(ws, {"id": 5, "method": "Input.dispatchMouseEvent", "params": {"type": "mouseMoved", "x": 50, "y": 30}, "sessionId": session}, 8)
            started = time.monotonic()
            wheel = cdp(ws, {"id": 6, "method": "Input.dispatchMouseEvent",
                             "params": {"type": "mouseWheel", "x": 50, "y": 30, "deltaX": 0, "deltaY": 150}, "sessionId": session}, 8)
            elapsed = round((time.monotonic() - started) * 1000)
            print("mouseWheel:", "answered in %d ms" % elapsed if wheel else "NO ANSWER within 8000 ms", flush=True)
            time.sleep(3)
        finally:
            proc.kill()
            for s in servers:
                s.shutdown()
        direct = [r for r in seen if r[0] == 18080]
        proxied = [r for r in seen if r[0] == 18081 and r[1].endswith(".svg")]
        print(json.dumps({"image_requests_via_proxy": len(proxied), "image_requests_direct_to_origin": len(direct),
                          "direct_requests": direct}, indent=1))
        return 1 if direct or not wheel else 0


if __name__ == "__main__":
    raise SystemExit(main())
```

</details>

### Expected behavior

- `Input.dispatchMouseEvent` answers within milliseconds; an image that has not arrived yet lays out as a placeholder (like Chromium before the first bytes) and gets its intrinsic size when the transport delivers it.
- The image origin only ever sees requests through the configured proxy (with `Proxy-Authorization`, page cookies, interception, `Network.setBlockedURLs` applied).

### Actual behavior

```
navigate: ok
mouseWheel: NO ANSWER within 8000 ms
{
 "image_requests_via_proxy": 2,
 "image_requests_direct_to_origin": 2,
 "direct_requests": [[18080, "/slow.svg", 2.23, false], [18080, "/slow.svg", 12.48, false]]
}
```

- The wheel command is silent for 10 s+ (with a longer client timeout: the first direct request at +0 s, retries at +10.2 s and +20.6 s, i.e. `http_get_bytes`' 3 attempts with 200/400 ms backoff).
- The origin receives **direct** requests without `Proxy-Authorization`, carrying the renderer's own `Accept: image/webp,image/apng,image/svg+xml,image/png,image/jpeg,...` header instead of the page transport's.
- `Network.setBlockedURLs ["*slow.svg"]` before navigation: the proxied request disappears, the direct requests still happen.
- With 20 `` on an unreachable origin (no route) the same wheel takes about 0.6 s per URL (3 connect failures + backoff each), so a page with ~80 images stalls for a minute; and because misses are negative-cached for only 2 s, the stall repeats on every later layout.
- `SIGTERM` is not honored while the thread is blocked in the loader.

### Rendering details

Viewport 800x600 default, DPR 1. Not a capture bug: `Page.captureScreenshot` itself is guarded, the synchronous reads before it are not.

No: headless Chrome renders or behaves correctly and obscura does not (Chrome loads subresources through the configured proxy and never blocks input on a pending image).

### Logs

```shell
# RUST_LOG=obscura_browser=debug shows only the transport warmups; the direct request is not logged.
prepared screenshot resources through page transport loaded=0 elapsed_ms=1001
prepared screenshot resources through page transport loaded=0 elapsed_ms=1001
```

I have a fix (page-owned runtimes become cache-only; layout misses are loaded through the existing transport in the background and applied when they land, bound to the document's lifetime) with tests; PR follows.

michabbb:
## What changed

Fixes #890.

For a page that owns a transport (`ObscuraHttpClient` or the stealth client), layout and paint fell back to the renderer's synchronous compatibility loader (`HttpResourceLoader` -> `ureq`) for every image, CSS `url()` or font the page transport had not seeded yet. That loader bypasses the page's proxy, cookie jar, callbacks, interception and `Network.setBlockedURLs`, and it blocks the V8 thread for the whole request (10 s timeout, up to 3 attempts) serially per unknown URL, so any synchronous geometry read (`elementFromPoint` in `Input.dispatchMouseEvent`, `getBoundingClientRect`, `getComputedStyle` flushes, `Runtime.evaluate` captures) stalled for seconds. Only the capture entry points were guarded.

- `obscura-render`: `RenderResourceCache` records every cache-only miss with its request identity (network URL plus image CORS profile, or none for CSS images and fonts) and hands them out through `take_sync_misses`. That is the single source of what to load: nothing is reconstructed from the DOM, so script-registered `FontFace`s, `@font-face` rules inside shadow roots and any other renderer-only source are covered. No behaviour change while the compatibility loader is enabled.
- `obscura-js`: runtimes that own a page transport get a cache-only cache at every construction point (`fresh_render_resources` in `set_dom`/`take_dom`, `set_http_client`/`set_stealth_client`, and frame realms via `share_resources_with`). Standalone render runtimes without a transport keep the loader. The runtime owns the whole load path, so it never depends on the page getting control back: `take_render_resource_requests` turns misses into transport requests (URLs matching `Network.setBlockedURLs`, or `Fetch.enable` patterns while interception is on, and non-http URLs are seeded as missing, in-flight ones skipped; same policy and matcher as `Page::should_block_url`), `start_render_resource_loads` runs them on a background task through the page transport (plain or stealth client, callbacks, image CORS profile; one page-wide `Semaphore(16)` shared by every load of the document, `catch_unwind` per load so one failure cannot strand the batch), and `apply_render_resource_results` seeds arrived bytes through the existing `seed_render_*` entry points. The plain client or the stealth client alone is a transport (same rule as `has_page_transport`). Dropping the runtime aborts its loads too, so a directly embedded runtime does not leave detached requests behind. `service_render_resources` (apply, then start new misses) runs at every event-loop turn (`run_autonomous_event_loop_turn`, each cooperative tick of `run_event_loop_bounded`, `run_event_loop_until_quiescent`, `run_load_delaying_event_loop_tick`), inside `resolve_promises_until` (the `awaitPromise` wait) and before each `evaluate`, so a miss that a script creates while a command is waiting is loaded and observed inside that wait. Results carry `document_generation`; `set_dom`, `take_dom` and `abandon_render_resources` abort the tasks and replace the result channel, so a load past the point of no return (abort is not a join) delivers into a dropped receiver even when the runtime survives, as after a failed navigation.
- `obscura-browser`: the light-DOM scan of `prepare_screenshot_resources` stays as the navigation warmup only (`render_resource_candidates`, `git diff -w` shows it unchanged); it now feeds the runtime's loader. Loads run to completion instead of being dropped at the warmup deadline; `prepare_screenshot_resources` keeps its contract (wait up to `max_ms`, return the number of loads). `queue_pending_render_resources` (after each command, pump turn and settle iteration) services the runtime and records the Network events of applied loads; `drain_render_resource_results` applies without starting. `retire_render_resources` (from `navigate_single`, `navigate_blank`, `suspend_js`, `Drop`) abandons the document's loads. `run_autonomous_event_loop_turn` wakes on finished loads and parks on the next one when the runtime is idle instead of reporting busy, which would spin the connection pump; every wait registers its `Notified` (and `enable`s it) first, then drains the channel, then checks the pending state, so a load that finished before the registration is applied by the drain and one that finishes after it wakes the waiter (`notify_waiters` stores nothing for later waiters). `enable_intercept` and `init_js` hand the `Fetch.enable` patterns to the runtime.
- `obscura-cdp`: `service_live_page_render_resources` (drain + queue) before and after each dispatched command and after each autonomous pump turn.

No-render builds are unaffected (the new page methods are no-ops there). No timeout, retry or proxy change to `HttpResourceLoader`; no resource filtering beyond the page's existing block and interception policy; no change to input handling.

Semantics: an image whose bytes are not known yet lays out with its `width`/`height` attributes or as an empty box (as Chromium before the first bytes), `complete === false`, `naturalWidth === 0`; when the transport delivers the bytes the retained geometry is invalidated and the next read or paint has the real intrinsic size. CSS images and fonts follow the same path. `Network.setBlockedURLs` and `Fetch.enable` patterns now also stop renderer-initiated loads (the intercepted URL is settled as missing; nothing is fetched behind the client's back).

## Validation

Focused regression tests (all new tests fail on `main` without the fix):

- `obscura-render`: `cache_only_misses_are_reported_with_their_request_identity`
- `obscura-js`: `page_transport_keeps_render_resources_cache_only_across_document_resets`
- `obscura-js`: `late_channel_answer_of_a_retired_document_is_discarded` (a load answering after `abandon_render_resources` hits a closed channel; a stale-generation answer on the live channel neither seeds nor clears the live request), `stealth_only_transport_starts_render_resource_loads` (`--features render,stealth`: a runtime with only `set_stealth_client` starts and settles its loads)
- `obscura-browser`: `cache_only_layout_never_blocks_on_a_slow_asset_and_late_bytes_update_geometry` (1.2 s asset: layout answers immediately, exactly one transport request, later `naturalWidth`/layout box = 20), `blocked_render_resources_are_never_fetched_by_layout_or_transport`, `retired_document_loads_never_seed_the_next_document` (document A's in-flight slow response is abandoned when B replaces it; B requests the same URL itself, sees its own bytes and keeps them after A's would have landed; one request per document), `renderer_misses_cover_script_fonts_and_shadow_root_styles` (a `FontFace` added from script and an `@font-face` inside a shadow root are requested through the transport although no light-DOM scan can see them), `late_bytes_apply_while_the_runtime_waits_for_a_promise` (`resolve_promises_until` observes a 1.2 s image without any protocol command or page pump), `a_miss_created_inside_an_awaited_expression_loads_during_the_wait` (one `evaluate_for_cdp_with_timeout(awaitPromise)` expression adds an `@font-face`, reads geometry and polls it: the 1.2 s font is requested and applied inside that wait, width changes before the expression returns; fails without the in-wait servicing: no request at all), `timers_inside_a_fixed_wait_observe_bytes_that_land_during_it` (a `setInterval` sampling geometry during one `run_event_loop_for_duration(2200)` sees the font that lands at 1.2 s; fails with servicing only at loop entry: every sample stays at the fallback width), `renderer_misses_honour_fetch_interception_patterns` (`*intercepted.ttf` under `Fetch.enable` is never fetched and settled as missing, another font still loads; after `Fetch.disable` the URL loads), `a_load_that_finished_during_the_scan_does_not_cost_the_deadline` (two worker threads, a 1 ms font answered while the warmup scans a 120k-node document: `prepare_screenshot_resources(1000)` applies it and returns well before the deadline; with the wait registered after the drain it waited the full second), `render_resource_loads_share_one_page_wide_concurrency_limit` (three groups of 14 CSS images added and queued one after the other, each starting loads while the previous group is still running, never exceed 16 concurrent connections).

Commands run on this branch (Debian bookworm builder, rustc 1.98.0, clang):

```
cargo build --release -p obscura-cli --bins --features render
cargo build --release -p obscura-cli --bins --features render,stealth
cargo build --release -p obscura-cli --bins --no-default-features
cargo build --release -p obscura-cli --bins --no-default-features --features stealth
cargo nextest run --release --features render --no-fail-fast -p obscura-render -p obscura-js -p obscura-browser -p obscura-cdp -E 'test(cache_only) | test(page_transport) | test(blocked_render) | test(render_resource) | test(retired_document) | test(renderer_misses) | test(late_bytes) | test(navigation_post_script) | test(autonomous) | test(pump) | test(navigat)'   # 52 passed
cargo nextest run --release --no-default-features --no-fail-fast -p obscura-js -p obscura-browser -p obscura-cdp -E 'test(page_transport) | test(navigat) | test(autonomous) | test(pump) | test(settle)'   # 38 passed
OBSCURA_BIN=target/release/obscura python3 obstacle-course/run.py --runs 1 --warmup 0   # 32/33
```

Full `--features render` nextest suites of the four touched crates on this branch: 1358 passed, 0 failed, 4 skipped (unpatched `main` 727cc46 with the same command: 1350 passed, 0 failed, 4 skipped).

Obstacle course: 32/33 before and after. `observer-intersection` (`expected 'io:50', got ''`) fails identically with unpatched `main` (727cc46) and with the v0.2.2 release binary, so it is not related to this change.

Manual checks (offline, loopback proxy + origin in a network namespace, stealth build, `OBSCURA_PROXY` set), before -> after:

- issue repro (12 s image, `mouseMoved` + `mouseWheel` after `Page.navigate`): no answer within 8 s and 2 direct origin requests -> answered in 3 ms, 0 direct requests, 1 proxied request; the image later reports `naturalWidth` 20, lays out at 20 px and is painted in `Page.captureScreenshot`
- 20 `` on an unreachable origin: wheel 8 s+ (about 0.6 s per URL) -> 5 ms
- `Network.setBlockedURLs`: direct requests despite the block -> no request at all
- image added by script after load, late CSS `background-image`, second navigation with a fresh cache: all load through the proxy and render
- navigation while a `no-store` image of document A is still pending: document B fetched and rendered its own bytes; A's late response was discarded
- `Page.captureScreenshot`, `Runtime.evaluate` captures, a Playwright-style CDP smoke suite (65 checks) and a scraping flow with native input: unchanged results

## Rendering

Not a paint or layout rule change. Not-yet-loaded images render as before the first bytes (attribute box or empty box, alt text or placeholder), then with the real bytes once the transport delivers them. Viewport 800x600 / 100x80 in tests, DPR 1. No before/after images beyond the manual checks above; happy to add a `render-repros/` fixture if you want one for the delayed-image case.

## Performance

The hot path gains one `try_recv` and one miss-flag check per command, pump turn, event-loop tick and promise-wait tick; no DOM scan outside the navigation warmup. Background loads use the existing transport under one page-wide limit of 16 concurrent requests (the same bound the warmup stream always had), so a page with hundreds of resources cannot open more connections than before. Measured on the delayed-image fixture: wheel dispatch 10 s+ -> 3-7 ms; navigation time unchanged (the warmups still wait their 1 s budgets). Memory: one unbounded result channel per document that holds finished responses until the next turn, plus the applied responses' bodies kept once more in `render_resource_events` until the page records them as Network events at the next command or pump turn (each body at most the transport's response limit; `Network.setBlockedURLs` and interception patterns keep blocked bodies out entirely); the in-flight set and the task handles. All replaced on navigation and dropped with the runtime. No obstacle-course latency change (3.02 s median before and after, dominated by the runner's fixed 3 s wait).

## Checklist

- [x] Public API or user-facing behavior changes are documented (new `pub` page methods `retire_render_resources`, `spawn_pending_render_resources`, `drain_render_resource_results`, `queue_pending_render_resources`, `has_pending_render_resources`, `prepare_screenshot_resources` unchanged; runtime `set_intercept_block_patterns`, `has_page_transport`, `document_generation`, `take_render_resource_requests`, `mark_render_resources_in_flight`, `start_render_resource_loads`, `service_render_resources`, `apply_render_resource_results`, `take_render_resource_events`, `abandon_render_resources`, `has_pending_render_resources`, `render_resource_notify`, `render_resource_sync_loading_enabled`; cache `set_sync_loading_enabled`/`sync_loading_enabled`, `take_sync_misses`, `has_sync_misses`; `ops::RENDER_RESOURCE_CONCURRENCY`, `RenderResourceLoad`, `RenderResourceResponse`, `RenderResourceEvent`).

Known limitation: frame realms share the page transport and are now cache-only too, but frame geometry still resolves against the main document's renderer state (pre-existing), so frame-scoped background loading is not wired up; documented in `share_resources_with`. A runtime embedded directly with a transport but without a Tokio context keeps its misses in a backlog until the next event-loop turn inside one. Not in this change: the lifecycle op `op_load_image_metadata` and the page-level loads may both request the same URL once when script touches an image before the first request completes (as before between the op and `prepare_screenshot_resources`); unifying them is a follow-up. Related: #662 (which CSS resources the transport should prefetch) is orthogonal; this PR keeps the existing candidate scan.

### DIFF
--- a/crates/obscura-browser/src/page.rs
+++ b/crates/obscura-browser/src/page.rs
@@ -891,18 +891,6 @@ fn css_import_rule_len(css: &str) -> Option<usize> {
     None
 }
 
-fn render_resource_type(url: &url::Url) -> ResourceType {
-    let path = url.path().to_ascii_lowercase();
-    if [".woff", ".woff2", ".ttf", ".otf", ".eot"]
-        .iter()
-        .any(|extension| path.ends_with(extension))
-    {
-        ResourceType::Font
-    } else {
-        ResourceType::Image
-    }
-}
-
 /// Pull leading `@import` rules out of a stylesheet. Returns each import target
 /// URL with its optional media condition plus the CSS with those `@import`
 /// statements removed. Browsers fetch media-gated imports even when they do
@@ -1865,6 +1853,8 @@ impl Page {
         // runtime does not exist yet, so the new runtime would otherwise start
         // with interception disabled and op_fetch_url would never intercept.
         rt.set_intercept_enabled(self.intercept_enabled);
+        #[cfg(feature = "render")]
+        rt.set_intercept_block_patterns(self.intercept_block_patterns.clone());
         rt.set_runtime_events_enabled(self.runtime_events_enabled.get());
 
         if let Some(dom) = self.dom.take() {
@@ -3029,6 +3019,11 @@ impl Page {
                     let _ = js.run_event_loop_until_quiescent(remaining, 150).await;
                 }
             }
+            #[cfg(feature = "render")]
+            {
+                self.drain_render_resource_results();
+                self.queue_pending_render_resources();
+            }
             if !self.advance_frames().await {
                 break;
             }
@@ -3078,15 +3073,71 @@ impl Page {
     /// higher-priority automation commands.
     #[doc(hidden)]
     pub async fn run_autonomous_event_loop_turn(&mut self) -> Result<bool, String> {
+        #[cfg(feature = "render")]
+        {
+            self.drain_render_resource_results();
+            self.queue_pending_render_resources();
+        }
         let reached_idle = match self.js.as_mut() {
-            Some(js) => js.run_autonomous_event_loop_turn().await,
-            None => Ok(true),
-        }?;
+            Some(js) => {
+                #[cfg(feature = "render")]
+                {
+                    // A finished transport load is page work too: wake up as
+                    // soon as it lands so the next layout, paint or lifecycle
+                    // read sees the real bytes instead of waiting for the next
+                    // protocol command. The runtime turn is cancel-safe (the
+                    // connection processor already races it against commands).
+                    // The waiter is registered before the pending check, so a
+                    // load finishing in between cannot be missed.
+                    let notify = js.render_resource_notify();
+                    let notified = notify.notified();
+                    tokio::pin!(notified);
+                    notified.as_mut().enable();
+                    let loads_pending = js.has_pending_render_resources();
+                    tokio::select! {
+                        biased;
+                        turn = js.run_autonomous_event_loop_turn() => turn?,
+                        _ = &mut notified, if loads_pending => false,
+                    }
+                }
+                #[cfg(not(feature = "render"))]
+                js.run_autonomous_event_loop_turn().await?
+            }
+            None => true,
+        };
         // Dynamic iframe fetches finish on the page event loop, but their
         // realms must be built by Page between turns. Keep the autonomous CDP
         // pump on the same generic frame path as settle(), so a client that
         // stays attached can observe and run child documents as they arrive.
         let frame_work = self.advance_frames().await;
+        #[cfg(feature = "render")]
+        {
+            if reached_idle && !frame_work && self.js.is_some() {
+                // The runtime may have nothing to do until a resource lands.
+                // Park on the next result instead of reporting "busy" (which
+                // would spin the connection pump). `notified` is cancel-safe,
+                // so the processor can still preempt this turn with a
+                // command. Register the waiter first, then service (a load
+                // that finished before this point is applied here; its wake
+                // is already past), then decide whether to park.
+                let notify = self.js.as_ref().unwrap().render_resource_notify();
+                let notified = notify.notified();
+                tokio::pin!(notified);
+                notified.as_mut().enable();
+                self.queue_pending_render_resources();
+                if self.has_pending_render_resources() {
+                    notified.await;
+                    self.drain_render_resource_results();
+                    return Ok(false);
+                }
+                return Ok(true);
+            }
+            // Geometry this turn produced may have missed resources; start
+            // their loads now.
+            self.queue_pending_render_resources();
+            return Ok(reached_idle && !frame_work);
+        }
+        #[cfg(not(feature = "render"))]
         Ok(reached_idle && !frame_work)
     }
 
@@ -3205,6 +3256,8 @@ impl Page {
     ) -> Result<(), PageError> {
         let url = Url::parse(url_str).map_err(|e| PageError::InvalidUrl(e.to_string()))?;
 
+        // The previous document's background loads end with the document.
+        self.retire_render_resources();
         self.lifecycle = LifecycleState::Loading;
         self.referrer = referrer.to_string();
         self.url = Some(url.clone());
@@ -3544,6 +3597,7 @@ impl Page {
     }
 
     pub fn navigate_blank(&mut self) {
+        self.retire_render_resources();
         self.pending_frame_work.clear();
         self.frames.clear();
         self.js = None;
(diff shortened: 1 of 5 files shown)


michabbb:
Follow-up: #998 addresses the remaining synchronous resource loader used when taking a screenshot after `Page::suspend_js()`, based on current main after #949. It also tests the existing URL-blocklist paths independently. The details and a thank-you for the original hint are in #890.

Issue #779, opened 2026-08-31
RemoteObject for a primitive contradicts itself: value is a JSON string, and undefined comes back as null


ntdatt812:
### Before you start

- [x] I searched existing issues and this is not a duplicate.
- [x] I tested against the latest release or `main`.
- [x] This is not a security vulnerability (those go through SECURITY.md, not a public issue).

### obscura version or commit

`9326880` (main, 2026-08-30)

### OS and architecture

Linux CI container, rustc 1.98, x86_64. The defect is in `RemoteObjectInfo` construction and is not platform-specific.

### How are you using obscura?

CDP (Puppeteer / Playwright / raw WebSocket)

### Stealth mode

None (same code path in both modes)

### Reproduction

Dispatch `Runtime.evaluate` through the CDP layer and print the `result` object verbatim:

```rust
let mut ctx = CdpContext::new();
let page_id = ctx.create_page();
let sid = Some("probe".to_string());
ctx.sessions.insert(sid.clone().unwrap(), page_id);
let out = handle("evaluate",
    &json!({ "expression": expr, "returnByValue": by_value }),
    &mut ctx, &sid).await;
```

### Expected behavior

A primitive's `value` is that primitive, its `type` names what it is, and no `objectId` is minted for it — the shape Chrome sends, and the shape Puppeteer and Playwright read.

### Actual behavior

```
1 + 1      byValue=false -> {"type":"number","description":"2","objectId":"…","value":"2"}
1 + 1      byValue=true  -> {"type":"number","description":"2.0","value":2.0}
true       byValue=false -> {"type":"boolean","description":"true","objectId":"…","value":"true"}
'hi'       byValue=false -> {"type":"string","description":"hi","objectId":"…","value":"hi"}
undefined  byValue=false -> {"type":"undefined","objectId":"…","value":""}
undefined  byValue=true  -> {"type":"object","subtype":"null","description":"null","value":null}
1.5        byValue=false -> {"type":"number","description":"1.5","objectId":"…","value":"1.5"}
```

Four distinct problems:

**1. `value` is a JSON string for every primitive on the handle path.** `info_from_meta` fills it by copying the description:

```rust
let value = if js_type != "object" && js_type != "function" {
    meta.get("description")
        .and_then(|v| v.as_str())
        .map(|s| serde_json::Value::String(s.to_string()))
} else {
    None
};
```

[...]

PR #938, opened 2026-09-09
fix(cdp): preserve undefined in by-value results

lkraider:
## Change

Preserve JavaScript `undefined` before the by-value CDP result passes through JSON. `Runtime.evaluate` and both `Runtime.callFunctionOn` by-value paths now report `{"type":"undefined"}` instead of `{"type":"object","subtype":"null","value":null}`. Other values use the existing serializer.

Refs #779. This covers the by-value case explicitly left out of #780. It does not change the handle/metadata path, primitive object IDs, numeric descriptions, or unserializable values.

## Reproduction

On the v0.2.2 macOS render+stealth release, both methods collapse synchronous `undefined` and awaited `Promise.resolve(undefined)` into null. Awaited examples:

```js
await cdp.send('Runtime.evaluate', {
  expression: 'undefined', returnByValue: true, awaitPromise: true
});
await cdp.send('Runtime.callFunctionOn', {
  objectId: windowObjectId,
  functionDeclaration: 'function() { return Promise.resolve(undefined); }',
  returnByValue: true, awaitPromise: true
});
```

A local Playwright fixture also reproduces `locator('.missing').count()` throwing `value: expected integer, got object`. Playwright uses `undefined` as its no-match sentinel. Chromium reports zero for that locator and preserves the undefined CDP type.

## Validation

Four CDP regression tests cover synchronous and awaited results from both methods. Each also checks null, zero, false, empty string, object and array controls.

- [Paired fork validation](https://github.com/lkraider/obscura/actions/runs/34359497259) completed and its uploaded logs were reviewed. The four new tests fail on untouched v0.2.2 and pass on this patch in both release/render and release/no-render.
- Existing Linux render suite: 1,632 passed, 4 skipped on both baseline and candidate. Affected no-render suite (`obscura-cdp`, `obscura-js`): 540 passed, 3 skipped on both. New regression tests are run separately from these counts.
- Linux render and no-render CLI builds pass. macOS render+stealth build and focused CDP tests pass.
- Pinned obstacle course (`6ebac8293d7477f59e837768bfd4e74173f04f1c`): baseline and candidate both pass 32/33. Both fail only `observer-intersection` (`expected 'io:50', got ''`). This is a baseline failure, not a full obstacle-course pass. One run without warmup does not establish a performance result.

No network, input, navigation, feature-gate or rendering changes are included. The patch adds one V8 undefined check to the by-value result conversion.

### DIFF
--- a/crates/obscura-cdp/tests/runtime_by_value_undefined.rs
+++ b/crates/obscura-cdp/tests/runtime_by_value_undefined.rs
@@ -1,0 +1,81 @@
+use obscura_cdp::dispatch::{dispatch, CdpContext};
+use obscura_cdp::types::CdpRequest;
+use serde_json::{json, Value};
+
+async fn check_by_value(method: &str, await_promise: bool) {
+    let mut ctx = CdpContext::new();
+    let page_id = ctx.create_page();
+    let session_id = "by-value".to_string();
+    ctx.sessions.insert(session_id.clone(), page_id);
+
+    for (expression, expected_type, expected_value) in [
+        ("undefined", "undefined", None),
+        ("null", "object", Some(Value::Null)),
+        ("0", "number", Some(json!(0.0))),
+        ("false", "boolean", Some(json!(false))),
+        ("''", "string", Some(json!(""))),
+        ("({answer: 42})", "object", Some(json!({"answer": 42}))),
+        ("[null, 1]", "object", Some(json!([null, 1]))),
+    ] {
+        let expression = if await_promise {
+            format!("Promise.resolve({expression})")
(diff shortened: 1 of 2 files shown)


lkraider:
Additional validation: the isolated patch now builds and passes its focused CDP tests on macOS with render+stealth. A local native integration build carrying this patch and the separate form-navigation PR #889 also passes Playwright no-match count and all four raw undefined checks, with Chromium as control. Remaining Linux suites/no-render checks are still running; keeping this draft until they are reviewed.

SGavrl (approved):

PR #943, opened 2026-09-10
fix(obscura-js): enforce CORS preflight method, headers, and Content-Type value (#942)

### LINKED ISSUE #942: op_fetch_url CORS preflight under-enforced: Content-Type value unchecked, Allow-Methods/Allow-Headers never validated
### Summary

`op_fetch_url`'s CORS handling (`crates/obscura-js/src/ops.rs:2559-2608`) under-enforces the preflight in three ways, letting cross-origin requests proceed that a browser would block.

### Details

**A — `Content-Type` treated as always CORS-safelisted (ops.rs:2564-2569).**
The "needs preflight" test excludes `content-type` from the non-safelisted-header check regardless of its value:
```rust
custom_headers.keys().any(|k| {
    let kl = k.to_lowercase();
    kl != "accept" && kl != "accept-language" && kl != "content-language" && kl != "content-type"
})
```
Per the Fetch spec, `Content-Type` is CORS-safelisted only when its essence is `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain`. A cross-origin `POST` with `Content-Type: application/json` should trigger a preflight but does not.

**B — `Access-Control-Allow-Methods` never validated (ops.rs:2591-2608).**
After the preflight, only `Access-Control-Allow-Origin` and `-Allow-Credentials` are checked. The actual request method is never checked against `Access-Control-Allow-Methods`. A server returning `Access-Control-Allow-Origin: *` without listing `DELETE`/`PUT` still gets the non-safelisted request.

**C — `Access-Control-Allow-Headers` never validated (same block).**
Non-safelisted request headers (e.g. `Authorization`) are never checked against `Access-Control-Allow-Headers`, so a server that allows the origin but not the header still receives it.

### Trigger

```js
// page on https://evil.example
fetch("https://api.internal.corp/admin/users", {
  method: "DELETE",
  headers: { "Authorization": "Bearer x", "Content-Type": "application/json" },
  body: '{"id":1}',
});
```
Preflight fires (due to Authorization), but if the server returns only `Access-Control-Allow-Origin: *` — without `DELETE` in Allow-Methods or `Authorization` in Allow-Headers — the request proceeds anyway.

### Fix

(1) Value-check `Content-Type` in the safelist. (2) After the preflight, reject if the request method is not CORS-safelisted and absent from `Access-Control-Allow-Methods`. (3) Reject if any non-safelisted request header is absent from `Access-Control-Allow-Headers` (honoring the `*` wildcard only for non-credentialed requests, and never for `Authorization`).

### Affected
- `crates/obscura-js/src/ops.rs` — `op_fetch_url` preflight path

mnaza:
## What & why

`op_fetch_url`'s CORS handling under-enforced the preflight, letting cross-origin requests through that a browser blocks:

- **Content-Type always safelisted** — a cross-origin `POST` with `Content-Type: application/json` skipped preflight (the Fetch spec safelists Content-Type only for `application/x-www-form-urlencoded`, `multipart/form-data`, `text/plain`).
- **`Access-Control-Allow-Methods` never validated** — after preflight, only `Allow-Origin`/`Allow-Credentials` were checked, so a server echoing `Access-Control-Allow-Origin: *` still received a `DELETE`/`PUT` it never authorized.
- **`Access-Control-Allow-Headers` never validated** — non-safelisted request headers (e.g. `Authorization`) went through regardless.

Closes #942.

## Fix

- Value-check `Content-Type` in the safelist (essence must be one of the three form/text types).
- After the preflight response, validate the request method against `Access-Control-Allow-Methods` and each non-safelisted header against `Access-Control-Allow-Headers`.
- The `*` wildcard is honored only for **non-credentialed** requests, and **never** covers `authorization` (per the Fetch spec).

Logic is split into small pure helpers — `is_cors_safelisted_content_type`, `cors_request_needs_preflight`, `preflight_allows_method`, `preflight_allows_header` — so it's unit-testable without the network.

Blast radius is limited to cross-origin CORS-mode requests that require a preflight; simple requests are unaffected.

## Tests

TDD RED→GREEN. New unit tests for all four helpers cover: JSON vs form Content-Type safelisting, JSON-POST-needs-preflight, method authorization (listed / `*` / credentialed / safelisted), and header authorization (listed / `*` / `authorization` exclusion / credentialed). Verified RED first (permissive stubs), then GREEN. Full `obscura-js` suite: **402 passed, 0 failed**.

https://claude.ai/code/session_01W4C7V9e3rXKRY8jn1rYCoY

### DIFF
--- a/crates/obscura-js/src/ops.rs
+++ b/crates/obscura-js/src/ops.rs
@@ -2300,6 +2300,74 @@ fn cors_response_allows(
     }
 }
 
+/// A request's `Content-Type` value is CORS-safelisted only when its essence is
+/// one of these (Fetch spec). Any other value (e.g. `application/json`) makes
+/// the request non-simple and requires a preflight.
+fn is_cors_safelisted_content_type(value: &str) -> bool {
+    let essence = value
+        .split(';')
+        .next()
+        .unwrap_or("")
+        .trim()
+        .to_ascii_lowercase();
+    matches!(
+        essence.as_str(),
+        "application/x-www-form-urlencoded" | "multipart/form-data" | "text/plain"
+    )
+}
+
+/// Whether a request header is CORS-safelisted (does not by itself force a
+/// preflight). `Content-Type` is safelisted only for safelisted values.
+fn is_cors_safelisted_request_header(name: &str, value: &str) -> bool {
+    match name.to_ascii_lowercase().as_str() {
+        "accept" | "accept-language" | "content-language" => true,
+        "content-type" => is_cors_safelisted_content_type(value),
+        _ => false,
+    }
+}
+
+/// Whether a cross-origin CORS-mode request needs a preflight: a non-safelisted
+/// method or any non-safelisted header makes it non-simple.
+fn cors_request_needs_preflight(
+    method: &reqwest::Method,
+    headers: &HashMap<String, String>,
+) -> bool {
+    let non_simple_method = !matches!(method.as_str(), "GET" | "HEAD" | "POST");
+    let non_simple_header = headers
+        .iter()
+        .any(|(k, v)| !is_cors_safelisted_request_header(k, v));
+    non_simple_method || non_simple_header
+}
+
+/// After a preflight, whether the actual method is authorized by
+/// `Access-Control-Allow-Methods`. CORS-safelisted methods are always allowed;
+/// `*` is a wildcard only for non-credentialed requests.
+fn preflight_allows_method(
+    method: &reqwest::Method,
+    allow_methods: &str,
+    credentialed: bool,
+) -> bool {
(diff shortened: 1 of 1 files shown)


mnaza:
Superseded by #946 (`fix/942-cors-preflight`), which is already merged to `main` and implements #942 more completely — byte-safety on header values, per-header (128B) and aggregate (1024B) safelist caps, `Range`-header safelisting, rejection of malformed `Access-Control-Allow-Methods`/`-Allow-Headers` lists, and a preflight-status check, in addition to the method/header/Content-Type enforcement here. Closing as redundant; #942 is resolved by #946.

https://claude.ai/code/session_01W4C7V9e3rXKRY8jn1rYCoY

PR #1004, opened 2026-09-17
fix(cdp): correct accessible names and visibility

Railly:
Fixes unnamed controls in CDP snapshots: derive names from HTML labels and content, preserve ARIA precedence, and filter hidden nodes without losing referenced labels.

CSS visibility filtering requires a render build. Full-suite timing/MCP failures remain unresolved; #671 is unchanged.


SGavrl (approved):

crates/obscura-js/src/runtime.rs
manojniroshan232-netizen: M

manojniroshan232-netizen: Manoj

crates/obscura-js/src/runtime.rs
```diff
@@ -3831,6 +3837,29 @@ impl ObscuraJsRuntime {
...
+        };
+        prepared
+            .layout()
+            .styles
+            .iter()
+            .map(|(id, style)| {
+                (
+                    *id,
+                    AccessibilityStyle {
+                        display_none: style.display == obscura_render::Display::None,
```
manojniroshan232-netizen: ~    let previous = state 
 ~        .render_resources 
 ~        .set_sync_loading_enabled(false); 
 ~    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| capture(state))); 
 ~    state 
 ~        .render_resources 
 ~        .set_sync_loading_enabled(previous); 
 ~    match result { 
 ~        Ok(value) => value, 
 ~        Err(payload) => std::panic::resume_unwind(payload), 
      } 
  } 
 
 
 #[derive(Clone, Copy, Debug, Default)] 
 pub struct AccessibilityStyle { 
     pub display_none: bool, 
     pub visibility_hidden: Option<bool>, 
 } 
  
  #[derive(Debug, Clone)] 
  pub struct RemoteObjectInfo { 
      /// True when this object is a value that was thrown or that a promise system

